Issue #73, opened 2019-01-10
Accept a facility list CSV upload / Create FacilityList and FacilityListItem rows


jwalgran:
## Overview

Allow a logged in user to upload a facility CSV.

### Describe the solution you'd like

  - Return helpful validation errors
    - Maximum size (5MB)
    - missing column headers (we need to have at least country, name, and address columns)

### Additional Context

Depends on #70 
Depends on #72

Issue #66, opened 2019-01-10
Run Django unit tests in `scripts/test`


kellyi:
## Overview

`scripts/test` should run Django unit tests.

PR #86, opened 2019-01-11
Upload facility list CSV

jwalgran:
## Overview

NOTE: Ideally this should be updated as needed after #87 is complete so that we can verify that authenticated file uploads work from the UI.

Adds a new `/api/facility-lists` endpoint to the Django API connected to a new `FacilityListViewSet` and, for backward compatibility, connects the legacy `uploadTempFactory` endpoint to the create method on the same view set.

Connects #73
Connects #66 

## Demo

```
vagrant@vagrant:/vagrant$ sudo pip install httpie
[... pip output...]
vagrant@vagrant:/vagrant$ http -f POST http://localhost:8081/api/facility-lists/ name="Some facilities" file@src/django/api/management/commands/facility_lists/2.csv
HTTP/1.1 200 OK
Allow: GET, POST, HEAD, OPTIONS
Connection: close
Content-Length: 86
Content-Type: application/json
Date: Wed, 16 Jan 2019 17:23:21 GMT
Server: gunicorn/19.9.0
Vary: Accept, Origin, Cookie
X-Frame-Options: SAMEORIGIN

{
    "file_name": "2.csv",
    "id": 29,
    "is_active": true,
    "is_public": true,
    "name": "Some facilities"
}
```

## Notes

This is the first PR that introduces Django unit tests.

The `MAX_UPLOADED_FILE_SIZE_IN_BYTES` of 5MB was taken from the existing code.

Added djangorestframework-gis which includes helpers working with geometry fields and automatically renders geometry fields as OpenLayers maps in the Django admin.
  
### Setup
* ./scripts/update
* Reset the database
  * ./scripts/manage reset_schema
  * ./scripts/manage migrate
  * ./scripts/manage createsuperuser
* Create an `Organization` using http://localhost:8081/admin/api/organization/add/
* Install `httpie` on the VM with `vagrant@vagrant:/vagrant$ sudo pip install httpie`

### Test
* Verify that ./scripts/test runs the Django unit tests without error
* POST a file and verify that it is successful
```
http -f POST http://localhost:8081/api/facility-lists/ name="Some facilities" file@src/django/api/management/commands/facility_lists/2.csv
```
* Verify the newly added facility list and items using the admin pages
  * http://localhost:8081/admin/api/facilitylist/
  * http://localhost:8081/admin/api/facilitylistitem/


kellyi:
Taking a look at this!

src/django/api/tests.py
```diff
@@ -1,3 +1,164 @@
...
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.urls import reverse
+
+from rest_framework import status
+from rest_framework.test import APITestCase
+
+from api.models import FacilityList, FacilityListItem, Organization, User
+
+
+class FacilityListCreateTest(APITestCase):
```
kellyi: :clap:

src/django/api/views.py
```diff
@@ -64,3 +71,85 @@ def total_factories(request):
...
+        if 'country' not in parsed_header \
+           or 'name' not in parsed_header \
+           or 'address' not in parsed_header:
+            raise ValidationError(
+                'Header must contain country, name, and address fields.')
+
+    @transaction.atomic
+    def create(self, request):
+        # TODO: Require authenticated user with related organization
+        if 'file' not in request.data:
```
kellyi: Consider setting `'file'`, `'name'`, `'replaces'` and others of these string keys as constants.

kellyi:
> Install httpie on the VM with vagrant@vagrant:/vagrant$ sudo pip install httpie

Wonder if we should just install this to have it available by default? It's a useful tool and the VM won't be deployed, so I don't think there are many consequences for making it generally available in development.

kellyi:
Working well!

screen shot 2019-01-17 at 9 58 15 am

screen shot 2019-01-17 at 9 58 09 am

If we'd like to defer the auth part til later this is good to go in; otherwise I can re-review when auth is in!

src/django/api/tests.py
jwalgran: 👍

src/django/api/views.py
jwalgran: Constants are a good idea. Thanks.

jwalgran:
> Wonder if we should just install [httpie] to have it available by default?

I think that would be helpful, and pushed a small change to the provisioner to include it.

jwalgran:
I added constants to both this and #96. All comments are addressed so these are ready for another look.

kellyi:
Setting this up again.

jwalgran:
Thanks for the review.
I'm going to fix up the conflicts and then dip in to see what it will take to connect upload page now that DRF auth is working. If it looks bigger than I expect, I'll open up a separate issue.

jwalgran:
On second thought, rather than having two stacked PRs hanging open for a while I think I will just merge them after resolving the conflicts. We are in such an early stage of the new backend there is no real harm in it.

jwalgran:
#105 is a follow up issue for connecting the UI to this API.

Issue #70, opened 2019-01-10
Add django-rest-auth / connect frontend authentication and registration forms to DRF endpoints


jwalgran:
## Overview

Registration and authentication on the frontend uses DRF.

### Describe the solution you'd like

Get the following flows working:
  - Signup (creates a User and an Organization)
  - Login
  - Logout

### Additional Context

Authentication ADR added in #63
Depends on #69

Issue #46, opened 2019-01-04
Configure CORs


rbreslow:
CORS isn't configured to allow the frontend to access the API in development:

<img width="1078" alt="screen shot 2019-01-03 at 10 34 34 am" src="https://user-images.githubusercontent.com/14170650/50646427-ffc77d00-0f43-11e9-9c65-b8d801c227cb.png">

* Evaluate the best approach to resolve issues with CORS and execute

See related issue: https://github.com/raster-foundry/prediction-api/pull/85

See discussion: https://github.com/azavea/open-apparel-registry/pull/42#discussion_r245040083

kellyi:
Heads up that I am going to try to fix this in passing while working on #70 because signup and login require making a post request.

I think the preferred solution here is to use some version of Create React App's proxy -- https://facebook.github.io/create-react-app/docs/proxying-api-requests-in-development -- which matches how we set up Webpack Dev Server in other projects.

Since the production app will be served from same origin using Django-SPA, it seems like we may not need the CORS middleware in Django.

Issue #31, opened 2018-12-21
Rename `ShowOnly` component's `if` prop:


kellyi:
The `ShowOnly` component is used as a wrapped which determines, based on a prop, whether or not to show its child components.

The prop is currently named `if`, which is a reserved word. We should rename that and update it throughout the app.

Issue #75, opened 2019-01-10
Implement API system based on DRF token auth


jwalgran:
## Overview

Allow authenticated requests to be made with a token instead of a username and password.

### Describe the solution you'd like

 - Add an authenticated DRF endpoint for creating and deleting tokens associated with a User.
  - Update the frontend to use the new endpoint. The old app uses a hardcoded API key, so we don't have to worry about breaking compatibility between the updated front end and the old backend. 
    - Users can now have multiple API keys. Modify the frontend to allow for this.
  - To exercise token auth, add a secured endpoint that returns all FacilityList rows associated with the provided API key / token (the `getLists` route from the old application).

### Additional Context

Depends on #70

kellyi:
> Users can now have multiple API keys. Modify the frontend to allow for this.

After discussing this with @jwalgran I'm going to create a new issue for the multiple tokens part and just set up the API to afford having one token per user. I'm updating the UI in #110 to potentially enable having multiple tokens, but the updates to the API for multiple tokens will be a little more involved.

DRF itself intentionally sticks to a one-token-per-user model. There are a couple libraries which enable having multiple tokens, and it's also potentially possible to create one without a library. I'll make a note to that effect on the new issue.

PR #87, opened 2019-01-11
Add django-rest-auth and enable signup, login, and logout

kellyi:
## Overview

This PR adds django-rest-auth and wires up the frontend to enable users to sign up, login, and logout.

Along the way I replaced the CORS setup with a create-react-app proxy and also renamed the `ShowOnly` component's `if` prop to `showChildren`, which avoids a conflict with the JS `if` keyword.

I also added an updated 0001 migration in order to use email addresses as usernames.

Connects #70
Connects #46 
Connects #31 

## Demo

register
login-error
login

## Testing Instructions

- vagrant destroy your VM
- `./scripts/setup` and verify that it concludes successfully
- `./scripts/server` then visit the app in the browser on port 6543
- click on the "Register" link to register, then try signing up for an account. Verify that:
    - all required fields have validation messages about being required
    - mismsatching passwords also show a validation message
    - the form can successfully be submitted and creates a new user
- after your user saves, click log out
- click on the "Login" link and enter your username and password
- sign in and verify that you are redirected to the placeholder profile page and see the navbar login state change
- refresh the page and verify that you are signed in since you have an active session
- run `./scripts/test` and verify that the linter and tests pass


src/app/src/util/propTypes.js
```diff
@@ -0,0 +1,26 @@
...
+    [registrationFieldsEnum.otherContributorType]: string.isRequired,
+    [registrationFieldsEnum.password]: string.isRequired,
+    [registrationFieldsEnum.confirmPassword]: string.isRequired,
+    [registrationFieldsEnum.newsletter]: bool.isRequired,
+    [registrationFieldsEnum.tos]: bool.isRequired,
+}).isRequired;
+
+export const registrationFormInputHandlersPropType = shape(Object
+    .values(registrationFieldsEnum)
+    .reduce((accumulator, key) => Object.assign({}, accumulator, { [key]: func.isRequired })))
```
kellyi: This needs an initial object after the assign function.

kellyi:
I'm going to make three-ish followup cards for this:

1. although the reset password dialog is available and the Redux actions are set up, submitting the form is currently a noop, so we'll need to create a reset password workflow

2. likewise, it seems like we might want to interpose a "Confirm registration via email" step in between signing up and having an active account

3. we'll also want to make user profiles editable again. This can possible be done while working on #75

src/app/src/components/inputs/ControlledSelectInput.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+        return (
+            <span>
+                <ShowOnly showChildren={!disabled}>
+                    <div style={{ position: 'relative' }}>
+                        <p
+                            role="presentation"
+                            onClick={this.toggleSelect}
+                            onKeyPress={this.toggleSelect}
+                            className="form__select-input-container"
```
kellyi: This probably needs a `tabIndex=0` to ensure that it gains focus when tabbing through the form.

kellyi:
Realized that it appears that a user is immediately "logged in" after signing up because of how the response data is hydrated into the Redux state -- but no corresponding session is created, so users still have to log in to get one to start.

This seems like the correct thing to me except that it shouldn't appear that users are logged in immediately after signup. I created #98 to implement an email activation workflow step here.

src/app/src/App.jsx
```diff
@@ -22,38 +21,29 @@ import Maintenance from './containers/Maintenance';
...
-            nextProps.user.loaded !== this.props.user.loaded ||
-            nextProps.user.betaAccess !== this.props.user.betaAccess
-        );
-    }
+    // shouldComponentUpdate(nextProps) {
+    //     return (
+    //         nextProps.user.loaded !== this.props.user.loaded ||
+    //         nextProps.user.betaAccess !== this.props.user.betaAccess
+    //     );
+    // }
```
kellyi: This part can probably just be deleted.

Likewise I realized that it might be that the legacy User reducer &c is no longer used in the app. I'll check and if so just delete it altogether.

src/app/src/components/RegisterForm.jsx
```diff
@@ -0,0 +1,259 @@
+import React, { Component } from 'react';
+import { arrayOf, bool, func, shape, string } from 'prop-types';
+import { connect } from 'react-redux';
+import { Link } from 'react-router-dom';
+import Grid from '@material-ui/core/Grid';
+
+import AppGrid from '../containers/AppGrid';
+import Checkbox from './inputs/Checkbox';
```
kellyi: It occurred to me that this should also have a ControlledCheckbox component so we can gradually remove the uncontrolled Checkbox. It's a pretty easy fix so I'll push a commit with that change.

kellyi:
Oof, I just realized that I did not include creating an organization for the user as part of this. @jwalgran do you think it would make sense to spin that part off into a discrete card? Or should I add it here?

src/app/src/components/RegisterForm.jsx
kellyi: Handled in 
59836aa 

I also just deleted the existing uncontrolled components in 
f58af56

src/app/src/App.jsx
kellyi: Deleted all of that in 
a0ecdb7 

I left some placeholders for Contribute.jsx and will make an issue for adjusting that component.

kellyi: See https://github.com/open-apparel-registry/open-apparel-registry/issues/101

jwalgran:
> I just realized that I did not include creating an organization for the user as part of this. @jwalgran do you think it would make sense to spin that part off into a discrete card? Or should I add it here?

Uploading and matching facility lists is the most important thing to get working, and that requires and `Organization`. The existing form already has fields that are intended for the `Organzation` model. I would be ok with splitting `Organization` creation into a separate issue, but I would want to start working on it right away.

kellyi:
> Uploading and matching facility lists is the most important thing to get working, and that requires and Organization. The existing form already has fields that are intended for the Organzation model.

Sounds good. Chatted with you about this and it turns out what we want is to create one organization per user for now in order to have Organizations as placeholders.

a6b7a7c adjusts the user creation action to create an organization in passing. To test that you should:

- create a new user through the UI
- verify that an organization is also created via the Django admin view

src/django/oar/urls.py
```diff
@@ -29,6 +29,10 @@
     url(r'^api-auth/', include('rest_framework.urls')),
     url(r'^rest-auth/', include('rest_auth.urls')),
     url(r'^rest-auth/registration/', include('rest_auth.registration.urls')),
+    url(r'^user-login/', views.LoginToOARClient.as_view(),
```
jwalgran: Looking at the docs, django-rest-auth appears to have login and logout routes baked in https://django-rest-auth.readthedocs.io/en/latest/api_endpoints.html

If we need to override them for some reason, I think we should update the commit message to explain why.

src/app/src/util/util.js
```diff
@@ -14,34 +14,33 @@ export function DownloadCSV(data, fileName) {
     }
 }
 
-export const makeUserLoginURL = () => `${process.env.REACT_APP_API_URL}/user-login/`;
-export const makeUserLogoutURL = () => `${process.env.REACT_APP_API_URL}/user-logout/`;
+export const makeUserLoginURL = () => '/user-login/';
+export const makeUserLogoutURL = () => `/user-logout/`;
```
jwalgran: This does not need to be a backtick string any longer.

src/django/api/views.py
```diff
@@ -1,17 +1,36 @@
...
+        serializer = self.get_serializer(data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            pk = serializer.data['id']
+            user = User.objects.get(pk=pk)
+
+            Organization.objects.create(
+                admin=user,
+                name=user.email,
```
jwalgran: The Organization name should come from this field on the registration form.

screen shot 2019-01-18 at 9 27 04 am

src/app/src/components/FactoryInfo.jsx
```diff
@@ -90,7 +90,7 @@ export default class FactoryInfo extends PureComponent {
                         </p>
                     </div>
 
-                    <ShowOnly if={otherNames.length > 0}>
+                    <ShowOnly showChildren={otherNames.length > 0}>
```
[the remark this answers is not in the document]
kellyi: Good idea -- I like `when` better, so I'll make an update to replace `showChildren` with `when`.

src/app/src/components/RegisterForm.jsx
```diff
@@ -0,0 +1,245 @@
...
+            inputUpdates,
+            submitForm,
+            sessionFetching,
+        } = this.props;
+
+        if (sessionFetching) {
+            return null;
+        }
+
+        const errorMessages = error && error.length
```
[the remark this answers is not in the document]
kellyi: Makes sense -- I'll move this down to into the rendering. I think in this case I used the ternary because `error.map` will only work when error is an array, but that check can be nested within the rest of it.

src/app/src/components/RegisterForm.jsx
```diff
@@ -0,0 +1,223 @@
...
+    return {
+        fetching,
+        error,
+        form,
+    };
+}
+
+function mapDispatchToProps(dispatch) {
+    return {
+        inputUpdates: {
```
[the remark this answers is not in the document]
kellyi: :+1: Much better, so I'll use this to streamline things.

src/django/oar/urls.py
kellyi: Will do!

src/app/src/util/util.js
kellyi: I may have fixed this in a commit which did some linting, but I'll double check.

kellyi:
> The profile page for a user/organization should still work if there is not a session in place. 

That makes sense to me. I think I'm going to create a new issue for actually implementing the profile page, but for this PR I will make it so the profile page does not redirect if there is no logged in user.

kellyi:
> The old application shows validation errors on the registration form as "toast."
>
> I think we need to keep some sort of error reporting functionality.

Do you want to keep the toast? I have set up error reporting to show a list of validation errors like this:

screen shot 2019-01-18 at 12 01 18 pm

But it appears that some of the Django-sent messages are getting suppressed, as in the case when a user email address already exists:

screen shot 2019-01-18 at 12 01 44 pm

I generally prefer the non-toast version, but I think these messages could be moved to appear above the "Register" button rather than at the top of the screen to make them more visible. Likewise, I'll adjust things to have more of the specific Django error messages propagate to the UI.

kellyi:
cb8524e moves those validation messages to the bottom of the screen:

screen shot 2019-01-18 at 12 27 49 pm

src/app/src/components/RegisterForm.jsx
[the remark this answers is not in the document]
kellyi: Adjusted in 
a44cb4e

src/app/src/components/FactoryInfo.jsx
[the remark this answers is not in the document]
kellyi: Adjusted in 
a44cb4e

src/app/src/util/constants.js
```diff
@@ -0,0 +1,94 @@
+export const OTHER = 'Other';
+
+export const contributorTypeOptions = Object.freeze([
```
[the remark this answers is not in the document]
jwalgran: That is the DRYest way to handle it, but adding complexity to accomplish it is not a requirement since:
  * The list only appears 2 places.
  * The list is hard coded and may never change. If it does change, it will be be an infrequent occurrence.

jwalgran:
>Do you want to keep the [validation error] toast?
No. Inline is good.
>I think these messages could be moved to appear above the "Register" button rather than at the top of the screen to make them more visible
I think that is a good idea.

src/app/src/util/constants.js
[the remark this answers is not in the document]
kellyi: Sounds good, I'll just add the comments then.

src/app/src/components/RegisterForm.jsx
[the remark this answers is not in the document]
kellyi: Done in 
83afb77

kellyi:
I made https://github.com/open-apparel-registry/open-apparel-registry/issues/104 for User Profiles.

src/django/api/views.py
kellyi: Addressed in 
933b9b2

src/app/src/util/constants.js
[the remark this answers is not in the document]
kellyi: See 
72de9c9  and 
296d0a0

src/app/src/components/UserProfile.jsx
```diff
@@ -0,0 +1,87 @@
...
+}) {
+    if (fetching) {
+        return null;
+    }
+
+    if (!user) {
+        return (
+            <div style={tempStyle}>
+                <h1>
+                    Not logged in
```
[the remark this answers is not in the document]
kellyi: Created https://github.com/open-apparel-registry/open-apparel-registry/issues/104 for this, and I'll also link that as a TODO near the commented-out Profile line

src/app/src/App.jsx
```diff
@@ -2,16 +2,15 @@ import React, { Component } from 'react';
...
 import history from './util/history';
-import * as userActions from './actions/user';
 import Navbar from './components/Navbar';
 import Footer from './components/Footer';
-import AuthRegister from './containers/AuthRegister';
-import AuthLogin from './containers/AuthLogin';
-import Profile from './containers/Profile';
+import RegisterForm from './components/RegisterForm';
+import LoginForm from './components/LoginForm';
+// import Profile from './containers/Profile';
```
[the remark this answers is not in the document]
kellyi: See 31f2fc7dca88dbba544f02ea3bb93f2146462f91

src/app/src/util/util.js
kellyi: Yeah, this has been fixed.

src/django/oar/urls.py
kellyi: The reason I added the override was to enable making a GET request for session sign in. The base view supports POST for login out of the box, but session signin is a GET.

Since the commit itself is currently buried behind a bunch of fixups, I'll wait to edit its message until I'm rebasing and know in what commit this changes ends up living.

kellyi:
411ad85 formats DRF field-related error messages into a list that the UI can display and also adds a test for that function.

jwalgran:
I just noticed that we need to rebuild the `users.json` fixture to match the new model

```
vagrant@vagrant:/vagrant$ ./scripts/manage loadfixtures
Starting vagrant_database_1 ... done
  File "/usr/local/lib/python3.7/site-packages/django/db/models/options.py", line 566, in get_field
    return self.fields_map[field_name]
KeyError: 'first_name'
```

I pushed up a branch with an additional fixup commit that patches the fixture generator and includes a new users.json

https://github.com/open-apparel-registry/open-apparel-registry/commit/e06ece4d20506dfc2c63214c7306d9d573dffb94

kellyi:
Cool, I'll bring that patch into this branch.

src/app/src/components/RegisterForm.jsx
```diff
@@ -173,63 +176,36 @@ function mapStateToProps({
...
 
-function mapDispatchToProps(dispatch) {
+const getStateFromEventForEventType = Object.freeze({
+    [inputTypesEnum.checkbox]: getCheckedFromEvent,
+    [inputTypesEnum.select]: identity,
+    [inputTypesEnum.text]: getValueFromEvent,
+    [inputTypesEnum.password]: getValueFromEvent,
+});
+
+const mapDispatchToProps = memoize((dispatch) => {
```
jwalgran: Nice use of memoize to optimize this method.

src/app/src/components/RegisterForm.jsx
```diff
@@ -173,63 +176,36 @@ function mapStateToProps({
...
+    const makeInputChangeHandler = (field, getStateFromEvent) => e =>
+        dispatch(updateSignUpFormInput({
+            value: getStateFromEvent(e),
+            field,
+        }));
+
+    const inputUpdates = Object
+        .values(registrationFieldsEnum)
```
jwalgran: Nice reuse of our existing fields list. We are creeping toward a forms library :-)

src/app/src/components/LoginForm.jsx
```diff
@@ -106,7 +106,7 @@ class LoginForm extends Component {
                         />
                     </div>
                     <SendResetPasswordEmailForm />
-                    <ShowOnly when={error && error.length}>
+                    <ShowOnly when={!!(error && error.length)}>
```
jwalgran: Interesting. Why did we have to do this? I could see myself forgetting to do this and expecting truthy values to just work.

jwalgran:
facepalm

I forgot about the special admin user in the users fixture and the fact that username is now deprecated. One more fixup here: https://github.com/open-apparel-registry/open-apparel-registry/commit/a8d8baeb20c1e212cd56b7e5b72b94a2e9223865

kellyi:
> I forgot about the special admin user in the users fixture and the fact that username is now deprecated. One more fixup here

Awesome, thanks! I just cherry-picked that onto this branch and git pushed

jwalgran:
Nice job working through a tricky refactor.

src/app/src/components/LoginForm.jsx
kellyi: I set `when` as a `bool` prop-type in the `ShowOnly` component, but  I guess it doesn't automatically cast so it was complaining about receiving a number when there were errors. I noticed that the prior uses of `ShowOnly` had `length > 0` type expressions in them, and I guess this might be why.

kellyi:
Thanks for all your help with this! I'm going to squash this down to a smaller set of commits before merging.

PR #103, opened 2019-01-18
Remove Filestack

kellyi:
## Overview

- remove Filestack dependency
- drop User Profile Photo element from NavbarLoginButtonGroup
- drop User Profile Photo upload capability from Profile component

Connects #6 

## Testing

- get this branch and rm rf this project's node_modules in `/var/cache`
- `./scripts/update` to install node modules again and verify that it works
- `./scripts/server` and verify that the client builds successfully
-  sign in to an existing account and verify that the navbar no longer includes a profile photo element between "Language" and your user email address


src/app/src/containers/Profile.jsx
```diff
@@ -1,9 +1,13 @@
+/* eslint-disable */
+/* This component's gradually going to be deprecated in favor of
+   UserProfile.jsx. We retain the file until it has been fully
+   replaced */
+
```
jwalgran: Below this there is an `import '../styles/css/filestack.css';` line that we can drop along with removing the actual referenced css file.

kellyi:
> I found some filestack residue

Thanks for catching this! I removed the function & the css file and also retargeted the new commit at develop.

kellyi:
Thanks!

PR #106, opened 2019-01-21
Rename remaining `showChildren` prop to `when`

kellyi:
## Overview

In #87 I renamed the `ShowOnly` component's `if` prop to `when`, but tried out renaming it to `showChildren` first. However, when finally renaming everything to use `when` I missed three uses. This PR renames those three.

Connects #31 

## Testing

- `./script/server` and verify that the app works
- search the project for `showChildren` and verify that you don't see anymore uses of that name


kellyi:
Thanks! This was my mistake because I was grepping for `ShowOnly showChildren=` but that didn't catch the versions which had a line break!

PR #107, opened 2019-01-21
Remove Django from requirements.txt

kellyi:
## Overview

Since Django is provided by the base Docker image, remove it from
requirements.txt.

## Testing Instructions

- `vagrant destroy`
- `./scripts/setup`
- `./scripts/server`
- `./scripts/test`

to verify that each step above works


kellyi:
Thanks!

PR #108, opened 2019-01-21
Configure redux-devtools

kellyi:
## Overview

This PR adjusts the Redux store to make use of the Redux devtools browser extension if the environment is development and if the browser has Redux devtools installed; otherwise it's just the usual setup.

One of the main benefits of this is the ability to do time travel debugging via manipulating the store. Demo gif below:

## Demo

timetravel

## Notes

Extension is available for Chrome and Firefox: https://github.com/zalmoxisus/redux-devtools-extension

## Testing Instructions

- get this branch and start the server in a browser which does not have the Redux devtools extension installed and verify that everything works as it did before
- install Redux devtools then reload the app and verify that you can use the tools to inspect the store


kellyi:
Thanks!

Issue #98, opened 2019-01-17
Add user registration confirmation workflow


kellyi:
## Overview

#87 implements signups and makes users immediately active by setting `is_active` to true. However, we should interpose a step in which users have to click a link in a confirmation email to activate their accounts.

### Describe the solution you'd like

On successfully submitting the signup form, a user should be redirected to a confirmation page & not "logged in".

django-rest-auth/all-auth/etc should send a signup confirmation email to the user containing a link to confirm the account, something like `/auth/confirmation?token=xyz...` which would

- make whatever request is required to confirm the account and make it active
- display an "Account is activated" message on success
- have a link to `/auth/login` which the user would click to login

### Additional Context

GARP has a similar email confirmation workflow so it may be worth looking at how it was done there.

PR #110, opened 2019-01-21
Enable Users to create API Authorization Tokens

kellyi:
## Overview

This PR enables users to create API Authorization Tokens through the UI. In prep for that I restored a not-currently-editable version of the User Profile page (shown only to the logged in user), added GET, POST, and DELETE endpoints for the DRF Auth Token, and added an "API Tokens" section to the profile page with the ability to get, copy, delete, and generate a new API token.

Connects #75

## Demo

auth-token

screen shot 2019-01-22 at 2 57 10 pm

## Notes

As described in #119 multiple-tokens-per-user isn't directly supported by DRF, although there are several viable workarounds. The client was set up in such a way as to afford the possibility of having multiple tokens in case we decide to implement #119.

## Testing

- get this branch, then `./scripts/update` and `./scripts/server`
- sign up with a new user through the UI, then log in with that user
- on your user profile page, verify that you see the "API Tokens" section beneath the user data form
- use the API tokens section
    - generate a new token
    - copy a token to the clipboard
    - refresh the page and verify that the token is fetched
    - delete a token
- generate a token, then in the VM run the following command with your token:

```sh
http http://localhost:8081/token-auth-example/ 'Authorization: Token <YOUR-TOKEN>'
```

...and verify that that returns 200 with a JSON response including your user name.

- try the following command and verify that it returns 401/unauthorized:

```sh
http http://localhost:8081/token-auth-example/ 'Authorization: Token hello-world'
```


kellyi:
Note for me: looks like there's a react-copy-to-clipboard library available in the app already so it could be used to having a copy token to clipboard type button:

https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/package.json#L17

jwalgran:
It does not look like a `token-auth-example` route exists.

kellyi:
Woops! Forgot to push the last commit: f7d02a2aaef0b390de7d8392859c21f323c07746

kellyi:
@jwalgran I pushed to commits to address some quirks we discovered while testing:

c76773b prevents the API from automatically creating a token on login. I tried out overriding the standard django-rest-auth setup via settings, but unfortunately that wouldn't work because the User retrieved on login is gotten via a reference to the use on the Token object django-rest-auth generates. Instead, I adjusted the form login method to use Django's login and authenticate methods to create a session.

ba46eb4 updates the UI's signup behavior so that it will no longer give the appearance that a session has been created when the user initially signs up. Instead of redirecting to `/profile/:id`, we now redirect to `/auth/login` and dispense with setting the user state in Redux from the signup response. 

The redirect for the latter still happens kind of silently -- without a success message -- but rather than creating some placeholder I think we should wait to do that when creating the confirm registration workflow in #98

kellyi:
Uh oh! Will make a fixup to address.

kellyi:
Here's the django-rest-auth code which deletes the token on logout: https://github.com/Tivix/django-rest-auth/blob/master/rest_auth/views.py#L119

kellyi:
55ca365 makes and override to ensure that django-rest-auth doesn't delete tokens when a user logs out, and also adds some tests to check that

- user can login via the API
- user can logout via the API
- no token is automatically generated on login
- no token is automatically deleted on logout

jwalgran:
Taking a look now

kellyi:
> I only see one small state management issue remaining.

Oh, good catch -- this will be easy to fix. Essentially I forgot to remove the logged in user from the profilereducer on logout, so I'll add a commit to do that before merging.

kellyi:
02be044 resets the profile state when a user logs out. Previously it was only resetting the user section, but leaving the tokens behind. Now it resets the whole reducer back to its initial state.

Thanks for all your help with this! Going to merge this in once the tests pass on CI.

kellyi:
CI is currently failing due to an issue with the geocoding token not being set. However, the tests for this branch are passing.

Going to merge. Thanks again for your help!

Issue #77, opened 2019-01-10
Add a geocode method that processes a parsed FactoryListItem


jwalgran:
## Overview

Add a function that gecodes a `FactoryListItem` and saves the results.

### Describe the solution you'd like

This will ultimately be run by a background processing pipeline

  - Add a nullable `geocoded_address` char field to the `FacilityListItem` model.
  - Called with a `FacilityListItem` argument
  - Raises if `FacilityListItem.state` is not `PARSED`
  - Make a geocode request with the `address` and `country_code` (or the full name of the country associated with the code if that is required by the geocoder)
  - If successful set state to `GEOCODED`, populate `geocoded_point` with the best result, and store the geocoder response in the `processing_results` dict under a `geocoding` key. If the geocode fails set status to `ERROR` and put the error details in the `processing_results` dict.

### Additional Context

Depends on #74
Depends on #109

PR #136, opened 2019-01-23
Make Google API key available to Django via .env.backend

### LINKED ISSUE #109: Create a development API key for server-side Google geocoding
## Overview

Create a development API key for server-side Google geocoding 

  - Create a new Google API project under an Azavea-controlled account.
    - Create or use an existing Google billing account and attach it to the API project
  - Create a development API key with access to the Geocoding API but **do not** add it to the repo. The development key cannot be restricted to a domain and therefore needs to be kept secret and only injected into a development setup at provisioning time.
  - Update the development provisioner to read the geocoding API key from a host environment variable and make it available as a Django setting at runtime.

### Additional Context

- Both the current and the Phase 2 contracts includes budget for hosting and third party software services.
- Google requires an API key attached to a billing account for all geocoding requests. We are only charged if our usage exceeds  
https://developers.google.com/maps/documentation/geocoding/usage-and-billing

hectcastro:
## Overview

Take advantage of the existing S3 file syncing mechanism to pull down environment variable files from a secure location. Make the file's contents available to Django, and ensure that the variable is converted to a top-level Django setting inside of `settings.py`.

Fixes #109 

## Testing Instructions

* Use `setup` to provision the virtual machine (needed for `bootstrap` step below to work)
* Run `bootstrap` within the virtual machine to get the remote environment variable files
* Comment out the reference to `GOOGLE_GEOCODING_API_KEY` in `.env.backend`
* Run `server` within the virtual machine; ensure that Django raises a `ImproperlyConfigured` exception

### DIFF
--- a/deployment/ansible/roles/open-apparel-registry.environment/tasks/main.yml
+++ b/deployment/ansible/roles/open-apparel-registry.environment/tasks/main.yml
@@ -5,6 +5,12 @@
     regexp: "^AWS_PROFILE="
     line: "AWS_PROFILE={{ aws_profile }}"
 
+- name: Set OAR_SETTINGS_BUCKET
+  lineinfile:
+    path: "/etc/environment" 
+    regexp: "^OAR_SETTINGS_BUCKET="
+    line: "OAR_SETTINGS_BUCKET={{ oar_settings_bucket }}"
+
 - name: Set vm.max_map_count for ElasticSearch
   sysctl:
     name: vm.max_map_count
--- a/src/django/oar/settings.py
+++ b/src/django/oar/settings.py
@@ -12,6 +12,8 @@ https://docs.djangoproject.com/en/2.0/ref/settings/
 
 import os
 
+from django.core.exceptions import ImproperlyConfigured
+
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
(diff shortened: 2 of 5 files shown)


deployment/ansible/roles/open-apparel-registry.environment/tasks/main.yml
```diff
@@ -5,6 +5,12 @@
     regexp: "^AWS_PROFILE="
     line: "AWS_PROFILE={{ aws_profile }}"
 
+- name: Set OAR_SETTINGS_BUCKET
```
hectcastro: Without this, you can't run `bootstrap` from within the virtual machine. It works within the context of `vagrant up`, but that's because we set it inside of the `Vagrantfile`.

src/django/oar/settings.py
```diff
@@ -177,3 +179,8 @@
 
 # Application settings
 MAX_UPLOADED_FILE_SIZE_IN_BYTES = 5242880
+
+GOOGLE_GEOCODING_API_KEY = os.getenv('GOOGLE_GEOCODING_API_KEY')
+if GOOGLE_GEOCODING_API_KEY is None:
+    raise ImproperlyConfigured(
```
kellyi: Does this effectively guarantee that the application cannot run without the key being set?

If so, I'll dispense with internal unset-key error handling for the work on #77

kellyi: Nevermind, I confirmed that that is indeed the case:

```sh
django_1    |     self._wrapped = Settings(settings_module)
django_1    |   File "/usr/local/lib/python3.7/site-packages/django/conf/__init__.py", line 106, in __init__
django_1    |     mod = importlib.import_module(self.SETTINGS_MODULE)
django_1    |   File "/usr/local/lib/python3.7/importlib/__init__.py", line 127, in import_module
django_1    |     return _bootstrap._gcd_import(name[level:], package, level)
django_1    |   File "/usr/local/src/oar/settings.py", line 186, in <module>
django_1    |     'Invalid GOOGLE_GEOCODING_API_KEY provided, must be set')
django_1    | django.core.exceptions.ImproperlyConfigured: Invalid GOOGLE_GEOCODING_API_KEY provided, must be set
```

hectcastro: Yeah, I think that having a geocoding API key is a fair prerequisite. Part of the app would be busted without it.

PR #138, opened 2019-01-23
Add Terraform configuration for staging environment

### LINKED ISSUE #58: Add Terraform configuration for ECR; publish container images to ECR from CI
* Bootstrap `deployment/terraform` directory structure and initialize Terraform S3 backend
* Add `infra` script
* Using https://github.com/azavea/terraform-aws-ecr-repository, instantiate an ECR repository
* Add `cipublish` script that publishes container images to ECR
* Add support for `cipublish` and `infra` stages in `Jenkinsfile` (gated to develop and release branches)

### LINKED ISSUE #59: Add Terraform configuration for staging environment
* ECS with Fargate
* Include VPC with [terraform-aws-vpc](https://github.com/azavea/terraform-aws-vpc)
* Include PostgreSQL RDS with [terraform-aws-postgresql-rds](https://github.com/azavea/terraform-aws-postgresql-rds)
* Include log aggregation with CloudWatch Logs
* Layer in CloudFront distribution to front ALB

The result should essentially mirror the Terraform configuration in https://github.com/azavea/maiden.

rbreslow:
## Overview

* Bootstrap `deployment/terraform` directory structure and initialize Terraform S3 backend
* Add `infra` script
* Using https://github.com/azavea/terraform-aws-ecr-repository, instantiate an ECR repository
* Add `cipublish` script that publishes container images to ECR
* Add support for `cipublish` and `infra` stages in `Jenkinsfile` (gated to develop and release branches)
* ECS with Fargate
* Include VPC with [terraform-aws-vpc](https://github.com/azavea/terraform-aws-vpc)
* Include PostgreSQL RDS with [terraform-aws-postgresql-rds](https://github.com/azavea/terraform-aws-postgresql-rds)
* Include log aggregation with CloudWatch Logs
* Provision ACM certificates with [terraform-aws-acm-certificate](https://github.com/azavea/terraform-aws-acm-certificate)
* Layer in CloudFront distribution to front ALB
* Install `requests` so we can query the ECS Task Metadata endpoint to allow health checks to pass
* Thread `GOOGLE_GEOCODING_API_KEY` through to ECS task definition

Resolves #58 
Resolves #59 

## Notes

For accessing the Bastion, I generated a set of  `oar-stg` and `oar-prd` key pairs. They live in marriage rings.

I also had to copy the `NS` records from the `staging.openapparel.org` hosted zone into the `openapparel.org` hosted zone, during the Terraform cycle, in order for the ACM validation to go through.

I created an `OAR_AWS_ECR_ENDPOINT` credential on `civicci01`.
 
I think we should be OK with the default resource allocation for now (0.25 vCPU and 512MB of memory).

You'll probably want to take a look at the values set in the `terraform.tfvars` file on S3 as part of reviewing this PR.

## Testing Instructions

See that Terraform plan is clean:

```bash
vagrant@vagrant:/vagrant$ docker-compose -f docker-compose.ci.yml run --rm terraform
bash-4.4# export GIT_COMMIT=4d3e21f
bash-4.4# ./scripts/infra plan
...
------------------------------------------------------------------------

No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, no
actions need to be performed.
+ popd
/usr/local/src
bash-4.4#
```

Visit https://staging.openapparel.org/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,7 +30,7 @@ node {
 
 		env.OAR_SETTINGS_BUCKET = 'openapparelregistry-staging-config-us-east-1'
 
-		if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME =~ /test\//) {
+		if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('test/')) {
 			// Publish container images built and tested during `cibuild`
 			// to the private Amazon Container Registry tagged with the
 			// first seven characters of the revision SHA.
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,9 +45,10 @@ node {
 				}
 			}
 
-			// Plan and apply the current state of the instracture as
-			// outlined by the `develop` branch of the `open-apparel-registry`
-			// repository.
+			// Plan and apply the current state of the infrastructure as
+			// outlined by whatever branch of the `open-apparel-registry`
+			// repository passes the conditional above (`develop`, 
+			// `release/*`, `test/*`).
 			stage('infra') {
 				wrap([$class: 'AnsiColorBuildWrapper']) {
 					sh 'docker-compose -f docker-compose.ci.yml run --rm terraform ./scripts/infra plan'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,8 +7,10 @@ node {
 			checkout scm 
 		}
 
-		env.AWS_PROFILE= 'open-apparel-registry'
-		env.OAR_SETTINGS_BUCKET = 'openapparelregistry-development-config-us-east-1'
+		env.AWS_PROFILE = 'open-apparel-registry'
+		env.AWS_DEFAULT_REGION = 'us-east-1'
+
+		env.OAR_SETTINGS_BUCKET = 'openapparelregistry-testing-config-us-east-1'
 
 	    // Execute `setup` wrapped within a plugin that translates
 	    // ANSI color codes to something that renders inside the Jenkins
@@ -23,6 +25,34 @@ node {
 		stage('cibuild') {
 			wrap([$class: 'AnsiColorBuildWrapper']) {
 				sh './scripts/cibuild'
+			}
+		}
+
+		env.OAR_SETTINGS_BUCKET = 'openapparelregistry-staging-config-us-east-1'
+
+		if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME =~ /test\//) {
+			// Publish container images built and tested during `cibuild`
+			// to the private Amazon Container Registry tagged with the
+			// first seven characters of the revision SHA.
+			stage('cipublish') {
+				// Decode the ECR endpoint stored within Jenkins.
+				withCredentials([[$class: 'StringBinding',
+								credentialsId: 'OAR_AWS_ECR_ENDPOINT',
+								variable: 'OAR_AWS_ECR_ENDPOINT']]) {
+					wrap([$class: 'AnsiColorBuildWrapper']) {
+						sh './scripts/cipublish'
+					}
+				}
+			}
+
+			// Plan and apply the current state of the instracture as
+			// outlined by the `develop` branch of the `open-apparel-registry`
+			// repository.
+			stage('infra') {
+				wrap([$class: 'AnsiColorBuildWrapper']) {
+					sh 'docker-compose -f docker-compose.ci.yml run --rm terraform ./scripts/infra plan'
+					sh 'docker-compose -f docker-compose.ci.yml run --rm terraform ./scripts/infra apply'
+				}
 			}
 		}
 	} catch (err) {
--- a/deployment/terraform/cdn.tf
+++ b/deployment/terraform/cdn.tf
@@ -77,4 +77,9 @@ resource "aws_cloudfront_distribution" "cdn" {
     minimum_protocol_version = "TLSv1.2_2018"
     ssl_support_method       = "sni-only"
   }
+
+  tags {
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
 }
--- a/deployment/terraform/cdn.tf
+++ b/deployment/terraform/cdn.tf
@@ -1,7 +1,7 @@
 resource "aws_cloudfront_distribution" "cdn" {
   origin {
     domain_name = "origin.${var.r53_public_hosted_zone}"
-    origin_id   = "alb-${var.aws_region}"
+    origin_id   = "originAlb"
 
     custom_origin_config {
       http_port              = 80
@@ -22,7 +22,7 @@ resource "aws_cloudfront_distribution" "cdn" {
   default_cache_behavior {
     allowed_methods  = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
     cached_methods   = ["GET", "HEAD", "OPTIONS"]
-    target_origin_id = "alb-${var.aws_region}"
+    target_origin_id = "originAlb"
 
     forwarded_values {
       query_string = true
@@ -44,7 +44,7 @@ resource "aws_cloudfront_distribution" "cdn" {
     path_pattern     = "static/*"
     allowed_methods  = ["GET", "HEAD", "OPTIONS"]
     cached_methods   = ["GET", "HEAD", "OPTIONS"]
-    target_origin_id = "alb-${var.aws_region}"
+    target_origin_id = "originAlb"
 
     forwarded_values {
       query_string = false
@@ -74,7 +74,7 @@ resource "aws_cloudfront_distribution" "cdn" {
 
   viewer_certificate {
     acm_certificate_arn      = "${module.cert.arn}"
-    minimum_protocol_version = "TLSv1"
+    minimum_protocol_version = "TLSv1.2_2018"
     ssl_support_method       = "sni-only"
   }
 }
--- a/deployment/terraform/cdn.tf
+++ b/deployment/terraform/cdn.tf
@@ -1,0 +1,80 @@
+resource "aws_cloudfront_distribution" "cdn" {
+  origin {
+    domain_name = "origin.${var.r53_public_hosted_zone}"
+    origin_id   = "alb-${var.aws_region}"
+
+    custom_origin_config {
+      http_port              = 80
+      https_port             = 443
+      origin_protocol_policy = "https-only"
+      origin_ssl_protocols   = ["TLSv1", "TLSv1.1", "TLSv1.2"]
+    }
+  }
+
+  enabled         = true
+  is_ipv6_enabled = true
+  http_version    = "http2"
+  comment         = "${var.project} (${var.environment})"
+
+  price_class = "${var.cloudfront_price_class}"
+  aliases     = ["${var.r53_public_hosted_zone}"]
+
+  default_cache_behavior {
+    allowed_methods  = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
+    cached_methods   = ["GET", "HEAD", "OPTIONS"]
+    target_origin_id = "alb-${var.aws_region}"
+
+    forwarded_values {
+      query_string = true
+      headers      = ["*"]
+
+      cookies {
+        forward = "all"
+      }
+    }
+
+    compress               = false
+    viewer_protocol_policy = "redirect-to-https"
+    min_ttl                = 0
+    default_ttl            = 0
+    max_ttl                = 300
+  }
+
+  ordered_cache_behavior {
+    path_pattern     = "static/*"
+    allowed_methods  = ["GET", "HEAD", "OPTIONS"]
+    cached_methods   = ["GET", "HEAD", "OPTIONS"]
+    target_origin_id = "alb-${var.aws_region}"
+
+    forwarded_values {
+      query_string = false
+
+      cookies {
+        forward = "none"
+      }
+    }
+
+    compress               = true
+    viewer_protocol_policy = "redirect-to-https"
+    min_ttl                = 0
+    default_ttl            = 300
+    max_ttl                = 300
+  }
+
+  logging_config {
+    include_cookies = false
+    bucket          = "${aws_s3_bucket.logs.id}.s3.amazonaws.com"
+  }
+
+  restrictions {
+    geo_restriction {
+      restriction_type = "none"
+    }
+  }
+
+  viewer_certificate {
+    acm_certificate_arn      = "${module.cert.arn}"
+    minimum_protocol_version = "TLSv1"
+    ssl_support_method       = "sni-only"
+  }
+}
--- a/deployment/terraform/container_service.tf
+++ b/deployment/terraform/container_service.tf
@@ -1,0 +1,201 @@
+locals {
+  app_image = "${module.ecr-repository.repository_url}:${var.image_tag}"
+}
+
+#
+# Security Group Resources
+#
+resource "aws_security_group" "alb" {
+  vpc_id = "${module.vpc.id}"
+
+  tags {
+    Name        = "sgLoadBalancer${title(var.project)}${title(var.environment)}"
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
+}
+
+resource "aws_security_group" "ecs_cluster" {
+  vpc_id = "${module.vpc.id}"
+
+  tags {
+    Name        = "sgEcsCluster${title(var.project)}${title(var.environment)}"
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
+}
+
+#
+# ALB Resources
+#
+resource "aws_lb" "oar" {
+  name            = "alb${title(var.project)}${title(var.environment)}"
+  security_groups = ["${aws_security_group.alb.id}"]
+  subnets         = ["${module.vpc.public_subnet_ids}"]
+
+  tags {
+    Name        = "alb${title(var.project)}${title(var.environment)}"
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
+}
+
+resource "aws_lb_target_group" "oar" {
+  name = "tg${title(var.project)}${title(var.environment)}"
+
+  health_check {
+    healthy_threshold   = "3"
+    interval            = "30"
+    matcher             = "200"
+    protocol            = "HTTP"
+    timeout             = "3"
+    path                = "/health-check/"
+    unhealthy_threshold = "2"
+  }
+
+  port     = "80"
+  protocol = "HTTP"
+  vpc_id   = "${module.vpc.id}"
+
+  target_type = "ip"
+
+  tags {
+    Name        = "tg${title(var.project)}${title(var.environment)}"
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
+}
+
+resource "aws_lb_listener" "oar" {
+  load_balancer_arn = "${aws_lb.oar.id}"
+  port              = "443"
+  protocol          = "HTTPS"
+  certificate_arn   = "${module.cert.arn}"
+
+  default_action {
+    target_group_arn = "${aws_lb_target_group.oar.id}"
+    type             = "forward"
+  }
+}
+
+#
+# ECS Resources
+#
+
+resource "aws_ecs_cluster" "oar" {
+  name = "ecsCluster${title(var.project)}${title(var.environment)}"
+}
+
+data "template_file" "oar_ecs_task" {
+  template = "${file("task-definitions/oar.json")}"
+
+  vars = {
+    cpu    = "${var.fargate_cpu}"
+    image  = "${local.app_image}"
+    memory = "${var.fargate_memory}"
+
+    postgres_host     = "${module.database.hostname}"
+    postgres_port     = "${module.database.port}"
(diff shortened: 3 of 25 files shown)


deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,201 @@
...
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
+}
+
+#
+# ALB Resources
+#
+resource "aws_lb" "oar" {
+  name            = "alb${title(var.project)}${title(var.environment)}"
```
hectcastro: We typically do not include the project name again in the resource name. In this case, everything in the account will be associated with the OAR project.

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,201 @@
...
+
+    aws_region = "${var.aws_region}"
+
+    project     = "${var.project}"
+    environment = "${var.environment}"
+  }
+}
+
+resource "aws_ecs_task_definition" "oar" {
+  family                   = "oar"
```
hectcastro: This task definition should have some form of environment namespacing.

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,201 @@
...
+  cpu                      = "${var.fargate_cpu}"
+  memory                   = "${var.fargate_memory}"
+
+  execution_role_arn = "${aws_iam_role.ecs_task_execution_role.arn}"
+
+  container_definitions = "${data.template_file.oar_management_ecs_task.rendered}"
+}
+
+resource "aws_ecs_service" "oar" {
+  name            = "ecsService${title(var.project)}${title(var.environment)}"
```
hectcastro: This name does not surface what the service is. I think including that in the name is more important than `ecsService` and project name.

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,201 @@
...
+
+#
+# CloudWatch Resources
+#
+resource "aws_cloudwatch_log_group" "oar" {
+  name              = "log${title(var.project)}${title(var.environment)}"
+  retention_in_days = 30
+}
+
+resource "aws_cloudwatch_log_group" "oar_mangement" {
```
hectcastro: s/mangement/management/

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,201 @@
...
+  depends_on = [
+    "aws_lb_listener.oar",
+  ]
+}
+
+#
+# CloudWatch Resources
+#
+resource "aws_cloudwatch_log_group" "oar" {
+  name              = "log${title(var.project)}${title(var.environment)}"
```
hectcastro: It is not clear what logs will be in this log group based on its name.

deployment/terraform/database.tf
```diff
@@ -0,0 +1,104 @@
...
+    Environment = "${var.environment}"
+  }
+}
+
+resource "aws_db_parameter_group" "default" {
+  name        = "${var.rds_database_identifier}"
+  description = "Parameter group for the RDS instances"
+  family      = "${var.rds_parameter_group_family}"
+
+  parameter {
```
hectcastro: I think that we need to make these Terraform variables that get threaded through.

deployment/terraform/database.tf
```diff
@@ -0,0 +1,104 @@
...
+    name  = "random_page_cost"
+    value = "1"
+  }
+
+  parameter {
+    name  = "log_min_duration_statement"
+    value = "500"
+  }
+
+  parameter {
```
hectcastro: We can default this and the disconnections setting to `0`.

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,201 @@
...
+
+data "template_file" "oar_ecs_task" {
+  template = "${file("task-definitions/oar.json")}"
+
+  vars = {
+    cpu    = "${var.fargate_cpu}"
+    image  = "${local.app_image}"
+    memory = "${var.fargate_memory}"
+
+    postgres_host     = "${module.database.hostname}"
```
hectcastro: I think that this should be making use of the internal DNS record vs. the raw RDS instance FQDN.

deployment/terraform/firewall.tf
```diff
@@ -0,0 +1,67 @@
...
+  from_port        = 443
+  to_port          = 443
+  protocol         = "tcp"
+  cidr_blocks      = ["0.0.0.0/0"]
+  ipv6_cidr_blocks = ["::/0"]
+
+  security_group_id = "${aws_security_group.alb.id}"
+}
+
+# allow outbound traffic from the alb to the ecs cluster on the app port
```
hectcastro: These comments are helpful, but I think it is more accurate to associate the security group rules with the ECS service vs. the ECS cluster (if there were two services, we could have two different security groups with each).

deployment/terraform/iam.tf
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    actions = [
+      "sts:AssumeRole",
+    ]
+  }
+}
+
+resource "aws_iam_role" "ecs_task_execution_role" {
+  name               = "ecsTaskExecutionRole${title(var.project)}${title(var.environment)}"
```
hectcastro: I think that project should be removed from these resource names. I also think that it'll read easier with environment earlier in the name vs. at the end.

deployment/terraform/storage.tf
```diff
@@ -0,0 +1,18 @@
...
+  acl    = "private"
+
+  tags {
+    Name        = "${lower(var.project)}-${lower(var.environment)}-logs-${var.aws_region}"
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
+}
+
+module "ecr-repository" {
```
hectcastro: I think that this resource name should use an underscore vs. a hyphen to be consistent with all of the others.

deployment/terraform/variables.tf
```diff
@@ -0,0 +1,157 @@
+variable "project" {
+  default = "OpenApparelRegistry"
```
hectcastro: We typically put spaces between these since they're expected to be tag values vs. names.

deployment/terraform/storage.tf
```diff
@@ -0,0 +1,18 @@
...
+  source = "github.com/azavea/terraform-aws-ecr-repository?ref=0.1.0"
+
+  repository_name = "${lower(var.project)}"
```
hectcastro: I feel as though this should match what we have in the Docker Compose file for CI so that it is easier to establish the connection between the two.

deployment/terraform/task-definitions/oar.json
```diff
@@ -0,0 +1,30 @@
+[
+  {
+    "name": "openapparelregistry",
```
hectcastro: I think that there is some value in naming this `django`.

scripts/infra
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+set -e
+
+if [[ -n "${OAR_DEBUG}" ]]; then
+    set -x
+fi
+
+DIR="$(dirname "$0")"
```
hectcastro: It seems as though this variable is only used once. Could be removed, if desired.

src/django/oar/settings.py
```diff
@@ -11,6 +11,9 @@
 """
 
 import os
+import requests
+
+from django.core.exceptions import ImproperlyConfigured
```
hectcastro: Should be able to get this import via a rebase of `develop`.

deployment/terraform/cdn.tf
```diff
@@ -0,0 +1,80 @@
...
+
+    compress               = false
+    viewer_protocol_policy = "redirect-to-https"
+    min_ttl                = 0
+    default_ttl            = 0
+    max_ttl                = 300
+  }
+
+  ordered_cache_behavior {
+    path_pattern     = "static/*"
```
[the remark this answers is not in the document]
rbreslow: >Because WhiteNoise sends appropriate cache headers with your static content, the CDN will be able to cache your files and serve them without needing to contact your application again.

I followed these instructions to setup the `ordered_cache_behavior` block:

http://whitenoise.evans.io/en/stable/django.html#restricting-cloudfront-to-static-files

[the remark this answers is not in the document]
hectcastro: Cool. I was able to confirm that cache control headers are being set (60s).

src/django/oar/settings.py
rbreslow: I did a quick rebase but wasn't able to absolve these changes.

deployment/terraform/storage.tf
rbreslow: There isn't anything defined for `OAR_AWS_ECR_ENDPOINT` in `docker-compose.ci.yml`. This is stored as a credential in Jenkins as to not reveal our AWS Account ID.

rbreslow: That being said, maybe we could rename this to the string literal `openapparelregistry`?

rbreslow:
@hectcastro Think this is ready for another pass.

Jenkinsfile
```diff
@@ -25,6 +27,34 @@ node {
...
+								credentialsId: 'OAR_AWS_ECR_ENDPOINT',
+								variable: 'OAR_AWS_ECR_ENDPOINT']]) {
+					wrap([$class: 'AnsiColorBuildWrapper']) {
+						sh './scripts/cipublish'
+					}
+				}
+			}
+
+			// Plan and apply the current state of the instracture as
+			// outlined by the `develop` branch of the `open-apparel-registry`
```
hectcastro: It looks like this comment is inaccurate. Instead of using the Terraform configuration on `develop`, it'll use the configuration of whatever branch passed the conditional above (`develop`, `release/*`, `test/*`).

Jenkinsfile
```diff
@@ -25,6 +27,34 @@ node {
...
+				withCredentials([[$class: 'StringBinding',
+								credentialsId: 'OAR_AWS_ECR_ENDPOINT',
+								variable: 'OAR_AWS_ECR_ENDPOINT']]) {
+					wrap([$class: 'AnsiColorBuildWrapper']) {
+						sh './scripts/cipublish'
+					}
+				}
+			}
+
+			// Plan and apply the current state of the instracture as
```
hectcastro: s/instracture/infrastructure/

deployment/terraform/cdn.tf
```diff
@@ -0,0 +1,80 @@
...
+    include_cookies = false
+    bucket          = "${aws_s3_bucket.logs.id}.s3.amazonaws.com"
+  }
+
+  restrictions {
+    geo_restriction {
+      restriction_type = "none"
+    }
+  }
+
```
hectcastro: It looks like the `aws_cloudfront_distribution` resource supports `tags` now. I think it would be valuable to add the standard ones here too.

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,205 @@
...
+}
+
+#
+# Security Group Resources
+#
+resource "aws_security_group" "alb" {
+  vpc_id = "${module.vpc.id}"
+
+  tags {
+    Name        = "sgLoadBalancer"
```
hectcastro: I think that the name of this resource should include some reference to the service behind it. In this case, it could be something like `App` (e.g., `sgAppLoadBalancer`).

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,205 @@
...
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
+}
+
+resource "aws_security_group" "ecs_service" {
+  vpc_id = "${module.vpc.id}"
+
+  tags {
+    Name        = "sgEcsService"
```
hectcastro: Similar comment here about the associated service. It looks like we associate it directly with the Django Fargate service, so adding something like `App` in the name helps avoid challenges in a future scenario where we need to add another ECS service that cannot share the security group.

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,205 @@
...
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
+}
+
+#
+# ALB Resources
+#
+resource "aws_lb" "app" {
+  name            = "alb${var.environment}"
```
hectcastro: Similar comment with regard to including the name of the service behind the load balancers (e.g., `alb${var.environment}App`).

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,205 @@
...
+
+  tags {
+    Name        = "alb${var.environment}"
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
+}
+
+resource "aws_lb_target_group" "app" {
+  name = "tg${var.environment}"
```
hectcastro: Similar comment here regarding a reference to the underlying service.

deployment/terraform/firewall.tf
```diff
@@ -0,0 +1,67 @@
...
+  from_port = "${var.app_port}"
+  to_port   = "${var.app_port}"
+  protocol  = "tcp"
+
+  security_group_id        = "${aws_security_group.alb.id}"
+  source_security_group_id = "${aws_security_group.ecs_service.id}"
+}
+
+# allow inbound traffic from the alb to the ecs service on the app port
+resource "aws_security_group_rule" "app_ecs_cluster_ingress_alb" {
```
hectcastro: The naming of this resource is slightly inconsistent with the ones above it. `app` is being used to represent the security group associated with the ECS service, but here it is `app_ecs_cluster`. That makes it seem like the contents of the `aws_security_group_rule` is referencing different entities, but they are the same (`aws_security_group.ecs_service.id`).

I think continuing to just use `app` makes the most sense.

deployment/terraform/firewall.tf
```diff
@@ -0,0 +1,67 @@
...
+  from_port = "${module.database.port}"
+  to_port   = "${module.database.port}"
+  protocol  = "tcp"
+
+  security_group_id        = "${aws_security_group.ecs_service.id}"
+  source_security_group_id = "${module.database.database_security_group_id}"
+}
+
+# allow all inbound traffic from the ecs service to the rds instance
+resource "aws_security_group_rule" "rds_ecs_ingress" {
```
hectcastro: Similar comment about the inconsistency in naming here with regard to `ecs` vs. `app`.

deployment/terraform/firewall.tf
```diff
@@ -0,0 +1,67 @@
...
+  from_port = "${module.database.port}"
+  to_port   = "${module.database.port}"
+  protocol  = "tcp"
+
+  security_group_id        = "${module.database.database_security_group_id}"
+  source_security_group_id = "${aws_security_group.ecs_service.id}"
+}
+
+# allow all outbound traffic from the ecs service to ecr
+resource "aws_security_group_rule" "app_ecs_cluster_egress_ecr" {
```
hectcastro: I think it would be better to use an `https` identifier here vs. `ecr`, because there is nothing specific to ECR in the security group rule.

deployment/terraform/iam.tf
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    actions = [
+      "sts:AssumeRole",
+    ]
+  }
+}
+
+resource "aws_iam_role" "ecs_task_execution_role" {
+  name               = "ecsTaskExecutionRole${var.environment}"
```
hectcastro: In other resources, the environment comes right after the prefix vs. at the end (e.g., `ecs${var.environment}TaskExecutionRole`).

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,205 @@
...
+  vpc_id = "${module.vpc.id}"
+
+  tags {
+    Name        = "sgLoadBalancer"
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
+}
+
+resource "aws_security_group" "ecs_service" {
```
hectcastro: The name of this resource didn't immediately make sense to me. Based on its usage, I associate it with a specific service, not any ECS service. For example, if we had to add another service for background processing, that security group would need a service specific name.

deployment/terraform/storage.tf
hectcastro: In `docker-compose.ci.yml`, we build the container image as `open-apparel-registry`. Here we are rendering it as `openapparelregistry`.

I think that it makes sense to make the repository name (minus the repository specific bits) consistent across the two so that it is easier to make the connection between them.

deployment/terraform/firewall.tf
rbreslow: I thought I did a find and replace on `ecs_cluster` with `ecs_service`, but I must have missed the security group rule names. Thank you for catching this.

rbreslow: I agree that it makes more sense to just refer to this security group by `app`.

rbreslow:
@hectcastro This is ready for another pass.

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,205 @@
...
+#
+# ALB Resources
+#
+resource "aws_lb" "app" {
+  name            = "alb${var.environment}App"
+  security_groups = ["${aws_security_group.alb.id}"]
+  subnets         = ["${module.vpc.public_subnet_ids}"]
+
+  tags {
+    Name        = "alb${var.environment}"
```
hectcastro: I think that the `Name` tag here should match the `name` attribute of `aws_lb`.

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  port     = "80"
+  protocol = "HTTP"
+  vpc_id   = "${module.vpc.id}"
+
+  target_type = "ip"
+
+  tags {
+    Name        = "tg${var.environment}"
```
hectcastro: Same here. `Name` tag should match `name` attribute of `aws_lb_target_group`.

docker-compose.ci.yml
```diff
@@ -1,4 +1,17 @@
...
+  terraform:
+    image: quay.io/azavea/terraform:0.11.11
+    volumes:
+      - ./:/usr/local/src
+      - $HOME/.aws:/root/.aws:ro
+    environment:
+      - AWS_PROFILE=${AWS_PROFILE:-open-apparel-registry}
+      - GIT_COMMIT=${GIT_COMMIT:-latest}
+      - OAR_DEBUG=1
+      - OAR_SETTINGS_BUCKET=${OAR_SETTINGS_BUCKET:-openapparelregistry-staging-config-us-east-1}
```
hectcastro: It might be valuable to skip the inheritance of `OAR_SETTINGS_BUCKET` here and just default to the `openapparelregistry-staging-config-us-east-1` literal. `OAR_SETTINGS_BUCKET` is set to `openapparelregistry-development-config-us-east-1` in the virtual machine.

deployment/terraform/variables.tf
```diff
@@ -0,0 +1,191 @@
...
+variable "app_port" {
+  default = "8080"
+}
+
+variable "google_geocoding_api_key" {}
+
+variable "django_secret_key" {}
+
+variable "app_count" {
+  default = "2"
```
hectcastro: I think that we can safely make this 1, or override it for staging so that `app_count=1`, to save on costs.

deployment/terraform/firewall.tf
```diff
@@ -0,0 +1,67 @@
+# allow all inbound traffic to the load balancer on port 443
```
hectcastro: The security groups for being able to SSH into the bastion do not exist here. There is a security group associated with the bastion (`sg-0cc2163aa727b806f`), but it has no ingress or egress rules.

rbreslow: Good catch.

docker-compose.ci.yml
rbreslow: Agreed.

rbreslow:
@hectcastro Resolved your comments.

deployment/terraform/firewall.tf
hectcastro: There are still no egress rules. `yum` works because I think the packages are hosted on S3 (and we have a VPC endpoint setup for it), but I cannot:

- Connect to the database
- Interact with the ECS services
- Hit any service on the public internet via HTTP/HTTPS

rbreslow:
@hectcastro Hang on. I think I need to adjust the security group rules to allow the Bastion to communicate with RDS, etc.

rbreslow:
Alright, this looks good:

```bash
[ec2-user@ip-10-0-0-191 ~]$ pg_isready -h database.service.oar.internal
database.service.oar.internal:5432 - accepting connections
```

Trying to think if there's any other services we're going to need to give the Bastion access to. I saw in collab-for-children the Bastion has SSH egress rules, but we're not running any EC2 container instances, so I'm not sure if that's necessary here.

hectcastro:
I think the last two here: https://github.com/open-apparel-registry/open-apparel-registry/pull/138#discussion_r252357688

rbreslow:
@hectcastro I think I resolved your concerns. I went through the `firewall.tf` file and reorganized everything as well.

I didn't see the comment you mentioned, I think that's because it was within a conversation I resolved.

deployment/terraform/firewall.tf
```diff
@@ -0,0 +1,155 @@
...
+  type        = "ingress"
+  from_port   = 22
+  to_port     = 22
+  protocol    = "tcp"
+  cidr_blocks = ["${var.external_access_cidr_block}"]
+
+  security_group_id = "${module.vpc.bastion_security_group_id}"
+}
+
+resource "aws_security_group_rule" "bastion_ssh_egress" {
```
hectcastro: We don't have anywhere to SSH now, but I think this makes sense to keep with the expectation that Batch support will be coming soon.

PR #139, opened 2019-01-23
Geocode facility list items

kellyi:
## Overview

This PR adds a pipeline for geocoding PARSED facility list items into GEOCODED facility list items.

First commit adds a `geocoded_address` field to the base modal via an update to the 0001 migration.

Second commit

- adds the requests library
- creates some utility functions for doing geocoding
- creates a `geocode_facility_list_item` function for changing the item's state to GEOCODED if geocoding was successful and ERROR if it wasn't
- adds some tests for the pipeline

Connects #77

## Testing

- get this branch, `reset_schema` and `./scripts/manage migrate`
- `./scripts/test` to verify that the new pipeline is working as expected


jwalgran:
Looking at this now

src/django/api/libs.py
```diff
@@ -0,0 +1,44 @@
+from django.conf import settings
```
jwalgran: Why `libs.py` instead of something like `geocoding.py`?

src/django/api/tests.py
```diff
@@ -259,3 +265,98 @@ def test_generated_token_is_not_deleted_on_logout(self):
...
+
+    def test_unparsed_item_raises_error(self):
+        facility_list = FacilityList(header='address,country,name')
+        item = FacilityListItem(
+            raw_data='1400 JFK Blvd, Philly,us,Shirts!',
+            facility_list=facility_list
+        )
+
+        with self.assertRaises(ValueError) as cm:
+            geocode_facility_list_item(item)
```
jwalgran: Today I learned (or possibly relearned) that the binding of a `with` block stays in scope after the block completes. Good to know.

src/django/api/libs.py
kellyi: No real reason, although `geocoding.py` is probably a better name. Will update!

kellyi: I updated this and force pushed an amended commit.

src/django/api/tests.py
kellyi: I saw this in the Python tests documentation, which seemed like a nice way to test the type and result of an Exception!

kellyi:
Thanks for your help with this! I'm going to merge.

PR #141, opened 2019-01-24
Set Google API key for collectstatic

hectcastro:
## Overview

Within the Django container image `Dockerfile`, ensure that the `collectstatic` command sets the `GOOGLE_GEOCODING_API_KEY` environment variable to avoid runtime exceptions.

Also fixes a small `shellcheck` linting issue with (wait for it), `lint`.

## Testing Instructions

Ensure that the following STRTA execute cleanly within the context of the virtual machine:

- `update`
- `cibuild`


kellyi:
Setting this up now!

Issue #91, opened 2019-01-15
Fuse `containers` and `components` dirs in React app


kellyi:
## Overview

Currently the React app attempts to segment its components according to whether they are Redux-aware: `containers` if so; `components` if not.

However, the distinction's kind of porous in the existing code: some Redux-connected components live in `components`, and `containers` has some non-Redux-aware files.

Since we typically connect components to Redux pretty often, we should get rid of the `containers` dir altogether and move everything to `components`.

### Describe the solution you'd like

All the React components live in a `components` dir & there is no `containers` dir.

PR #142, opened 2019-01-24
Remove `containers/` dir

kellyi:
## Overview

The original React source had a `components` dir and a `containers` dir,
which is an idiomatic distinction meant to map on to a distinction
between React components which are connected to Redux and React
components which are not.

However, the original source files often didn't adhere to that
distinction, and our typical team design pattern is to use Redux state
pretty freely.

To reduce the dev environment's complexity, this commit

- moves the last five remaining files from `containers/` into
`components/`
- updates the import paths to work with the new file locations
- deletes the `containers/` dir

Connects #91

## Testing Instructions

- get this branch
- `./scripts/server` to verify that the application still works
- `./scripts/test` to verify that the tests still work

### DIFF
--- a/src/app/src/App.jsx
+++ b/src/app/src/App.jsx
@@ -12,10 +12,10 @@ import Footer from './components/Footer';
 import RegisterForm from './components/RegisterForm';
 import LoginForm from './components/LoginForm';
 import UserProfile from './components/UserProfile';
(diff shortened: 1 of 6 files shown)


jwalgran:
Looking at this now

kellyi:
Thanks!

Issue #148, opened 2019-01-24
Allow facility lists to be uploaded from the UI at /contribute and displayed on /lists


jwalgran:
## Overview

Restore the  /contribute and /lists pages connect them to the new /api/facility-lists endpoint

### Describe the solution you'd like

* Authenticated users can upload a CSV from the /contribute page
    * An optional name can be specified
    * The CSV can optionally replace another CSV. Show a list of existing CSVs associated with the authenticated user.
* Validation errors are shown to the user
* Uploaded CSVs for the authenticated user are show on /lists

Issue #105, opened 2019-01-18
Use organization attached to authenticated user when uploading CSV


jwalgran:
## Overview

Uploaded CSVs should be connected to the organization associated with the authenticated user.

### Describe the solution you'd like

  * Use token-auth to test this from the command line (see #86 and #110)
  * Associate the organization with the facilitylist
  * Reject unauthenticated uploads and uploads where the authenticated user does not have an organization

kellyi:
Consider possibly doing this in passing too: https://github.com/open-apparel-registry/open-apparel-registry/issues/101

PR #151, opened 2019-01-25
Enable users to upload and view facility lists

### LINKED ISSUE #101: Refactor Contribute.jsx component
## Overview

The Contribute.jsx component previously depended on a UserReducer and a set of user Redux actions. In #87 I replaced that infrastructure with an AuthReducer, so we'll need to refactor the Contribute.jsx component to work without relying on the UserReducer.

Refactor the Contribute component so that it uses either the AuthReducer or a newly-created set of Redux actions.

In passing it is likely also worthwhile to refactor the rest of this component if necessary:

- component state should be made as small as possible, and we should potentially use a function component if we can
- any involved data processing functions should go into utils
- normalize the mapStateToProps and mapDispatchToProps functions to use existing Civic Apps patterns rather than the current setup

kellyi:
## Overview

This PR enables users to upload and view facility lists through the UI.

Closes #101 
Connects #148
Connects #105 

## Demo

facility-lists
facility-lists-2

## Testing Instructions

- get this branch then
- `./scripts/manage reset_schema`
- `./scripts/manage migrate`
- `./scripts/server`
- visit the app on 6543 and register as a new user
- log in
- click on "Contribute" and verify that you are taken to the contribute form
- verify that you are able to upload new lists in the following series of conditions:
    - name and description empty
    - name or description empty
    - name and description but no replaces
    - name, description, and replaces
- verify that a list that has been replaced is listed as "(inactive)" in the dropdown
- click the link to view your lists
- verify that you see the table above
- sign out
- visit :6543/lists and verify that you have an indication that you should sign in
- visit :6543/contribute and verify that you have an indication that you should sign in


src/app/src/components/FacilityLists.jsx
```diff
@@ -0,0 +1,122 @@
...
+
+        const insetComponent = facilityLists.length
+            ? (
+                <Fragment>
+                    <Grid item xs={12}>
+                        <FacilityListsSelect />
+                    </Grid>
+                    <Grid
+                        container
+                        classNAme="margin-bottom-64"
```
kellyi: Typo

kellyi:
Looking at the gif above I realized that the replace list select dropdown wasn't very harmonious with the other existing styles, to I force pushed an amended last commit to change its style to look similar to the other inputs:

screen shot 2019-01-28 at 5 10 58 pm

src/app/src/actions/upload.js
```diff
@@ -0,0 +1,49 @@
...
+                form: {
+                    name,
+                    description,
+                },
+            },
+        } = getState();
+
+        window.console.dir(name, description, file);
+
+        // Validate that name does not have any special char:
```
jwalgran: Is this a TODO?

src/django/api/views.py
```diff
@@ -216,14 +215,18 @@ def create(self, request):
         header = csv_file.readline().decode().rstrip()
         self._validate_header(header)
 
-        # TODO: Get the organization from the authenticated user
-        organization = Organization.objects.first()
+        organization = Organization.objects.get(name=request.user.name)
```
jwalgran: I ran into a not found exception on this line. The `request.user` object has an `organization` property that should be used here. It's optional, so it needs to be checked for non-None. (Fixing this would take care of #105)

src/django/api/views.py
```diff
@@ -262,3 +266,9 @@ def create(self, request):
 
         serializer = self.get_serializer(new_list)
         return Response(serializer.data)
+
+    def list(self, request):
+        organization = Organization.objects.get(name=request.user.name)
```
jwalgran: The `request.user` object has an `organization` property that should be used here. It's optional, so it needs to be checked for non-None. (Fixing this would take care of #105)

src/app/src/actions/upload.js
kellyi: I ultimately removed this altogether. The legacy implementation had the validation done client-side in the React component, but I pushed it further out and ultimately we decided that we didn't need these validations yet and could add them server-side if we ever did.

src/django/api/views.py
kellyi: Good catch! I'll add this check.

src/django/api/views.py
kellyi: Added in 
15c572b along with a test.

src/django/api/views.py
kellyi: Added in 
15c572b along with a test.

kellyi:
@jwalgran I added 15c572b to handle validating the user organization as described in #105. I also added two tests to verify that the create and lists method reject with status 400 when the user does not have an organization.

jwalgran:
Looking at this now

kellyi:
Thanks! Going to merge.

Issue #79, opened 2019-01-10
Add a basic match method that processes a geocoded FactoryListItem


jwalgran:
## Overview

Add a function that geocodes `FactoryListItem` records.

### Describe the solution you'd like

This will ultimately be run by a background processing pipeline.

This is the most naive implementation, and will ultimately be the first of a stack of more sophisticated match checks that will be added.

  - Called with a `FacilityListItem` argument
  - Raises if `FacilityListItem.state` is not `GEOCODED`
  - Look for case-insensitive exact string matches in the `Facility` table. Name, address, and country must match.
    - If found, create a `FactoryMatch` row with a status of `AUTOMATIC`. Write a message to the `results` dict documenting:
      - A "version number" of the match logic. If matching evolves after we are in production we can tell when matches were made by an older version.
      - The fact that it was an exact string match.
    - If not found, make a new `Facility` row that references the `FacilityListItem`.

PR #152, opened 2019-01-25
Add naive facility matching function

jwalgran:
## Overview

The `match_facility_list_item` function will eventually be run in a background task to process geocoded `FacilityListItem` objects. This naive string matching approach is a placeholder so we can build out the entire CSV processing pipeline.

Connects #79

## Testing Instructions

* Verify that `./scripts/test --django` completes without error

### DIFF
--- a/src/django/api/tests.py
+++ b/src/django/api/tests.py
@@ -12,8 +12,11 @@ from rest_framework.authtoken.models import Token
 from rest_framework.test import APITestCase
 
 from api.constants import ProcessingResultSection
-from api.models import FacilityList, FacilityListItem, Organization, User
-from api.processing import parse_facility_list_item, geocode_facility_list_item
+from api.models import (Facility, FacilityList, FacilityListItem,
+                        FacilityMatch, Organization, User)
+from api.processing import (parse_facility_list_item,
+                            geocode_facility_list_item,
+                            match_facility_list_item)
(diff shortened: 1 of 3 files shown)


src/django/api/tests.py
```diff
@@ -12,351 +12,431 @@
...
-
-    def test_ungeocodable_address_returns_value_error(self):
-        with self.assertRaises(ValueError) as cm:
-            geocode_address('hello world', '$#')
-
-        self.assertEqual(cm.exception.args, ('No results were found',))
-
-
-class FacilityListItemGeocodingTest(TestCase):
+# class FacilityListCreateTest(APITestCase):
```
kellyi: Wanted to check to make sure it was intentional to comment out all these tests.

jwalgran: 🤦‍♂️

jwalgran: I restored the tests and rebased.

jwalgran:
I restored the tests I mistakenly left commented out. Thanks for the heads up.

kellyi:
> I restored the tests I mistakenly left commented out. Thanks for the heads up.

Cool! This is good to go in once the tests pass on CI!

jwalgran:
Thanks for the review. CI builds are failing on an unrelated geocoding unit test, so I am going to go ahead and merge.

Issue #90, opened 2019-01-15
Replace favicon and change or remove "powered by" link in app footer


kellyi:
## Overview

The favicon and "powered by" footer link in the React app both reference the prior technology vendor. We should remove or replace them

### Describe the solution you'd like

- the favicon should be replaced
- the "powered by" link in the footer should either be changed or removed altogether

kellyi:
I'm going to delete the favicon for now but I'll make a new card for adding a new one.

PR #156, opened 2019-01-29
Replace technology vendor link & favicon

kellyi:
## Overview

- replace technology vendor link in the footer
- replace favicon.ico with an empty file: previously it was the prior vendor's logo, but we'll eventually want to replace it with something OAR specific

Connects #90

## Demo

screen shot 2019-01-29 at 2 38 37 pm

## Testing Instructions

- serve this branch and visit :6543
- verify that you see the new link in the footer and that displays as expected
- verify that there's no favicon displayed


rajadain:
Can you explain why the favicon was replaced by an empty file, instead of being just deleted, or being replaced by a [blank valid image](http://transparent-favicon.info/)?

kellyi:
34e4381 uses a blank favicon instead of an empty file

Issue #145, opened 2019-01-24
Replace the filter dropdowns on the map page with autocomplete lists and populate them with data fetched from the new API


jwalgran:
## Overview

The contributor and country lists will have dozens or hundreds of items. Make it easy to select items by using an autocomplete

### Describe the solution you'd like

* Choose an autocomplete UI component and add it to the page.
* Populate the autocompletes with data fetched from new API endpoints.

### Additional Context

<img width="731" alt="screen shot 2019-01-24 at 11 43 00 am" src="https://user-images.githubusercontent.com/17363/51700918-77ca1580-1fcd-11e9-9fb3-021543ae58ed.png">

kellyi:
Material-UI, the UI library being used in the application, has this:

https://material-ui.com/demos/autocomplete/

Issue #81, opened 2019-01-11
Refactor ControlPanel.jsx component


kellyi:
## Overview

Currently the ControlPanel.jsx component blends numerous concerns:

- it renders the control panel view
- it handles making the API requests to retrieve facilities and countries and sources
- it handles mapping the response data from those requests into a shape the app expects

We should refactor it so that:

- the API requests go through Redux actions
- the response data mapping happens in testable utility functions
- the ControlPanel component is mostly concerned with rendering a view.

### Describe the solution you'd like

The Control Panel component is concerned only with rendering a view.

Mapping data happens in utils and those functions have unit tests.

Making API requests is done via Redux.

kellyi:
It's probably worthwhile to wait on doing this until DRF can return data to filter using the control panel. Going to mark as blocked for now.

kellyi:
Appears that there's an issue in one of the existing lifecycle methods:

![screen shot 2019-01-25 at 2 09 27 pm](https://user-images.githubusercontent.com/4165523/51767294-e4671200-20aa-11e9-94f3-88e4aaca028f.png)

Issue #147, opened 2019-01-24
Show a filtered list of facilities on the map


jwalgran:
## Overview

Connect the filters and the map to a new `/api/facilities` endpoint. 

Depends on #145 

### Describe the solution you'd like

* When at least one filter value has been selected, or a text search has been entered, make an API request to fetch matching `Facility` rows.
* Show dots on the map for each matching facility row.
* Show a message if no results match the filter.
* Multiple items selected from the same autocomplete list should be combined with OR (see #131).

PR #159, opened 2019-01-29
Update search filters & enable autocomplete

kellyi:
## Overview

This PR updates the app to retrieve contributor, contributor type, and country options from the API, then to render them in auto-complete-enabled multiselect boxes in the sidebar search thing.

I stored the state of selections in Redux so they'll be available via `getState` for a facilities search request. Last commit here has basic facilities reducer + action setup which demonstrates that those values are available in the action (so they don't have to be read from the view). 

Connects #145 
Connects #81 

## Demo

autocomplete

## Notes

I undertook a mild-to-medium refactor of the main page here: previously these search filters lived in a ControlPanel.jsx component that's pretty difficult to follow -- and that was embedded in the Map component.

I created a new FilterSidebar component for the search stuff so we can deprecate the ControlPanel, and I moved the search controls up one level in the component hierarchy so that they're now a sibling of the Map.jsx component. This required a mild change to the render method markup for Map.jsx.

I also removed two unused legacy reducers since they were no longer being used: Source and Navbar. This means that the Map reducer's the only one remaining that hasn't been redone with redux-act/immutability-helper.

## Testing Instructions

- serve this branch, then try out the sidebar search fields and verify that everything works as it should
- run `./scripts/test` and verify that everything passes


kellyi:
@rajadain this PR stops at the place where #147 would start; the last commit has a placeholder Redux action set up for retrieving facilities based on the selected filter options.

jwalgran:
Looking at this now.

jwalgran:
If we can do it easily, setting a minimum width on the sidebar would be a simple way to keep this UI usable on a portrait iPad

screen shot 2019-01-30 at 10 52 06 am

kellyi:
> If we can do it easily, setting a minimum width on the sidebar would be a simple way to keep this UI usable on a portrait iPad

Cool, that should be easy -- we're using inline CSS-in-JS styles for this so I'll just have to add a declaration to the style object.

kellyi:
Wow, fixing the bar size for iPad portrait width was trivially easy using the Material UI Grid: 
d84ae99 

screen shot 2019-01-30 at 2 23 45 pm

I made #163 for addressing the navbar, which seems like it'd also be pretty easy to do.

src/app/src/components/FilterSidebar.jsx
```diff
@@ -0,0 +1,178 @@
...
+            fetchContributors,
+            fetchContributorTypes,
+            fetchCountries,
+        } = this.props;
+
+        if (allListsAreEmpty(contributorsData, contributorTypesData, countriesData)) {
+            return fetchFilterOptions();
+        }
+
+        if (!contributorsData.length) {
```
jwalgran: Can you help me understand why we need to have these individual fetch calls? I think that we will always be fetching all the options, so `fetchFilterOptions` looks like all we need. It's good that we are dispatching three separate actions so we can fetch them in parallel.

src/django/oar/urls.py
```diff
@@ -46,6 +46,10 @@
         name='api_token_auth'),
     url(r'^token-auth-example/', views.token_auth_example,
         name='token_auth_example'),
+    url(r'^contributors/', views.all_contributors, name='all_contributors'),
```
jwalgran: What do you think about moving these endpoints under `/api`? Programatic consumers of the data will likely want access to these lists of valid filter options.

src/app/src/components/FilterSidebar.jsx
kellyi: Sure -- so this all happens in `componentDidMount` for this component. It's possible and likely that this specific component will be unmounted for a variety of reasons in the normal use of the app: for instance, I land on the main page, the options get fetched, then I go to the login page, then return to the main page.

On returning to the main page after login, I thought it'd be desirable not to have to re-fetch all these options again if they were already available in the Redux state. However, it's possible that 2/3 succeed and the other fails for whatever reason, so in that case I set this up so that the component will fetch options for any of them that are empty arrays in the Redux state in `componentDidMount`.

In practice I think it'll typically be the case that all of them or none of them succeed, but I can imagine the possibility that one of them succeeds and the others don't.

That said I can set this up to make it so that we `fetchAllOptions` unconditionally every time this component mounts and reset all options every time it unmounts. That would cause each client to make 3 get requests each time this component re-mounts, but that may not be too problematic.

src/django/oar/urls.py
kellyi: That sounds good to me -- I'll update the app to make this change.

kellyi: Done in 
90053b6

src/app/src/components/FilterSidebar.jsx
jwalgran: Thanks for the explanation. It's sound reasoning and it makes sense to leave it as is.

jwalgran:
Would prefer CI to pass before merging if possible.

Issue #131, opened 2019-01-22
Using AND logic in the search filters is the opposite of standard proceedure


katieashaw:
## Overview

A site user has fed back that using AND logic in the search filters is the opposite of standard procedure. 

### Expected Behavior

The user feels the site should use OR logic in the search filters and that this should either be changed, or stated explicitly in the UI that we're using AND logic.

### Actual Behavior

The site uses AND logic in the search filters. 

### Steps to Reproduce

* Search the site, selecting multiple filters from the Contributor and Country fields in combination. 

### Demo

N/A

### Additional context

The email from the site user is here: Removed link to avoid sharing personal contact details. Text of email in comment below.

The FAQ page the email refers to is here, and the question is in the "Searching the OAR" section: https://info.openapparel.org/faq/

dboyer:
Pasting content of email here to avoid displaying personal contact details:

My colleagues and I are very excited at the prospect of working with the data you’ve collected here, but I felt I should contact you about a small issue we all experienced.

It seems you’re aware of this issue as it’s listed in your FAQ, but I felt the need to reach out and emphasize to you the counter­intuitive nature of using AND logic instead of
OR.  I understand your reasoning for setting it up this way after reading your FAQ entry on the topic, but I still think you should reconsider the current design of the search box. 
My colleagues and I are very interested in the data you’ve accumulated, but it confused us all when we first interacted with it that these fields are using AND logic.  The issue is
not that you’ve chosen to use AND logic, but that you don’t indicate it anywhere in your UI when it is the opposite of standard procedure for a checkbox list in a query builder.  I
feel like titling each of the drop­downs “Filter by X” actually suggests that they all function the same.
 
I suggest adding a little note in the title for that dropdown, or just replacing it with OR logic and adding a new box that says something like “Search for Shared Facilities” or
something similar.  The UI should definitely reflect the fact that this is a known issue, as it’s the only FAQ entry for Search.

PR #166, opened 2019-01-30
Retrieve facilities as geojson & display on map

### LINKED ISSUE #85: Refactor Redux actions
## Overview

We should refactor the app's existing Redux actions to use redux-act & immutability-helper.

All the Redux actions normalize on using redux-act & immutability-helper.

### LINKED ISSUE #111: Refactor Map.jsx component & Redux actions
## Overview

Refactor the Map.jsx component and associated Redux actions to match our usual internal style.

Additionally:

- move Map.jsx into `components` dir
- change the component name to something more specific so that it doesn't clash with the builtin JS Map data type and won't be confused with the map operation

- reducer and Redux actions should be cleaned up using Redux Act & Immutability Helper
- some component state could possibly go into Redux state
- disabled ESLint rules should be re-enabled and the code fixed to get rid of the warnings & errors
- there are a bunch of mutating or otherwise complicated functions that can be streamlined and perhaps moved into utils and made testable

kellyi:
## Overview

This PR adjusts the API to return facilities from the API as a geojson feature collection, featured by whatever querystring parameters have been passed in via the client. It also updates the client to render facilities as markers on the map.

I kept some of the existing styles and behaviors from the old Map.jsx component, but moved it into OARMap.jsx. In particular: the markers are set to cluster when necessary at high zoom levels. Clicking on a cluster will zoom to the level at which the markers de-cluster, and clicking on a single facility navigates to a `/facility/:id` details view which currently has the placeholder geojson data for a facility.

Connects #147 
Closes #85 
Closes #111 
Connects #131 

## Demo

autocomplete

## Notes

~The last commit here adds filtering by contributor type or ID, but the syntax won't pass the linter so I'd be interested in suggestions on how to clean it up.~ (We silenced these with #NOQA for now.)

## Testing

- get this branch and `./scripts/update`
- `./scripts/server` then visit the main page and search for some facilities by some parameters
- different searches should return different sets of results, although the mock fixtures are all currently situated in New Jersey
- full search should return the full set of results
- if there are zero results, you should see the error message
- clicking on the marker cluster should zoom you to a bbox and level where the cluster breaks apart, either into single points or smaller clusters
- clicking on a single point should navigate you to a detail view where you'll see geojson data
- hard refreshing the detail view should retrieve data for the facility with that OAR ID, if any


src/app/src/styles/css/Map.css
```diff
@@ -36,6 +36,12 @@
   width: 100%;
 }
 
+#map-controls {
+    top: 64px !important;
+    right: 0;
+    padding: 10px;
+}
```
kellyi: Just so I don't forget: this change can be reverted.

kellyi:
Also: I didn't have this changeset handle rendering the detail view or make the map appearance/behavior change for the detail view, but I can create a new card for that work.

src/django/api/views.py
```diff
@@ -159,6 +164,65 @@ def all_countries(request):
...
+            queryset = queryset.filter(name__icontains=name)
+
+        if countries is not None and len(countries):
+            queryset = queryset.filter(country_code__in=countries)
+
+        if len(contributor_types):
+            type_match_facility_ids = [
+                match.facility.id
+                for match
+                in FacilityMatch
```
jwalgran: The `FacilityMatch` table may contain rows that are pending or rejected suggestions from the matching algorithm. We will want to add a `filter(status__in[FacilityMatch.AUTOMATIC, FacilityMatch.CONFIRMED])` to exclude pending or rejected matches.

src/django/api/views.py
```diff
@@ -159,6 +164,65 @@ def all_countries(request):
...
+                .filter(facility_list_item__facility_list__organization__org_type__in=contributor_types) # NOQA
+            ]
+
+            queryset = queryset.filter(id__in=type_match_facility_ids)
+
+        if len(contributors):
+            name_match_facility_ids = [
+                match.facility.id
+                for match
+                in FacilityMatch
```
jwalgran: The `FacilityMatch` table may contain rows that are pending or rejected suggestions from the matching algorithm. We will want to add a `filter(status__in[FacilityMatch.AUTOMATIC, FacilityMatch.CONFIRMED])` to exclude pending or rejected matches.

src/django/api/views.py
```diff
@@ -159,6 +164,65 @@ def all_countries(request):
...
+                .objects
+                .filter(facility_list_item__facility_list__organization__org_type__in=contributor_types) # NOQA
+            ]
+
+            queryset = queryset.filter(id__in=type_match_facility_ids)
+
+        if len(contributors):
+            name_match_facility_ids = [
```
jwalgran: We could probably use a [values QuerySet](https://docs.djangoproject.com/en/2.0/ref/models/querysets/#django.db.models.query.QuerySet.values) here rather to avoid hydrating a bunch of model instances when we only want the ID.

```python
 [int(f['facility__id']) for f in FacilityMatch.objects.filter(...).values('facility__id')]
```

src/django/api/views.py
```diff
@@ -159,6 +164,65 @@ def all_countries(request):
...
+
+        if name is not None:
+            queryset = queryset.filter(name__icontains=name)
+
+        if countries is not None and len(countries):
+            queryset = queryset.filter(country_code__in=countries)
+
+        if len(contributor_types):
+            type_match_facility_ids = [
+                match.facility.id
```
jwalgran: We could probably use a [values QuerySet](https://docs.djangoproject.com/en/2.0/ref/models/querysets/#django.db.models.query.QuerySet.values) here rather to avoid hydrating a bunch of model instances when we only want the ID.

```python
 [int(f['facility__id']) for f in FacilityMatch.objects.filter(...).values('facility__id')]
```

src/app/src/components/OARMap.jsx
```diff
@@ -190,7 +190,10 @@ class OARMap extends Component {
                 0,
                 (err, geojson) => {
                     if (!err) {
-                        this.oarMap.fitBounds(getBBoxForArrayOfGeoJSONPoints(geojson));
+                        this.oarMap.fitBounds(
+                            getBBoxForArrayOfGeoJSONPoints(geojson),
+                            { padding: 50 },
```
jwalgran: It would be helpful to add a commit message explaining why we need this padding and how we chose 50.

kellyi:
> if it is relatively simple to get the detail JSON to update when a second dot is clicked it will make for a more effective demo

Cool, this should be very easy to do -- the `/facilities/:id` string is updating so it's just a matter of adding some code in a lifecycle method to have it re-fetch data.

src/app/src/util/constants.oarmap.js
```diff
@@ -0,0 +1,124 @@
...
+
+export const oarMapLayerIDEnum = Object.freeze({
+    points: 'points',
+    pointsShadow: 'pointsShadow',
+    clusterCount: 'clusterCount',
+    unclusteredPoint: 'unclusteredPoint',
+    highlightedPoint: 'highlightedPoint',
+});
+
+export const oarMapLayers = Object.freeze([
```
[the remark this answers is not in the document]
kellyi: Yeah, this was extracted from this `addMapLayers` method in the legacy Map.jsx component:

https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/components/Map.jsx#L386

Aside from setting some values to constants, I didn't otherwise adjust it.

src/app/src/util/constants.oarmap.js
```diff
@@ -0,0 +1,124 @@
+export const MAPBOX_TOKEN = process
```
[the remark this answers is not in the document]
kellyi: Sounds good to me. It occurred to me that some of these values, like hex codes for colors, can just be hard coded in the app, so I'll do that for anything that doesn't need to remain secret.

src/app/src/components/OARMap.jsx
kellyi: Sounds good -- will update. I added this because I noticed in the demo gif that when it zoomed to the bbox, points on the edge of the bbox would be on the edge of the viewport, or nearly out of view.

src/app/src/util/constants.oarmap.js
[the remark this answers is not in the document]
kellyi: Fixed in 
6d3c6f9

kellyi:
> if it is relatively simple to get the detail JSON to update when a second dot is clicked it will make for a more effective demo

Done in 
bbace61

src/django/api/views.py
kellyi: Sounds good! I made this change in 9c62667e2f896af9a63df28a7636edb7ca7de292

src/django/api/views.py
kellyi: Sounds good! I made this change in 9c62667e2f896af9a63df28a7636edb7ca7de292

src/app/src/components/OARMap.jsx
kellyi: Haven't updated this yet but I will make sure to do so when rebasing before merging.

src/django/api/views.py
kellyi: That makes sense. I added this line to both the contributor type and contributor filter code in 
560a475

kellyi:
Thanks for your help with this! I squashed it down to 4 commits and will merge when tests pass on CI.

Issue #161, opened 2019-01-30
Create management commands to execute processing functions


jwalgran:
## Overview

Create a Django management commands that takes a processing step name and a `FacilityList` ID as arguments, optionally read a row index from an environment variable, and run a processing function on 1 or more `FacilityListItems`

### Describe the solution you'd like

- Rename `ProcessingResultSection` to `ProcessingAction`
  - Rename the options and drop `-ing` (i.e. `PARSE = 'parse`)
- Create a management command called `batch_process`
  - Requires an `--action` argument that must be one of the `ProcessingAction` values.
  - Requires a `--list-id` argument that must be an existing `FacilityList` id.
  - Checks for an `AWS_BATCH_JOB_ARRAY_INDEX` environment variable
    - If found, gets a query set of `FacilityListItem`s where the `id` and `row_index` match
    - If not found, gets a query set of `FacilityListItem`s where the `id` matches, ordered by `row_index`
  - Looks up the processing function based on the value of `--action` and runs each item in the query set through it, calling `.save()` on any models that are returned from the processing function.
    - The body of this loop should be wrapped in a transaction.

PR #172, opened 2019-02-05
Add Terraform resources for AWS Batch

### LINKED ISSUE #158: Add Terraform configuration for AWS Batch
* `aws_batch_compute_environment`
* `aws_batch_job_queue`
* `aws_batch_job_definition`
* Create an ECS Container Instance IAM Role and thread to `aws_iam_instance_profile`
* Create Batch IAM role
* Create Spot Fleet IAM role
* Utilize [ECS-Optimized Amazon Linux AMI](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html)

Once everything's wired up, things should look similar to what we have here: https://github.com/azavea/raster-foundry-deployment/tree/develop/core/terraform

jeancochrane:
## Overview

Create Terraform resources required to run jobs on AWS Batch.

Closes #158.

## Testing Instructions

* Run a `terraform plan` cycle to confirm that no new resources are required:

```
$ vagrant ssh
vagrant@vagrant:/vagrant$ docker-compose -f docker-compose.ci.yml run --rm terraform
bash-4.4# export GIT_COMMIT=<latest_develop_commit>
bash-4.4# ./scripts/infra plan
```

* If commits have come into `develop` that wiped out Batch resources, run `./scripts.infra apply` to create resources
* Log into OAR AWS account and search for the Batch console
* Select `Jobs > Submit Job`
* Add a dummy `Job Name`, select the most recent revision of `jobStagingDefault`, and hit `Submit Job`
* Confirm that your job transitions from `Submitted -> Runnable -> Succeeded`
    * If no Batch instances are running at the time you submit the job, you may have to wait a few minutes for EC2 to provision an instance

### DIFF
--- a/deployment/terraform/batch.tf
+++ b/deployment/terraform/batch.tf
@@ -24,7 +24,7 @@ resource "aws_batch_compute_environment" "default" {
     ]
 
     security_group_ids = [
-      "${aws_security_group.container_instance.id}",
+      "${aws_security_group.batch.id}",
     ]
 
     subnets = [
--- a/deployment/terraform/batch.tf
+++ b/deployment/terraform/batch.tf
@@ -51,9 +51,7 @@ data "template_file" "default_job_definition" {
   template = "${file("job-definitions/default.json")}"
 
   vars {
-    image_url      = "${module.ecr_repository.repository_url}:${var.image_tag}"
-    default_vcpus  = "${var.batch_default_jd_vcpus}"
-    default_memory = "${var.batch_default_jd_memory}"
+    image_url = "${module.ecr_repository.repository_url}:${var.image_tag}"
 
     postgres_host     = "${aws_route53_record.database.name}"
     postgres_port     = "${module.database.port}"
--- a/deployment/terraform/batch.tf
+++ b/deployment/terraform/batch.tf
@@ -1,0 +1,75 @@
+resource "aws_batch_compute_environment" "default" {
+  depends_on = ["aws_iam_role_policy_attachment.batch_policy"]
+
+  compute_environment_name = "batch${var.environment}DefaultComputeEnvironment"
+  type                     = "MANAGED"
+  state                    = "ENABLED"
+  service_role             = "${aws_iam_role.container_instance_batch.arn}"
+
+  compute_resources {
+    type           = "SPOT"
+    bid_percentage = "${var.batch_default_ce_spot_fleet_bid_percentage}"
+    ec2_key_pair   = "${var.aws_key_name}"
+    image_id       = "${var.batch_ami_id}"
+
+    min_vcpus     = "${var.batch_default_ce_min_vcpus}"
+    desired_vcpus = "${var.batch_default_ce_desired_vcpus}"
+    max_vcpus     = "${var.batch_default_ce_max_vcpus}"
+
+    spot_iam_fleet_role = "${aws_iam_role.container_instance_spot_fleet.arn}"
+    instance_role       = "${aws_iam_instance_profile.container_instance.arn}"
+
+    instance_type = [
+      "${var.batch_default_ce_instance_types}",
+    ]
+
+    security_group_ids = [
+      "${aws_security_group.container_instance.id}",
+    ]
+
+    subnets = [
+      "${module.vpc.private_subnet_ids}",
+    ]
+
+    tags {
+      Name               = "BatchWorker"
+      ComputeEnvironment = "Default"
+      Project            = "${var.project}"
+      Environment        = "${var.environment}"
+    }
+  }
+}
+
+resource "aws_batch_job_queue" "default" {
+  name                 = "queue${var.environment}Default"
+  priority             = 1
+  state                = "ENABLED"
+  compute_environments = ["${aws_batch_compute_environment.default.arn}"]
+}
+
+data "template_file" "default_job_definition" {
+  template = "${file("job-definitions/default.json")}"
+
+  vars {
+    image_url      = "${module.ecr_repository.repository_url}:${var.image_tag}"
+    default_vcpus  = "${var.batch_default_jd_vcpus}"
+    default_memory = "${var.batch_default_jd_memory}"
+
+    postgres_host     = "${aws_route53_record.database.name}"
+    postgres_port     = "${module.database.port}"
+    postgres_user     = "${var.rds_database_username}"
+    postgres_password = "${var.rds_database_password}"
+    postgres_db       = "${var.rds_database_name}"
+
+    environment              = "${var.environment}"
+    django_secret_key        = "${var.django_secret_key}"
+    google_geocoding_api_key = "${var.google_geocoding_api_key}"
+  }
+}
+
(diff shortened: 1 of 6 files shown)


deployment/terraform/job-definitions/default.json
```diff
@@ -0,0 +1,16 @@
+{
+  "image": "${image_url}",
+  "vcpus": ${default_vcpus},
+  "memory": ${default_memory},
+  "command": ["--version"],
```
jeancochrane: For the sake of simplicity I initialized the job definition to inherit from the app Gunicorn container and perform a dummy task. It seems like the ultimate goal for this job definition is to [read a row from the database and process it](https://github.com/open-apparel-registry/open-apparel-registry/issues/161), so if there's some more relevant work that we could do upfront to prove out important functionality (e.g. publish a new container image with a `python` entrypoint and use it in a job definition to ping the database via Django ORM) let me know and I can add it on.

deployment/terraform/job-definitions/default.json
```diff
@@ -0,0 +1,16 @@
+{
+  "image": "${image_url}",
+  "vcpus": ${default_vcpus},
+  "memory": ${default_memory},
```
jeancochrane: Memory and vCPU limits need to be integers in job definitions, so the interpolation requires that these variables remain unquoted.

deployment/terraform/variables.tf
```diff
@@ -189,3 +189,54 @@ variable "rds_free_memory_threshold_bytes" {
...
+variable "batch_service_role_policy_arn" {
+  default = "arn:aws:iam::aws:policy/service-role/AWSBatchServiceRole"
+}
+
+variable "spot_fleet_service_role_policy_arn" {
+  default = "arn:aws:iam::aws:policy/service-role/AmazonEC2SpotFleetTaggingRole"
+}
+
+variable "batch_default_ce_spot_fleet_bid_percentage" {
+  default = "40"
```
jeancochrane: We use this bid percentage in Raster Foundry, so I copied it over here as well.

deployment/terraform/variables.tf
```diff
@@ -189,3 +189,54 @@ variable "rds_free_memory_threshold_bytes" {
...
+variable "batch_default_ce_min_vcpus" {
+  default = "0"
+}
+
+variable "batch_default_ce_desired_vcpus" {
+  default = "0"
+}
+
+variable "batch_default_ce_max_vcpus" {
+  default = "16"
```
jeancochrane: It's hard to think of a reasonable maximum vCPU limit for Batch. I picked this one somewhat randomly, although looking now, it seems like Raster Foundry sets this to `256`.

deployment/terraform/variables.tf
```diff
@@ -189,3 +189,54 @@ variable "rds_free_memory_threshold_bytes" {
...
+    "m5",
+  ]
+}
+
+variable "batch_default_jd_vcpus" {
+  default = 2
+}
+
+variable "batch_default_jd_memory" {
+  default = 4096
```
jeancochrane: Memory and vCPU allocation were also set somewhat randomly, based on what I thought were reasonable upper bounds for the task defined in https://github.com/open-apparel-registry/open-apparel-registry/issues/161.

deployment/terraform/container_service.tf
```diff
@@ -25,6 +25,26 @@ resource "aws_security_group" "app" {
...
+  vpc_id = "${module.vpc.id}"
+
+  tags {
+    Name        = "sgContainerInstance"
+    Project     = "${var.project}"
+    Environment = "${var.environment}"
+  }
+}
+
+resource "aws_security_group_rule" "container_instance" {
```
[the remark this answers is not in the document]
jeancochrane: Good catch here. Come to think of it, is there any reason why the Batch container instances can't use the security group configured for the Fargate container instances? That seems to already have all the right ingress/egress rules for the bastion and database.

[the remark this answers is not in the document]
rbreslow: They could. We'd want to rename the app security group though to remove `EcsService`. I'm leaning against this though, because the app security group contains rules for the load balancer we wouldn't want applied to the Batch container instances.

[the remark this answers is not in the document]
jeancochrane: Gotcha. I think some duplication makes sense here in order to keep the two security groups separate.

deployment/terraform/batch.tf
```diff
@@ -0,0 +1,75 @@
...
+  state                = "ENABLED"
+  compute_environments = ["${aws_batch_compute_environment.default.arn}"]
+}
+
+data "template_file" "default_job_definition" {
+  template = "${file("job-definitions/default.json")}"
+
+  vars {
+    image_url      = "${module.ecr_repository.repository_url}:${var.image_tag}"
+    default_vcpus  = "${var.batch_default_jd_vcpus}"
```
rbreslow: I think we should embed values for VCPUs and memory within the job definition json, as we do in Raster Foundry. 

I don't see what value we get out of abstracting it here, because the job definition will still need to be regenerated, and we're not going to want to reuse the same VCPU and memory configuration for each job definition.

deployment/terraform/variables.tf
```diff
@@ -189,3 +189,54 @@ variable "rds_free_memory_threshold_bytes" {
...
+
+variable "batch_default_ce_max_vcpus" {
+  default = "16"
+}
+
+variable "batch_default_ce_instance_types" {
+  type = "list"
+
+  default = [
+    "c5",
```
rbreslow: Do you think the address matching algorithms are going to need instances this large, at least from the start? What do you think about starting with `t3`?

Also, I'm curious why we're only specifying the instance family. How does Batch determine which specific instance type to use?

jeancochrane: > What do you think about starting with `t3`?

I think `t3` is reasonable to start with, especially on staging. Later iterations of the address matching algorithm may need more compute depending on how complicated the approach becomes. (I've leaned towards `m5` instances for big dedupe jobs in the past.) I didn't actually know this Batch setup was for the deduplication process, though -- these values were just copied over from the Raster Foundry Batch setup.

> How does Batch determine which specific instance type to use?

The [docs](https://docs.aws.amazon.com/batch/latest/userguide/compute_environment_parameters.html) don't explain in detail, but based on anectodal experience it seems like Batch selects the instance type within the family that meets your desired vCPU count. With a desired count of 2 I've gotten exclusively `m5.large` instances, but the one time I accidentally set the desired count to 512 (wrong units!) I got a bunch of `m5.24xlarge` instances that each spun up for a few seconds. I'm not quite sure what the specific algorithm is for fulfilling the desired count, though (e.g. whether it will prioritize larger instances, or cheaper purchases).

rbreslow: Thank you for your detailed answers to my questions.

I think that using `m3` instances as we discussed out-of-band on Slack makes sense.

I also think your anecdotal experience about how Batch selects the instance type makes sense too.

🙆‍♂️

Issue #170, opened 2019-02-01
Remove organization-related fields from the User model


jwalgran:
## Overview

Our revised database schema splits the concept of `User` from `Organization`. The initial implementation of the `User` model includes some fields that are duplicates of fields on the `Organization` model and in one case a field that should be moved to the `Organization` model.

The `User` model should have the following fields:
  - is_staff
  - is_active 
  - email
  - username
  - should_receive_newsletter
  - has_agreed_to_terms_of_service
  - created_at
  - updated_at

The `Organization` model should have the following fields:
  - admin (FK to `User`)
  - name
  - description 
  - website 
  - org_type 
  - other_type
  - created_at
  - updated_at

For now we are treating user registration and organization creation as the same thing. The `User` registration / user profile forms should update the fields on the related `Organization` where appropriate.

Issue #135, opened 2019-01-23
In the Registration process, the free text description field for "Other" is not obligatory


katieashaw:
## Overview

In the process of Registering for a Contributor account, users are asked to select which type of organisation they represent. There is an "Other" option, with a free text box below to share more detail. This free text box has not been made obligatory if users select the "Other" category. 

### Expected Behavior

If a user selects the "Other" category for their organisation type, it should be obligatory that they include some information in the free text box below before they are able to sign up. 

### Actual Behavior

Users are able to select the "Other" organisation category and complete their registration process on the site without including any additional information in the free text box. 

### Steps to Reproduce

* Click "Register" on the OAR site
* Complete the necessary fields in the sign-up form, selecting the "Other" category for the organisation type
* Leave the free text box blank
* Click "Submit"

### Demo

N/A

### Additional context

N/A

PR #173, opened 2019-02-05
Update user model

kellyi:
## Overview

This PR updates the User and Organization models to have them have the fields presented in #170 and updates the user registration handler endpoint to match.

In addition, the changes:

- adjust the registration endpoint to make the free text description for "Other Contributor Type" required when "Other" had been selected
- update the makefixtures command and standard fixtures to move name, description, org_type, and other_org_type to the Organization data
- update the tests to account for `name` having been moved off the User model.

Connects #170 
Connects #135 

## Testing Instructions

- get this branch, then `./scripts/manage reset_schema`, `./scripts/manage migrate` and `./scripts/manage loadfixtures`
- through the UI, register for a new account and select "Other" as the contributor type, leaving the other contributor type free text field blank but completing all other required fields. Verify that you see an error message indicating that the free text description is required
- add a description, then save the account and verify that it saves correctly
- register another new account but this time select "Contributor Type" of "Auditor" and verify that it saves as expected
- verify that the new model matches what's described in #170


rajadain:
Taking a look now.

rajadain:
Can we make "Other Contributor Type" have the same red star as other required fields do?

image

kellyi:
> Can we make "Other Contributor Type" have the same red star as other required fields do?

Sure!

kellyi:
> Can we make "Other Contributor Type" have the same red star as other required fields do?

Handled in 7dd1c1d 

screen shot 2019-02-06 at 3 29 43 pm

rajadain:
The `has_agreed_to_terms_of_service` field is checked before submission in the front-end, but doens't seem to be set correctly in the database:

```sql
SELECT email, has_agreed_to_terms_of_service
FROM api_user
WHERE email LIKE 'ttuhinanshu%';

            email            | has_agreed_to_terms_of_service
-----------------------------+--------------------------------
 ttuhinanshu+oar1@xyz.com | f
 ttuhinanshu+oar2@xyz.com | f
```

kellyi:
> The `has_agreed_to_terms_of_service` field is checked before submission in the front-end, but doens't seem to be set correctly in the database:

Interesting, looks like I neglected to persist that or the newsletter field. Will push an update.

kellyi:
> The `has_agreed_to_terms_of_service` field is checked before submission in the front-end, but doens't seem to be set correctly in the database:

Fixed in 9ea439d 

```
Python 3.7.2 (default, Jan 23 2019, 02:35:07) 
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> u = User.objects.first()
>>> u.email
'kinnes@azavea.com'
>>> u.should_receive_newsletter
True
>>> u.has_agreed_to_terms_of_service
True
```

src/django/api/serializers.py
```diff
@@ -18,6 +23,41 @@ def create(self, validated_data):
         user.save()
         return user
 
+    def get_name(self, user):
+        try:
+            user_organization = Organization.objects.get(admin=user)
```
rajadain: Do you know if this `Organization.objects.get(admin=user)` result is cached in any way? Are we querying the database 5 times per serialization?

Also, why make the API different from the underlying models? Convenience, or do we expect the underlying models to change again?

kellyi: > why make the API different from the underlying models? Convenience, or do we expect the underlying models to change again?

This was partly convenience and partly because the original client's list of form fields used the `contributor type` and `other contributor type` nomenclature. Changing the variable names would then cast them out of sync with the client's form field labels rather than out of sync with the models. "Contributor" was also the term used on the client's search sidebar; the Organization model is something we introduced in the transition to DRF, but it's currently one-to-one with Users and isn't currently exposed anywhere in the client.

In general I thought it'd remain fairly easy to change it wholesale later if desired.

Good q about the caching: I don't know the answer offhand.

rajadain: I don't have enough context to see which way this project is headed, but in general simplicity is always better. If we can update the client's form fields to match the new variable names, that would be better. But that can also be a follow up task, doesn't necessarily have to be done as part of this.

kellyi: Makes sense to me. I think handling it in a new card would be best; I noticed that the filter reducers also use the `contributor` and `contributorType` variables, so it'd probably be worthwhile to have the changes be made & reviewed wholesale.

kellyi: Made https://github.com/open-apparel-registry/open-apparel-registry/issues/185 for addressing that btw.

rajadain: Great, thanks!

kellyi:
Thanks! Going to merge then rebase the other one.

PR #174, opened 2019-02-05
Add Management Command for Batch Processing Items

rajadain:
## Overview

Adds a management command for batch processing of facility list items.

Connects #161 

## Demo

image

## Notes

What should be done in the case of partial failures? For example, if geocoding fails for a few results, matching cannot be run at all. But neither can geocoding be re-run, leaving the user in a trapped state:

image

Should there be a flag to ignore error cases? Or to abort if there is a single error?

## Testing Instructions

* Check out this branch
* Run `manage batch_process --help`
* Run `manage batch_process --action [parse|geocode|match] --list-id [id]`
* Ensure it works correctly and handles failure cases

### DIFF
--- a/src/django/api/management/commands/batch_process.py
+++ b/src/django/api/management/commands/batch_process.py
@@ -1,0 +1,110 @@
+import os
+import sys
+
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from api.constants import ProcessingAction
+from api.models import FacilityList, FacilityListItem
+from api.processing import (parse_facility_list_item,
+                            geocode_facility_list_item,
+                            match_facility_list_item)
+
+ACTIONS = {
+    ProcessingAction.PARSE: parse_facility_list_item,
+    ProcessingAction.GEOCODE: geocode_facility_list_item,
+    ProcessingAction.MATCH: match_facility_list_item,
+}
+
+
+class Command(BaseCommand):
+    help = 'Run an action on all items in a facility list. If ' \
+           'AWS_BATCH_JOB_ARRAY_INDEX environment variable is set, will ' \
+           'process those items whose row_index matches it. Otherwise, will ' \
+           'process all items for the given facility list.'
+
+    def add_arguments(self, parser):
+        # Create a group of arguments explicitly labeled as required,
+        # because by default named arguments are considered optional.
+        group = parser.add_argument_group('required arguments')
+        group.add_argument('-a', '--action',
+                           required=True,
+                           help='The processing action to perform. '
+                                'One of "parse", "geocode", "match"')
+        group.add_argument('-l', '--list-id',
+                           required=True,
+                           help='The id of the facility list to process.')
+
+    def handle(self, *args, **options):
+        action = options['action']
+        list_id = options['list_id']
+
+        # Crash if invalid action specified
+        if action not in ACTIONS.keys():
+            self.stderr.write('Validation Error: Invalid action "{}". '
+                              'Must be one of "parse", "geocode", "match".'
+                              .format(action))
+            sys.exit(1)
+
+        process = ACTIONS[action]
+
+        # Crash if invalid list_id specified
+        try:
+            facility_list = FacilityList.objects.get(pk=list_id)
+        except FacilityList.DoesNotExist:
+            self.stderr.write('Validation Error: '
+                              'No facility list with id {}.'.format(list_id))
+            sys.exit(1)
+
+        row_index = os.environ.get('AWS_BATCH_JOB_ARRAY_INDEX')
+        if row_index:
+            items = FacilityListItem.objects.filter(
+                facility_list=facility_list,
+                row_index=row_index)
+        else:
+            items = FacilityListItem.objects.filter(
+                facility_list=facility_list)
+
+        result = {
+            'success': 0,
+            'failure': 0,
+        }
+
+        # Process all items, save affected items, facilities, matches,
+        # and tally successes and failures
+        for item in items:
+            try:
+                with transaction.atomic():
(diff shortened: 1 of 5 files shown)


jwalgran:
Looking at this now

jwalgran:
>What should be done in the case of partial failures?

Good question. When we actually run these jobs using AWS Batch the plan is to have each row be processed as a separate Array Job instance so that they are all run in parallel. When a Job fails with an application-level error, like a geocode failure, we can choose to exit with a zero status code or a non-zero status code. I am not completely sure which is going to be the best course of action. If exiting non-zero causes the entire job or chain or of dependent batch jobs to be aborted, that is probably not what we want. Based on my reading of the documentation that is what will happen, so my current thinking is that we move the item with a failing geocode to the `ERROR` state and exit zero as we are doing. When the match Job for that item starts up it is able to fail fast on the status check and exits zero as well. The full Array Job will then be "successful" at the system level, and we can report on the finer-grained failures in the application.

There are different ways in which the individual tasks can fail. How we deal with retrying them is an open question that I think we will want to get feedback from our advisors on after we have the system up and running. It can be as simple as giving them the ability to move the rolling the `status` of an item back and queuing up a new, one-off chain of Jobs for that item.

rajadain:
Rebased atop `develop` to resolve merge conflicts.

That matches the current behavior then, where the exit code is always 0 regardless of the process succeeding or not:

image

(I did it from within Vagrant this time to be closer to the command whose exit code we're inspecting)

rajadain:
Thanks for catching those! Makes sense. Have a look at 29af85f and see if it covers it.

src/django/api/management/commands/batch_process.py
```diff
@@ -0,0 +1,104 @@
...
+                            match.save()
+                    else:
+                        process(item)
+                        item.save()
+
+                    if item.status == FacilityListItem.ERROR:
+                        result['failure'] += 1
+                    else:
+                        result['success'] += 1
+                except DatabaseError:
```
rajadain: I'm catching a `DatabaseError` here because that's what `transaction.atomic()` will throw. `process` has error handlers inside it, and will catch any local errors.

src/django/api/management/commands/batch_process.py
```diff
@@ -0,0 +1,104 @@
...
+                try:
+                    if action == ProcessingAction.MATCH:
+                        facility, match = process(item)
+                        with transaction.atomic():
+                            item.save()
+                            facility.save()
+                            match.save()
+                    else:
+                        process(item)
+                        item.save()
```
rajadain: Should this single save also be wrapped in `transaction.atomic()`?

rajadain:
Thanks for all the detailed and thoughtful feedback! I really appreciate it!

I added in your two commits to this branch.

I also added a final fixup b85c439 which does the following:

* Crashes immediately on incorrect `action` specification, rather than going through a `try` `raise` `except` dance
* Moves `transaction.atomic()` further up to encapsulate all the `process` and `.save()` calls
* Adds an `except ValueError` block within the loop to log and count those as failures, in a way that doesn't stop the larger process
* Updates the `except DatabaseError` block to log the error and crash

src/django/api/management/commands/batch_process.py
```diff
@@ -0,0 +1,110 @@
...
+                        if action == ProcessingAction.MATCH:
+                            facility, match = process(item)
+                            item.save()
+                            facility.save()
+                            match.save()
+                        else:
+                            process(item)
+                            item.save()
+
+                        if item.status == FacilityListItem.ERROR:
```
rajadain: This probably doesn't need to be within `transaction.atomic()`

src/django/api/management/commands/batch_process.py
```diff
@@ -0,0 +1,110 @@
...
+
+            # Print failures
+            if result['failure'] > 0:
+                self.stdout.write(
+                    self.style.ERROR(
+                        '{}: {} failures'.format(
+                            action, result['failure'])))
+
+        except FacilityList.DoesNotExist:
+            self.stderr.write('Validation Error: '
```
rajadain: Should we also `sys.exit(1)` here?

src/django/api/management/commands/batch_process.py
jwalgran: Correct. We are done with the database at this point, so we could exit the `with` block here. Not a big deal since we aren't doing anything expensive before we start the next iteration.

src/django/api/management/commands/batch_process.py
jwalgran: I think so. It makes sense to fail the Job the AWS Batch level if one of the arguments was wrong.

rajadain:
The new commits seem to have made some tests fail though:

```
======================================================================
ERROR: test_matches_existing_facility (api.tests.FacilityListItemMatchingTest)
----------------------------------------------------------------------
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/utils.py", line 83, in _execute
    return self.cursor.execute(sql)
psycopg2.IntegrityError: insert or update on table "api_facilitymatch" violates foreign key constraint "api_facilitymatch_facility_id_7280e334_fk_api_facility_id"
DETAIL:  Key (facility_id)=() is not present in table "api_facility".
```

Should there be a migration to go along with the new GUID stuff?

src/django/api/management/commands/batch_process.py
rajadain: Should I do the same for `CommandError` beneath this?

rajadain:
Alright, removed a bunch of unnecessary nesting in `try` and `sys.exit(1)` which canceled out nicely in 30fb6cf. Going to squash changes and merge when green. Thanks for all your help with this!

PR #181, opened 2019-02-06
Configure Amazon SES for Django

### LINKED ISSUE #61: Configure SES for Django
* Enable outbound email on `openapparel.org`
* Request sending limit increase to `50,000` to get out of SES sandbox
* Add the IAM policy necessary to the IAM role associated with the ECS task
* Wire up https://github.com/azavea/django-amazon-ses

rbreslow:
## Overview

* Wire up Amazon SES in Terraform
* Add SES IAM policy
* Wire up [django-amazon-ses](https://github.com/azavea/django-amazon-ses)
* Add `sendtestemail` management command

Closes #61 

## Testing Instructions

First, I triggered Jenkins to deploy my changes to the staging environment:

```bash
$ git push -u origin feature/jrb/amazon-ses:test/jrb/amazon-ses --force
```

Then, I kicked off a `StagingAppCLI` ECS task to run the `sendtestemail` management command:

https://console.aws.amazon.com/ecs/home?region=us-east-1#/clusters/ecsStagingCluster/tasks/b6f3376e-58ce-4fe0-be13-e57a74ad7457/details

image

image

image


deployment/terraform/iam.tf
```diff
@@ -15,10 +18,30 @@ data "aws_iam_policy_document" "ecs_task_execution_assume_role" {
...
+data "aws_iam_policy_document" "ses_send_email" {
+  statement {
+    effect = "Allow"
+
+    resources = ["*"]
+    actions   = ["ses:SendRawEmail"]
+  }
+}
+
+resource "aws_iam_role" "ecs_task_role" {
```
rbreslow: This is different than an ECS Task Execution Role: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_IAM_role.html

src/django/api/management/commands/sendtestemail.py
```diff
@@ -0,0 +1,30 @@
+from django.core.management.base import BaseCommand
```
hectcastro: I would consider omitting this management command from the PR prior to merging. It is helpful to test this PR, but has limited value later.

If we needed to test email functionality further down the road, I suspect that we would copy and paste an email sending code snippet from the application into a Django shell (to ensure we're replicating what the application does).

deployment/terraform/ses.tf
```diff
@@ -0,0 +1,7 @@
+resource "aws_ses_domain_identity" "app" {
+  domain = "${var.r53_public_hosted_zone}"
+}
+
+resource "aws_ses_domain_dkim" "app" {
```
hectcastro: This is interesting. Until now, we haven't codified this process. Interested to see how it plays out.

deployment/terraform/iam.tf
```diff
@@ -15,10 +18,30 @@ data "aws_iam_policy_document" "ecs_task_execution_assume_role" {
...
+  name               = "ecs${var.environment}TaskRole"
+  assume_role_policy = "${data.aws_iam_policy_document.ecs_assume_role.json}"
+}
+
+resource "aws_iam_role_policy" "ses_send_email_policy" {
```
hectcastro: I would consider dropping the `_policy` suffix from this resource name since it is redundant (the resource is named `aws_iam_role_policy`).

```suggestion
resource "aws_iam_role_policy" "ses_send_email" {
```

deployment/terraform/iam.tf
hectcastro: I would consider renaming this resource so that it is application specific. Similar to the security group associated with an ECS service, if we had a background worker service, it is possible that the role associated with that task would be different than this one.

```suggestion
resource "aws_iam_role" "app_task_role" {
```

deployment/terraform/iam.tf
```diff
@@ -1,4 +1,7 @@
-data "aws_iam_policy_document" "ecs_task_execution_assume_role" {
+#
+# ECS IAM resources
+#
+data "aws_iam_policy_document" "ecs_assume_role" {
```
hectcastro: Makes sense to reuse this trust relationship. 👍

deployment/terraform/dns.tf
```diff
@@ -73,3 +73,20 @@ resource "aws_route53_record" "www_ipv6" {
...
+resource "aws_route53_record" "ses_verification" {
+  zone_id = "${aws_route53_zone.external.zone_id}"
+  name    = "_amazonses.${var.r53_public_hosted_zone}"
+  type    = "TXT"
+  ttl     = "300"
+  records = ["${aws_ses_domain_identity.app.verification_token}"]
+}
+
+resource "aws_route53_record" "ses_dkim" {
+  count   = 3
```
hectcastro: Not super important, but it would be interesting to try and get the length of `dkim_tokens` with `length()` vs. hardcoding 3 here.

deployment/terraform/container_service.tf
```diff
@@ -159,6 +164,7 @@ resource "aws_ecs_task_definition" "app_cli" {
   cpu                      = "${var.fargate_cpu}"
   memory                   = "${var.fargate_memory}"
 
+  task_role_arn      = "${aws_iam_role.ecs_task_role.arn}"
```
hectcastro: See justification for this suggestion where the resource is defined below.

```suggestion
  task_role_arn      = "${aws_iam_role.app_task_role.arn}"
```

deployment/terraform/container_service.tf
```diff
@@ -120,6 +122,7 @@ resource "aws_ecs_task_definition" "app" {
...
```
hectcastro: See justification for this suggestion where the resource is defined below.

```suggestion
  task_role_arn      = "${aws_iam_role.app_task_role.arn}"
```

src/django/api/management/commands/sendtestemail.py
rbreslow: Agreed. I will remove this prior to merging.

deployment/terraform/iam.tf
```diff
@@ -15,10 +18,30 @@ data "aws_iam_policy_document" "ecs_task_execution_assume_role" {
...
+}
+
+resource "aws_iam_role" "ecs_task_role" {
+  name               = "ecs${var.environment}TaskRole"
+  assume_role_policy = "${data.aws_iam_policy_document.ecs_assume_role.json}"
+}
+
+resource "aws_iam_role_policy" "ses_send_email_policy" {
+  name   = "SESSendEmail"
+  role   = "${aws_iam_role.ecs_task_role.name}"
```
rbreslow:
```suggestion
  role   = "${aws_iam_role.app_task_role.name}"
```

deployment/terraform/ses.tf
rbreslow: Me too. Seems to have worked well so far.

deployment/terraform/dns.tf
rbreslow: I tried this and received this error when running `$ terraform plan`:

```bash
Error: Error refreshing state: 1 error(s) occurred:

* aws_route53_record.ses_dkim: aws_route53_record.ses_dkim: value of 'count' cannot be computed
```

When Terraform says "computed" here it means a value that isn't known until its associated resource is created. It works on a subsequent run because the resource has already been created by then and so the value is known.

All the documentation I've found online seems to suggest that there will only ever be 3 DKIM tokens generated, and the example in the Terraform docs hardcodes 3 as well. 

See: 
* https://github.com/awsdocs/amazon-ses-developer-guide/blob/master/doc-source/easy-dkim-dns-records.md
* https://www.terraform.io/docs/providers/aws/r/ses_domain_dkim.html#example-usage

Wondering what your thoughts are.

rbreslow:
@hectcastro Thanks for your feedback Hector. I believe this is ready for another pass.

Issue #180, opened 2019-02-06
Add babel polyfill for IE11


kellyi:
## Overview

The client code is currently missing polyfills for IE11, which means it won't work there:

![screen shot 2019-02-06 at 1 29 12 pm](https://user-images.githubusercontent.com/4165523/52364481-51679980-2a13-11e9-9b25-31a77f11534e.png)

Add the required polyfill so the site works in IE11+

### Expected Behavior

Site should load in IE11

### Actual Behavior

Site fails to load because of some ES6/7 syntax IE11 can't parse.

### Additional context

We typically add the polyfill in our webpack configuration file, but since this is set up in CRA we will have to figure out an alternate way to handle it.

caseycesari:
Is this a good opportunity to use https://babeljs.io/docs/en/babel-preset-env?

kellyi:
Oh, cool -- that's already available in the package.json file but not yet configured, so I'll see about adjusting package.json to use it.

kellyi:
I looked into this a bit more and it seems like the immediate problem is a missing polyfill for `String.prototype.startsWith`, which is required by redux-act.

Issue #177, opened 2019-02-06
Can't download contributor template from Firefox  [Shares point with 176]


jwalgran:
## Overview

The "Download OAR Contributor Template" button does not work in Firefox 64.

### Expected Behavior

When browsing `/contribute` with Firefox 64 clicking the "Download OAR Contributor Template" button causes a CSV file to be downloaded.

### Actual Behavior

When browsing `/contribute` with Firefox 64 clicking the "Download OAR Contributor Template" button does nothing.

### Steps to Reproduce

* Browse `/contribute` with Firefox 64. Login if necessary.
* Click "Download OAR Contributor Template"

Issue #176, opened 2019-02-06
Do not show the email address in the navbar


jwalgran:
## Overview

We should show the `name` field instead of the email.

### Expected Behavior

The `name` of the logged in user is displayed.

### Actual Behavior

The `email` address of the logged in user is displayed

### Steps to Reproduce

* Log in

### Demo

<img width="1000" alt="screen shot 2019-02-05 at 4 53 20 pm" src="https://user-images.githubusercontent.com/17363/52312251-2293e980-2967-11e9-9388-e9936268905e.png">

### Additional context

`NavbarLoginButtonGroup` is the component.

PR #182, opened 2019-02-06
Replace email with user name and enable downloading CSVs across browsers

kellyi:
## Overview

This PR:

- updates the navbar's indication of who is the logged in user to show the name rather than the email addres
- replaces the existing DownloadCSV function implementation with the https://www.npmjs.com/package/file-saver library, which works across browsers and is better tested.

Connects #176 
Connects #177 

## Notes

The app currently isn't working in IE11 due to some missing polyfills, including one for the string `startsWith` method used by redux-act. I briefly looked into fixing this but then made a dedicated card for it: #180 

## Testing Instructions

- get this branch, then `./scripts/update`
- `./scripts/server` then visit the app on 6543 and login
- verify that you see the name rather than the email address displayed in the top right corner
- click on the `contribute` link then click on "Download OAR Contributor Template" and verify that the file downloads correctly

(repeat the last step in Firefox, Chrome, and Safari. IE11 currently won't work, as noted above, but `file-saver` supports IE10+ https://www.npmjs.com/package/file-saver#supported-browsers)

### DIFF
--- a/src/app/src/util/util.js
+++ b/src/app/src/util/util.js
@@ -10,7 +10,9 @@ import negate from 'lodash/negate';
 import omitBy from 'lodash/omitBy';
 import isEmpty from 'lodash/isEmpty';
 import flow from 'lodash/flow';
+import noop from 'lodash/noop';
 import { featureCollection, bbox } from '@turf/turf';
+import { saveAs } from 'file-saver';
 
 import {
     OTHER,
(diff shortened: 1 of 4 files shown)


rajadain:
Taking a look now.

src/app/src/util/util.js
```diff
@@ -20,17 +22,12 @@ import {
...
-    } else {
-        const csvURL = window.URL.createObjectURL(csvData);
-        const tempLink = document.createElement('a');
-        tempLink.href = csvURL;
-        tempLink.setAttribute('download', fileName);
-        tempLink.click();
-    }
+    saveAs(
+        new Blob([data], { type: 'text/csv;charset=utf-8;' }),
+        fileName,
```
rajadain: In testing, when downloading from Chrome, the file had a `.csv` extension, but neither Firefox nor Safari did. Maybe we have to explicitly add it here?

kellyi: Interesting! Looks like we do. I left this function as is -- so callers will have to supply the extension -- but updated it here  77f2948fb0e727f7e66ab700477d27feca33bbf6

PR #184, opened 2019-02-06
Allow Postgres to be accessible from host machine

rajadain:
## Overview

This allows the use of custom Postgres GUI clients, such as [Postico](https://eggerapps.at/postico/), to be used to inspect the database.

## Demo

image

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -17,6 +17,9 @@ Vagrant.configure("2") do |config|
     vb.memory = 4096 # React ran out of memory at 1024
   end
 
+  # Postgres Database
+    config.vm.network :forwarded_port, guest: 5432, host: 5432
+
   # React development server
   config.vm.network :forwarded_port, guest: 6543, host: 6543
 
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -1,5 +1,17 @@
 version: '2.3'
 services:
+  database:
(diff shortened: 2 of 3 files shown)


rbreslow:
@jeancochrane Does this have the potential to effect what was done in https://github.com/open-apparel-registry/open-apparel-registry/pull/171?

rajadain:
I'll wait until we have clarification on the above.

jeancochrane:
Thanks for flagging this @rbreslow, I agree that publishing the database ports in the Compose file will cause collisions on CI.

If isolating database inspection in a separate script would work for dev needs, we could bring up the database with something like `docker-compose run -p 5432:5432 database` in order to temporarily publish the ports. We could also use an override Compose file to similar effect. Happy to help vet ideas if it would be useful for moving this PR along.

rbreslow:
@jeancochrane In an ideal world, we should be able to publish the PostgreSQL service ports without requiring a separate script. How do you feel about reopening azavea/operations#284?

rajadain:
Would adding the `database` service to `docker-compose.ci.yml` without the port revelation take care of this? As done in 94e0229? I suppose this causes some redundancy and we'd need to update the settings in two files instead of one, which is less than ideal.

I'd also be up for having a separate script for making the database inspectable, as suggested above.

rajadain:
Closing this for now. I'll do the changes locally for when I need them.

Issue #162, opened 2019-01-30
Persist filter selections as query string arguments to allow bookmarking/sharing a search


jwalgran:
## Overview

Persist the facility filter choices on the map page in the query string. Restore the filter selections from the query string arguments when the page is loaded. 

### Additional Context

Depends on #159

kellyi:
The app also current has a "Share" button on the map which generates some kind of string -- we should change how that "Share" button works to use the same querystring.

kellyi:
As part of #166 I updated the "Share this search" button on the map to copy `window.location.href` to the clipboard, which means it'll just copy the base url along with any querystring added here.

PR #186, opened 2019-02-06
Sync filter state with querystring parameters

kellyi:
## Overview

This PR enables syncing the selected search filter options with querystring parameters via:

- setting the querystring parameters to reflect the filter selection state when filters have been selected
- setting the filter selection state from the querystring parameters when the user visits the app via a shareable link like:

http://localhost:6543/?name=shoe&contributors=2&contributors=4&contributor_types=Auditor&contributor_types=Union&countries=AO&countries=BY

To accomplish this I created a `withQueryStringSync` higher order component which can wrap whatever other component that will manage keeping the querystring and Redux state in sync. I have it applied to the `MapAndSidebar` component now, but it theoretically could wrap only the actual filter component OR wrap the entire application.

I also made corresponding changes to Redux actions, utility functions, and utility function tests. 

Connects #162 

## Demo

qs-persistence

## Notes

See https://reactjs.org/docs/higher-order-components.html for explanation of higher order components generally. We use them with libraries but haven't commonly written them; in this case it made sense to use that pattern.

Note that there's a brief moment in which the labels for the query parameters are incorrect; this happens because the list of `contributors` and `countries` options comes back from the API, so we do not know what the label is until checking in the returned response. There's also a check to ensure these parameters are valid according to the API's set of valid options.

I set the select, search, and reset inputs to be disabled while fetching valid options from the API btw.

## Testing Instructions

- serve this branch and visit

http://localhost:6543/?name=shoe&contributors=2&contributors=4&contributor_types=Auditor&contributor_types=Union&countries=AO&countries=BY

to verify that the query parameters are set from the string

- click "Reset" to reset the parameters and verify that they stay in sync
- select some options and verify that they are persisted in the querystring
- visit the Register/Login/Contribute page and verify that the querystring params disappear
- click back to return to the main map page and verify that the querystring params reappear
- click "Share this search" to copy the url to the clipboard, then paste it back in the browser and hit return to verify that your selected params are re-hydrated

- visit http://localhost:6543/?name=shoe&countries=Hello World and verify that the invalid country is dropped from the select option -- and removed from the querystring params -- when the valid set of filters returns from the API

### DIFF
--- a/src/app/src/reducers/FiltersReducer.js
+++ b/src/app/src/reducers/FiltersReducer.js
@@ -7,7 +7,14 @@ import {
     updateContributorTypeFilter,
     updateCountryFilter,
     resetAllFilters,
+    updateAllFilters,
 } from '../actions/filters';
+
+import {
+    completeFetchContributorOptions,
+    completeFetchContributorTypeOptions,
+    completeFetchCountryOptions,
+} from '../actions/filterOptions';
 
 const initialState = Object.freeze({
     facilityName: '',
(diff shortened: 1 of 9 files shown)


src/app/src/util/util.js
```diff
@@ -52,7 +55,18 @@ export const makeGetFacilitiesURL = () => '/api/facilities/';
 export const makeGetFacilityByOARIdURL = oarId => `/api/facilities/${oarId}/`;
 export const makeGetFacilitiesURLWithQueryString = qs => `/api/facilities/?${qs}`;
 
-export const getValueFromObject = ({ value }) => value;
+
+// Get a value from a one of the React Select filter option objects if it also
+// has a label. (Values without labels were hydrated from the query-string but
+// not subsequently confirmed by the API-sent list of available options, so we
+// discard them.
+export const maybeGetValueFromFilterObject = ({ value, label = null }) => {
```
kellyi: I think this change can be dropped, as it was from an iteration of this where I was trying to distinguish selections staged from the querystring params and those validated by the choices returned from the API by not providing a label for the staged ones. I updated it later to use the value as the label until the selection had been validated, then replace it.

src/app/src/reducers/FiltersReducer.js
```diff
@@ -7,15 +7,109 @@ import {
...
+    if (!state.contributorTypes.length) {
+        return state;
+    }
+
+    // filter out any options set from the querystring which turn out
+    // not to be valid according to the API's response
+    const updatedContributorTypes = state
+        .contributorTypes
+        .reduce((accumulator, { value }) => {
+            const validOption = payload
```
kellyi: The implementation of these reduce functions is nearly the same, but each of them needs access to the distinct `payload` array, so it wasn't possible to extract a function like `(acc, { value }) => acc.concat(value);` or similar to share across them.

jwalgran:
Looking at this now

jwalgran:
I was expecting the search to be executed automatically when refreshing the page with query options selected. Consider implementing that behavior, or creating a follow up issue if you feel it is too far out of scope of the original issue.

kellyi:
> I was expecting the search to be executed automatically when refreshing the page with query options selected.

Oh! This should be trivial to add, so I'll see about setting it up.

kellyi:
>     * Showing nothing and a spinner while things load.

Sounds good to me. I added that along with the immediate search in 
669292a 

re: immediate search: it seemed to me that this should only happen on the first visit to the app in a single session, and not every time the component unmounts/remounts within the same session. I added a check, therefore, to only have the facilities immediately load if:

- there is a querystring with parameters to try to fetch
- there are no filters already set in the Redux store, as there would be in the case of navigating from screen to screen within the same session

I also added some logic to clear the facilities every time the querystring sync component unmounts in order to prevent the set of displayed facilities from being unaccountably out of sync with the visible set of search filters in cases when the user:
   - searches for and retrieves a set of facilities
   - changes the filter options and updates the query string parameters
   without undertaking a new search
   - visits a different page, then revisits the map and search page

kellyi:
Awesome, thanks for your help with this! Going to squash and then merge once the tests pass on CI.

Issue #188, opened 2019-02-07
Exception when loading the Django admin


jwalgran:
## Overview

Cannot load the Django admin

### Expected Behavior

Browsing `/admin/` shows a list of models.

### Actual Behavior

An unhandled exception prevents that page from loading

### Steps to Reproduce

* Browse `/admin/` login

### Additional context

```
django_1    | [2019-02-07 18:59:37 +0000] [8] [DEBUG] GET /admin/
django_1    | Internal Server Error: /admin/
django_1    | Traceback (most recent call last):
django_1    |   File "/usr/local/lib/python3.7/site-packages/django/core/handlers/exception.py", line 35, in inner
django_1    |     response = get_response(request)
django_1    |   File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 158, in _get_response
django_1    |     response = self.process_exception_by_middleware(e, request)
django_1    |   File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 156, in _get_response
django_1    |     response = response.render()
django_1    |   File "/usr/local/lib/python3.7/site-packages/django/template/response.py", line 106, in render
django_1    |     self.content = self.rendered_content
django_1    |   File "/usr/local/lib/python3.7/site-packages/django/template/response.py", line 83, in rendered_content
django_1    |     content = template.render(context, self._request)
django_1    |   File "/usr/local/lib/python3.7/site-packages/django/template/backends/django.py", line 61, in render
django_1    |     return self.template.render(context)
django_1    |   File "/usr/local/lib/python3.7/site-packages/django/template/base.py", line 175, in render
django_1    |     return self._render(context)
django_1    |   File "/usr/local/lib/python3.7/site-packages/django/template/base.py", line 167, in _render
django_1    |     return self.nodelist.render(context)
django_1    |   File "/usr/local/lib/python3.7/site-packages/django/template/base.py", line 943, in render
django_1    |     bit = node.render_annotated(context)
django_1    |   File "/usr/local/lib/python3.7/site-packages/django/template/base.py", line 910, in render_annotated
django_1    |     return self.render(context)
django_1    |   File "/usr/local/lib/python3.7/site-packages/django/template/loader_tags.py", line 155, in render
django_1    |     return compiled_parent._render(context)
django_1    |   File "/usr/local/lib/python3.7/site-packages/django/template/base.py", line 167, in _render
django_1    |     return self.nodelist.render(context)
[...]

PR #191, opened 2019-02-07
Add support for Rollbar on Django

### LINKED ISSUE #62: Add support for Rollbar on Django
See: https://docs.rollbar.com/docs/python/

rbreslow:
## Overview

* Wire up Rollbar for Django (https://docs.rollbar.com/docs/python/#section-django)
* Add Rollbar [deploy notifications](https://docs.rollbar.com/reference#post-deploy)
* Build container images before `cibuild`

I included what I did out-of-band in the parent issue.

Closes #62 

## Testing Instructions

First, I had Jenkins cycle my changes onto staging:

```bash
$ git push -u origin feature/jrb/django-rollbar:test/jrb/django-rollbar
```

See deploy notification in #oar Slack channel:

image

Next, I triggered #188 by logging into https://staging.openapparel.org/admin:

image

https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/3/

### DIFF
--- a/scripts/cibuild
+++ b/scripts/cibuild
@@ -23,6 +23,9 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     if [ "${1:-}" = "--help" ]; then
         usage
     else
+        # Ensure container images are current
+        docker-compose build
+
         # Install frontend NPM modules
         docker-compose \
             run --rm --no-deps app \
--- a/scripts/cibuild
+++ b/scripts/cibuild
@@ -42,6 +42,6 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         # Make sure the database is up for Django tests
         docker-compose up -d database
 
-        ./scripts/test
+        ./scripts/test --ci
     fi
 fi
--- a/scripts/cibuild
+++ b/scripts/cibuild
@@ -23,18 +23,10 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     if [ "${1:-}" = "--help" ]; then
         usage
     else
-        # Ensure container images are current
-        docker-compose build
-
         # Install frontend NPM modules
         docker-compose \
             run --rm --no-deps app \
             yarn install
-
-        # Make sure the database is up for Django tests
-        docker-compose up -d database
-
-        ./scripts/test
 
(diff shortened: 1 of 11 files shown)


scripts/cibuild
```diff
@@ -23,6 +23,9 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     if [ "${1:-}" = "--help" ]; then
         usage
     else
+        # Ensure container images are current
```
rbreslow: Since the Python dependencies listed in `requirements.txt` are baked into the container, and the container image isn't rebuilt before running tests on CI, tests were failing because the `rollbar` module was missing.

hectcastro: When I tested this locally, the image produced by this command without `-f docker-compose.ci.yml` is tagged with `:latest`. It seems like we want the image built by this command to be tagged with the Git commit SHA, so that we can avoid scenarios were `:latest` is being built by one CI job, but tested by another.

It is possible that we need to tweak the body of `test_django` in `test` to test with the appropriate image in CI (but also the appropriate image when run locally).

rbreslow: What if we just build the tagged `django` container image twice?

rbreslow:
@hectcastro This is ready for another pass.

I think my solution for the disconnect between image references works -- but I'm open to (and looking for) a more elegant solution.

src/django/oar/settings.py
```diff
@@ -223,3 +225,12 @@
...
         'Invalid GOOGLE_GEOCODING_API_KEY provided, must be set')
+
+if not DEBUG:
+    ROLLBAR = {
+        'access_token': os.getenv('ROLLBAR_SERVER_SIDE_ACCESS_TOKEN'),
+        'environment': ENVIRONMENT.lower(),
+        'root': BASE_DIR,
+    }
+    import rollbar
+    rollbar.init(**ROLLBAR)
```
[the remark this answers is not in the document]
hectcastro: Odd discrepancy. I hope they aren't doing this differently here just to support a `manage.py shell` example.

Either way, it looks like the most recent version included support for calling `init()` multiple times (before it used to emit a warning). Given that, I think we're probably OK to keep it as is.

PR #194, opened 2019-02-07
Avoid a crash in Django admin by specifying a custom OarUserAdmin

jwalgran:
## Overview

The built-in admin forms were looking for fields and values that do not exist on our custom User model. Adding a customer admin class allows us to specify the fields that should be excluded and the fields that should appear on the form.

Connects #188

## Demo

screen shot 2019-02-07 at 3 14 07 pm

screen shot 2019-02-07 at 3 14 23 pm

screen shot 2019-02-07 at 3 14 32 pm

## Testing Instructions

* Load fixtures if they have not been loaded already.
* Browse `/admin` and log in as `admin@openapparel.org`.
* Verify that the user-related pages are browsable.

### DIFF
--- a/src/django/api/models.py
+++ b/src/django/api/models.py
@@ -139,13 +139,13 @@ class User(AbstractBaseUser, PermissionsMixin):
     updated_at = models.DateTimeField(auto_now=True)
 
     def __str__(self):
-        return self.name
+        return self.email
 
     def get_full_name(self):
-        return self.name
+        return self.email
 
     def get_short_name(self):
-        return self.name
+        return self.email
 
     def first_name(self):
         pass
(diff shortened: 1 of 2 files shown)


src/django/api/models.py
```diff
@@ -139,13 +139,13 @@ class User(AbstractBaseUser, PermissionsMixin):
     updated_at = models.DateTimeField(auto_now=True)
 
     def __str__(self):
-        return self.name
+        return self.username or ''
```
kellyi: Since we're using email addresses for usernames -- and `username` will theoretically always be unset -- would it make sense for these to be `self.email`?

jwalgran: I was concerned about spreading the email address around but after thinking about it I don't think that concern is warranted. We should only be implicitly converting `User` to string in shell or admin context, so email will be more helpful.

jwalgran: Rebased with this change.

jwalgran:
Thanks for the review and the smart suggestion.

Issue #178, opened 2019-02-06
Implement facility details sidebar and corresponding details map view


kellyi:
## Overview

Implement the Facility Details Sidebar component added in #166 and also adjust how the map view works based on the details view.

### Describe the solution you'd like

#166 adds a base Facility Details Sidebar component that lives on a `/facilities/:oarID` route and will retrieve geojson data for the facility with the given `:oarID`, either from the local store (if it's available there) or from the API if it's not already available client-side.

Currently the details sidebar view just displays the facility's geojson, but we should adjust it to have the data displayed as it was in the legacy https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/components/FactoryInfo.jsx component.

We should also update the map view to work with the details view. It seems like the desired behavior for the map is:

- on the details view, only the single detail facility is shown
- if the user has arrived on the details view by clicking on a facility in the main map view and then clicks the back button, the prior search state, prior facilities state, and prior map bbox state are restored
- if the user has arrived on the details view via a direct link -- and there is no prior map view state -- the map and search are just set to their usual initial states

kellyi:
In addition, the details view should also display a list of referenced contributor source lists.

kellyi:
I reread the legacy Map.jsx component and it seems like the map behavior when the details sidebar is shown is to keep the prior set of points but to highlight the current point:

```
// Highlight the selected point
                // Update the filter in the highlighted-point layer to only show
                // the matching state, thus making a hover effect.
                this.map.setFilter('highlighted-point', [
                    '==',
                    'uniqueId',
                    p.target.id,
]);
```

PR #195, opened 2019-02-08
Processing results are a list

jwalgran:
## Overview

Converts `FacilityListItem.processing_results` from an object to a list. `processing_results` now acts as a historical log of processing events and allows the same processing action to potentially be retried in the case of a failure without losing and previously logged information.

Connects #193

## Demo

```
openapparelregistry=# select jsonb_pretty(processing_results) from api_facilitylistitem where facility_list_id = 2 and row_index = 0;
                                                    jsonb_pretty
--------------------------------------------------------------------------------------------------------------------
 [                                                                                                                 +
     {                                                                                                             +
         "error": false,                                                                                           +
         "action": "parse",                                                                                        +
         "started_at": "2019-02-08 20:28:34.461622",                                                               +
         "finished_at": "2019-02-08 20:28:34.462310"                                                               +
     },                                                                                                            +
     {                                                                                                             +
         "data": {                                                                                                 +
             "status": "OK",                                                                                       +
             "results": [                                                                                          +
                 {                                                                                                 +
                     "types": [                                                                                    +
                         "street_address"                                                                          +
                     ],                                                                                            +
                     "geometry": {                                                                                 +
                         "location": {                                                                             +
                             "lat": 27.966814,                                                                     +
                             "lng": 120.66201                                                                      +
                         },                                                                                        +
                         "viewport": {                                                                             +
                             "northeast": {                                                                        +
                                 "lat": 27.9681629802915,                                                          +
                                 "lng": 120.6633589802915                                                          +
                             },                                                                                    +
                             "southwest": {                                                                        +
                                 "lat": 27.9654650197085,                                                          +
                                 "lng": 120.6606610197085                                                          +
                             }                                                                                     +
                         },                                                                                        +
                         "location_type": "ROOFTOP"                                                                +
                     },                                                                                            +
                     "place_id": "ChIJ52tPIaZjTzQRPPPUsIeA2pE",                                                    +
                     "plus_code": {                                                                                +
                         "global_code": "7QV2XM86+PR",                                                             +
                         "compound_code": "XM86+PR Ouhai, Wenzhou, Zhejiang, China"                                +
                     },                                                                                            +
                     "partial_match": true,                                                                        +
                     "formatted_address": "8 Shuangbao W Rd, Ouhai Qu, Wenzhou Shi, Zhejiang Sheng, China, 325000",+
                     "address_components": [                                                                       +
                         {                                                                                         +
                             "types": [                                                                            +
                                 "street_number"                                                                   +
                             ],                                                                                    +
[...]

PR #196, opened 2019-02-08
Implement Facility Details View

kellyi:
## Overview

This PR implements the facility details view by adapting the previous FactoryInfo component into the FacilityDetailSidebar component. If a facility has any other names or addresses, these are now returned from the API in the GeoJSON properties. (Otherwise, these fields are empty lists).

Likewise, the set of contributors is also listed in the new sidebar as depicted below. Had a brief discussion with @mmcfarland in which we decided to show both the contributor name and also the name of the list on which the matching facility list item appeared -- this is depicted in the static image below.

The second commit here makes some updates to the `makefixtures` command to create lists and items with non-pending statuses. Additionally, I manually adjusted the facility list items connected to 

http://localhost:6543/facilities/176

to make it possible to demo the functionality.

Happy to keep or to drop these fixture changes, but the details component is set up in such a way that it won't render anything for empty lists for other names, other addresses, or contributors and the serializer's logic requires the whole chain have a specific set of statuses.

Connects #178
Connects #81

## Demo

screen shot 2019-02-11 at 4 07 44 pm

facility-details

## Notes

As described in a comment on #178 the prior map interactions seemed to keep all data but just change the highlighted marker color when selecting a facility. I've kept that here, and also added logic to set the main facility map data from the details data in cases when the user arrives directly at the details page.

I didn't otherwise do much with the map interactions and I think it makes sense to defer working on it more until after the 2.0 release, although if there are any quick-to-implement suggestions about how to make it work a little better, I'd be open to making such changes.

## Testing

- get this branch, then `./scripts/manage flush` and `./scripts/manage loadfixtures`
- `./scripts/manage server` and visit 6543
- search from facilities and then click through to the details page to verify that the details view works as expected

Given the current state of the fixtures, most facilities probably won't have alternate address or name data, and many may have no contributor data. However, if you visit

http://localhost:6543/facilities/176

directly with the set of fixtures added here, then you'll be able to see how that rendering code works.


src/app/src/components/FacilityDetailSidebar.jsx
```diff
@@ -49,46 +68,119 @@ class FacilityDetailSidebar extends Component {
...
+                        </h1>
+                        <p key="_id" className="control-panel__body">
+                            {data.properties.oar_id}
+                        </p>
+                    </div>
+                    <div className="control-panel__group">
+                        <h1 className="control-panel__heading">
+                            GPS Coordinates:
+                        </h1>
+                        <p key="geo" className="control-panel__body">
```
kellyi: These `key` props aren't hurting anything by remaining, probably, but they could also potentially be removed.

jwalgran:

src/django/api/management/commands/makefixtures.py
```diff
@@ -182,14 +187,16 @@ def make_facility_lists(max_id=14):
 
 def make_facility_list_item(list_pk, item_pk, row_index, raw_data):
     (created_at, updated_at) = make_created_updated()
+    status = random.choice([s for (s, _) in FacilityListItem.STATUS_CHOICES])
```
jwalgran: We were running into problems testing our batch processing in #195 because the fixture data was inconsistent with the way data will actually appear in the database during normal usage.

Any random choices we make when building fixtures need to be threaded through the FacilityListItem->FacilityMatch->Facility relationship correctly.

Here is the commit where I fixed this as part of #195

https://github.com/open-apparel-registry/open-apparel-registry/pull/195/commits/1ad3ede81c867094f50cd5abee75413d66c05b4b

kellyi: Cool. Should that make it so there is some match data in the existing set of fixtures to render? If so, I can adjust the changes for this second commit.

jwalgran: I think you will need to extend the idea. The version in #195 results in only having `UPLOADED` and `MATCHED` list items. To get some things in the in between states, we will just have to make sure that the `status` of the `FacilityListItem` is set to `POTENTIAL_MATCH` when there is a related `FacilityMatch`.

kellyi: Got it, that makes sense. I'll adapt it to work with this.

src/app/src/components/OARMap.jsx
```diff
@@ -113,26 +116,28 @@ class OARMap extends Component {
             fetching,
             error,
             data,
+            singleFacilityData,
         } = this.props;
 
         if (fetching) {
-            return null;
+            return noop();
```
jwalgran: Why are we returning the result of calling noop here? Why not return `undefined` or keep the `null` return?

src/django/api/serializers.py
```diff
@@ -68,13 +73,83 @@ class Meta:
...
+        ]
+
+        return {
+            match.name
+            for match
+            in facility_list_item_matches
+            if len(match.name) != 0
+            and match.name is not None
+            and match.name != facility.name
+            and match.facility_list.is_active
```
jwalgran: This is some important logic. Since we are doing "soft deletes" we need to make sure that we are confident that we will never incorrectly show that a contributor is associated with a facility. It looks right, but I think we should consider adding some test coverage around it. If it is easier, we could move the bulk of the logic out to a method on the `FacilityListItem` model.

kellyi:
I rebased this on develop and dropped the commit that had an initial change to the fixture data, so I'll create a new version of that along with the set of other fixes to come.

src/app/src/components/OARMap.jsx
kellyi: I think I used this once and then decided to replace all of `null` with `noop`. However, for consistency's sake it'd be better to standardize on returning null here.

kellyi: Changed back to null in 868953dfba001c3275bfe3b1193ad80a98adee00

src/app/src/components/FacilityDetailSidebarInfo.jsx
```diff
@@ -0,0 +1,34 @@
+import React, { memo } from 'react';
+import { arrayOf, string } from 'prop-types';
+
+import ShowOnly from './ShowOnly';
+
+const FacilityDetailSidebarInfo = memo(({
```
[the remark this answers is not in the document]
kellyi: Good question: it appears that it only does a shallow compare with complex objects (e.g. arrays) -- https://reactjs.org/docs/react-api.html -- but it also takes an optional equality check function, so I'll add one.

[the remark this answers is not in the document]
kellyi: Done in e4a7fb23ebe1366f9cd0f9a9e67630b004ad3f8a

kellyi:
Removed the unused key props in 49f35811acf9c8068c91fad80787c5ab9092ea53

src/app/src/components/FacilityDetailSidebar.jsx
kellyi: Dropped in 49f35811acf9c8068c91fad80787c5ab9092ea53

src/django/api/serializers.py
kellyi: Sounds good. I'll make these testable and add some unit tests around them.

kellyi: I moved these methods into the Facility model and then created a suite of unit tests in af190436461150f9473860f2f5b202204e489d00

kellyi:
c03027f92e320bc4998bcbf401cd5b004681ae87 adjusts the makefixtures command to create some facilities which have multiple matches, and to give those matches different names and periodically make their lists public and active. I also checked in the new set of fixtures for testing, so you can just install them in place of the previous set.

Not every facility will have every field displayed (and this will likely be the case in the production data), but here's a representative sample of what the outputs look like:

screen shot 2019-02-14 at 3 51 27 pm
screen shot 2019-02-14 at 3 51 49 pm
screen shot 2019-02-14 at 3 51 40 pm
screen shot 2019-02-14 at 3 51 34 pm

To test this, load the new fixtures, then serve the app and check out the following facilities:

- 39
- 350
- 339
- 413

jwalgran:
Looking at this now.

src/django/api/management/commands/makefixtures.py
```diff
@@ -168,8 +168,8 @@ def make_facility_list(pk, organization_pk=None):
             'name': name,
             'file_name': name + '.csv',
             'header': 'country,name,address',
-            'is_active': make_bool(),
-            'is_public': make_bool(),
+            'is_active': make_bool(pk % 2 == 0),
```
jwalgran: I like the use of `%` to make the generator more repeatable.

kellyi:
> Tests and fixture generator updates look great. Thanks.

Great! Thanks for your help with this. Going to merge!

PR #197, opened 2019-02-08
Add script for running one-off management commands on ECS

### LINKED ISSUE #167: Add STRTA for running one-off tasks against the ECS cluster
Add a STRTA script that has the ability to identify the most recent CLI task definition, then creates an instance of it on the appropriate ECS cluster.

(Useful for running migrations and other one-off management command tasks.)

- Space separated arguments passed to deployment/scripts/manage are correctly passed as the command override (may need to be comma separated)
- A full url to the launched task is printed to the console i.e.
https://console.aws.amazon.com/ecs/home?region=us-east-1#/clusters/ecsStagingCluster/tasks/GUID-d853-4774-a97a-6e226cf690b6/details

jeancochrane:
## Overview

Add an `ecsmanage` management command to allow running one-off Django management commands as ECS tasks on staging or production.

Closes #167.

## Demo

screen shot 2019-02-12 at 12 27 35 pm

## Notes

* We had discussed a few possible approaches to this command, including pure shell AWS CLI and Boto. The command-line parsing wound up being complex enough that I went with a Django management command using Boto for the sake of simplicity.

* The `-e|--env` flag controls the environment, so we can run e.g. production migrations like so:

```
$ ./scripts/manage ecsmanage --env production migrate
```

## Testing Instructions

* Shell into the VM
* Run `./scripts/manage ecsmanage help`, the equivalent of `python manage.py help`
  * Note that this will run on the staging environment by default
* Click on the link to the details page and confirm that it takes you to the right place
* On the details page, navigate to `Containers > django > Log Configuration`, open up the logs, and confirm that the command exited successfully


deployment/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '2.3'
+services:
+  manage:
+    build: ./scripts/mgmt
```
jeancochrane: I don't love the directory structure here. Since the Boto script requires a custom Dockerfile + requirements file + Python script in addition to `./deployment/scripts/manage`, I wanted to package them all up in a separate directory to avoid tossing a bunch of cruft directly in `./deployment/scripts`. I wound up going with `./deployment/scripts/mgmt/`, which is a sibling to `./deployment/scripts/manage`. Better ideas for directory structure are welcome!

deployment/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+    volumes:
+      - ./scripts/mgmt:/usr/local/src/
+      - $HOME/.aws:/root/.aws:ro
+    environment:
+      - AWS_PROFILE=${AWS_PROFILE:-open-apparel-registry}
+    working_dir: /usr/local/src
+    entrypoint:
+      - python3
+      - manage.py
```
jeancochrane: `manage.py` in this case refers to the custom Boto script, not the `django-admin` script. If this naming scheme is confusing, I can change it.

deployment/scripts/mgmt/manage.py
```diff
@@ -0,0 +1,176 @@
...
+            overrides=overrides,
+            networkConfiguration=network_configuration,
+            count=1,
+            launchType='FARGATE'
+        )
+
+        task = self.parse_response(task_response, 'tasks', 0)
+
+        # Task ARNS use the format arn:aws:ecs:<region>:<aws_account_id>:task/<id>
+        task_id = task['taskArn'].split('/')[1]
```
jeancochrane: This isn't super pretty, but [Boto doesn't return a task ID from `client.run_task`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Client.run_task), so the choice was between substring parsing and an extra Boto call just to get the task ID.

deployment/scripts/mgmt/manage.py
```diff
@@ -0,0 +1,176 @@
...
+        type=str,
+        choices=['staging', 'production'],
+        default='staging',
+        help="Environment to run the task in (staging or production)"
+    )
+
+    parser.add_argument(
+        'cmd',
+        type=str,
+        nargs='+',
```
jeancochrane: `nargs='+'` tells argparse to expect one or more arguments, assign them to the `cmd` variable as a list, and raise an error if the argument is missing.

rajadain:
What would be the syntax to run this for production? How do we differentiate between the two target environments?

jeancochrane:
Sorry about that @rajadain, I forgot to document that behavior in the PR body. Updated it with some explanation:

> * The `-e|--env` flag controls the environment, so we can run e.g. production migrations like so:
>
>```
>$ ./deployment/scripts/manage --env production migrate
>``` 

Staging is the default, so the command in the testing instructions is equivalent to `./deployment/scripts/manage --env staging migrate`.

hectcastro:
I have a high level comment before digging into the PR details:

How bad would it be if we tucked this functionality into a Django management command? It feels a little awkward because we'd be tightly coupling the Django application with its execution environment, but doing so allows us to remove a bunch of the scaffolding in this PR.

It is also possible that we'll need to use Boto3 within Django (for user image uploads?), which would cancel out having it as an extra dependency now.

cc: @jwalgran

jwalgran:
We will almost certainly be using boto3 for submitting batch jobs, and I think that we will need it for our SES email backend. No problem there.

I'm not opposed to the idea. Our theoretical "wrapper" management command could use [`call_command`](https://docs.djangoproject.com/en/2.0/ref/django-admin/#django.core.management.call_command) to launch the actual command. Should be one-liner to parse the options passed to the script and pass them on to the `*args` parameter of `call_command`.

src/django/api/management/commands/remote.py
```diff
@@ -0,0 +1,164 @@
...
+            overrides=overrides,
+            networkConfiguration=network_configuration,
+            count=1,
+            launchType='FARGATE'
+        )
+
+        task = self.parse_response(task_response, 'tasks', 0)
+
+        # Task ARNS use the format arn:aws:ecs:<region>:<aws_account_id>:task/<id>
+        task_id = task['taskArn'].split('/')[1]
```
jeancochrane: This isn't super pretty, but [Boto doesn't return a task ID from `client.run_task`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Client.run_task), so the choice was between substring parsing and an extra Boto call just to get the task ID.

src/django/api/management/commands/remote.py
```diff
@@ -0,0 +1,164 @@
+import boto3
+from django.core.management.base import BaseCommand
+
+class Command(BaseCommand):
+    help = 'Run a one-off management command on an ECS cluster.'
```
jeancochrane: I don't know how I feel about `remote` as the name of this command. It was the best I could come up with in less than 5 minutes; if you have anything better, let me know!

jwalgran: `ecsmanage` is another suggestion that incorporates the fact that it is still running a management command and that the command will be running on an ECS instance.

jeancochrane: That seems much clearer, I'll go with that.

jwalgran:
I think this is best signed off on by @hectcastro since he has more experience in ECS best practices.

src/django/api/management/commands/ecsmanage.py
```diff
@@ -0,0 +1,170 @@
...
+
+        task_id = self.run_task(task_def_arn,
+                                security_group_id,
+                                subnet_id,
+                                cmd)
+
+        url = 'https://console.aws.amazon.com/ecs/home?region=us-east-1#'
+        url += f'/clusters/ecs{self.env}Cluster/tasks/{task_id}/details'
+
+        print('Task started! View status here:')
```
hectcastro: See [here](https://docs.djangoproject.com/en/2.1/howto/custom-management-commands/) for recommendations on how to use `self.{stdout,stderr}` to write output from a management command.

src/django/api/management/commands/ecsmanage.py
```diff
@@ -0,0 +1,170 @@
...
+        print()
+        print(url)
+
+    def parse_response(self, response, key, idx=None):
+        """
+        Perform a key-value lookup on a response from the AWS CLI, wrapping it
```
hectcastro: s/AWS CLI/AWS API/

src/django/api/management/commands/ecsmanage.py
```diff
@@ -0,0 +1,170 @@
...
+        security_group_id = self.get_security_group()
+        subnet_id = self.get_subnet()
+
+        task_id = self.run_task(task_def_arn,
+                                security_group_id,
+                                subnet_id,
+                                cmd)
+
+        url = 'https://console.aws.amazon.com/ecs/home?region=us-east-1#'
+        url += f'/clusters/ecs{self.env}Cluster/tasks/{task_id}/details'
```
hectcastro: Did not know about this new `f` business. I'd consider using the multiline approach described below vs. `+=`.

See: https://realpython.com/python-f-strings/#multiline-f-strings

src/django/api/management/commands/ecsmanage.py
```diff
@@ -0,0 +1,170 @@
...
+
+        print('Task started! View status here:')
+        print()
+        print(url)
+
+    def parse_response(self, response, key, idx=None):
```
hectcastro: Nice utility function.

src/django/api/management/commands/ecsmanage.py
jeancochrane: Didn't know about the multiline f-strings! Makes me like it even more.

src/django/api/management/commands/ecsmanage.py
jeancochrane: Good catch, I forgot about `self.stdout`.

PR #199, opened 2019-02-08
Add corejs library for use in IE11 polyfills

kellyi:
## Overview

Add corejs library which enables importing polyfills for IE11 on a
method-by-method basis, and to adjust which polyfills are imported while
the development server is running.

I chose this solution as an alternative to using `@babel/polyfill` which
unfortunately caused a 10x increase in the time it took for the client
dev server to start, seemingly because it was parsing all the code in
node_modules for every polyfill.

corejs also lets you import additional polyfills on the fly, so it seemed like a lightweight way to
help the set of necessary polyfills keep up with whatever syntax we were using.

See https://github.com/zloirock/core-js/tree/v2

Connects #180 

## Demo

screen shot 2019-02-08 at 5 16 57 pm

## Testing Instructions

- get this branch and `./scripts/update`
- visit localhost:6543 in FF, Chrome, Safari and verify the app works as before in each
- visit the app in IE11 and verify that it also works

### DIFF
--- a/src/app/src/index.js
+++ b/src/app/src/index.js
@@ -1,3 +1,11 @@
+// polyfills for IE11
+// catalog of other is available at https://github.com/zloirock/core-js/tree/v2
+import 'core-js/fn/string/starts-with';
+import 'core-js/fn/object/values';
+import 'core-js/fn/array/find';
+import 'core-js/fn/array/find-index';
+import 'core-js/fn/object/assign';
+
 import React from 'react';
 import { render } from 'react-dom';
 import { Provider } from 'react-redux';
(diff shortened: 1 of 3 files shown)


src/app/src/index.js
```diff
@@ -1,3 +1,10 @@
+// polyfills for IE11
+// catalog of other is available at https://github.com/zloirock/core-js/tree/v2
+import 'core-js/fn/string/starts-with';
+import 'core-js/fn/object/values';
+import 'core-js/fn/array/find';
+import 'core-js/fn/array/find-index';
+
```
kellyi: Note for me: we should probably also add one for Object.assign

jwalgran:
Looking at this now

Issue #192, opened 2019-02-07
Allow matches to be approved on the facility list detail page / show match results


jwalgran:
## Overview

When a `FacilityListItem` has the status of `POSSIBLE_MATCH`, show the related `FacilityMatch` rows and allow the user to select one or reject them all. If they are all rejected, create a new `Facility`.

### Describe the solution you'd like

  - On the list detail page, show rows with `POSSIBLE_MATCH` status in an "expanded" state. List the possible match(s) underneath with "Accept" and "Reject" buttons.
     - If "Accept" is clicked, POST to an endpoint. Update state to show other matches as rejected.
     - If "Reject" is clicked on all options, POST to an endpoint and create a new `Facility`.
  - Allow any row that is not in process to be "expanded."
    - If the item is matched, show the `Facility` row to which it is matched.
    - If the item is in the error state, show the error. 

### Additional Context

The previous matching UI

<img width="1132" alt="screen shot 2019-02-07 at 1 54 15 pm" src="https://user-images.githubusercontent.com/17363/52442247-e637c900-2adf-11e9-83de-20c62d56e81e.png">

https://docs.google.com/presentation/d/1vfEb1QnVD-9bvfQyMtYUpXQT3ogJvkQvURI7QoK4z7s/edit#slide=id.g4c5179bf71_0_5

kellyi:
Blocked by #190

kellyi:
API side is mostly done for #206, aside from figuring out how to generate some non-automatic matches using the existing naive matching algorithm.

Frontend's partly done, although I'll need to make some adjustments to the UI:

![screen shot 2019-02-19 at 5 37 02 pm](https://user-images.githubusercontent.com/4165523/53052595-2185b080-346d-11e9-89e9-458b1b8bfbdf.png)

Some remaining stuff on #206:

- [ ] create row components for displaying errors
- [ ] create row components for displaying automatic matches
- [ ] clean up confirm/match UI
- [ ] interleave confirm/reject API response data back into Redux facilityDetails data

PR #206, opened 2019-02-14
Create facility list items table interactions

kellyi:
## Overview

This PR accomplishes a few things:

- updates the API to handle confirming or rejecting potential facility matches
- adds tests for API confirm/reject logic
- updates the UI to add several interactions to the facility list items list table, including:
   - showing toggleable detail rows for list items in the matched or error states, along with additional details
    - showing a confirm/reject match UI along with a confirmation dialog to use for confirming or rejecting matches
- adds links to a user's facilities on the main list page and also in the navigation bar dropdown
- update the match functionality to create some random matches if the list item id % 4 === 0. This is done because the previous naive matching function was setting everything with 100% confidence

Connects #192

## Demo

confirm-match

reject-matches

## Testing

- get this branch
- resetdb
- serve the app, then register for and log in with a new account
- upload a list from the set of lists committed to the repo (pick one of the larger ones to ensure lots of matches)
- in the VM, run the following commands to process the list:

```
./scripts/manage batch_process --action parse --list-id 15
./scripts/manage batch_process --action geocode --list-id 15
./scripts/manage batch_process --action match --list-id 15
```

- visit your lists page in the UI, then click on the list to go to the details page
- try the following interactions:
   - clicking on an ERROR or MATCH or CONFIRMED_MATCH row should expand that row to display details, and close all other non-CONFIRM details rows
   - clicking on an expanded non-CONFIRM detail row should cause it to collapse back to a normal table row
    - confirming a match through the UI should update that table row to display it as a CONFIRMED_MATCH with a link to the facility that has matched
    - rejecting one of several matches should keep the confirm row open, but show the rejected match as rejected
    - rejecting the last potential match for a list item should create a new facility then update the details row to show it in CONFIRMED_MATCH state

- run `./scripts/test` to verify that they pass


kellyi:
Just so I'll have it ready at hand, here's the command for parsing the first list uploaded after the initial fixture data is in:

./scripts/manage batch_process --action parse --list-id 15

src/django/api/tests.py
```diff
@@ -181,7 +181,8 @@ def test_cant_replace_twice(self):
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
         self.assertEqual(
             json.loads(response.content),
-            ['FacilityList 1 has already been replaced.'])
+            ['FacilityList {} has already been replaced.'
+                .format(response_json['id'])])
```
kellyi: Not sure why this was breaking, but it kept trying to use "FacilityList 15 has already been", which made it seem like it was uploading the new list after loading the non-test fixture data.

jwalgran:
Looking at this now

src/app/src/components/FacilityListItemsMatchTableRow.jsx
```diff
@@ -0,0 +1,121 @@
...
+            />
+        </TableCell>
+        <TableCell
+            padding="none"
+            style={listTableCellStyles.nameCellStyles}
+        >
+            <FacilityListItemsDetailedTableRowCell
+                title={name || ' '}
+                subtitle="Facility Match Name"
+                hsIsHidden={false}
```
jwalgran: Typo? hs -> hr?

src/app/src/components/FacilityListItemsConfirmationTableRow.jsx
```diff
@@ -0,0 +1,174 @@
...
+                />
+            </TableCell>
+            <TableCell
+                padding="none"
+                style={listTableCellStyles.nameCellStyles}
+            >
+                <FacilityListItemsDetailedTableRowCell
+                    title={item.name || ' '}
+                    subtitle="Matched Name"
+                    hsIsHidden={false}
```
jwalgran: Typo? hs -> hr?

src/app/src/components/FacilityListItemsConfirmationTableRow.jsx
```diff
@@ -0,0 +1,174 @@
...
+                />
+            </TableCell>
+            <TableCell
+                padding="none"
+                style={listTableCellStyles.statusCellStyles}
+            >
+                <FacilityListItemsDetailedTableRowCell
+                    title={item.status}
+                    subtitle="Actions"
+                    hsIsHidden={false}
```
jwalgran: Typo? hs -> hr?

src/app/src/components/FacilityListItemsMatchTableRow.jsx
kellyi: Excellent catch! Will squash in a fixup to address.

kellyi:
squashed in a fixup to replace the `hs` prefix with `hr`, then tested everything out again. Everything's still working so will merge once the tests pass.

Thanks for your help with this!

Issue #203, opened 2019-02-14
Support Front-end Environment Variables


rajadain:
## Overview

In different environments, such as development, test, staging, and production, the app may need to behave differently or connect to different third party services (e.g. Rollbar, Mapbox, etc.). Currently there is no way to achieve this, since the environment variables are baked in to a Docker image at compile-time, not run-time. This organization requires compiling different images for different environments, since the environment variables can only be injected at compile-time. This limits are confidence in the parity of staging and production (and other environments), leading to potential bugs not being caught until a later environment is launched.

### Describe the solution you'd like

A code organization that would read variables from the environment and use them to populate values in the front-end. This should behave the same way in _all_ environments: development, test, staging, and production.

At least one of the environment variables should be the name of the environment, so it can be used by the app to potentially flag certain behaviors.

The way environment variables are specified should ensure that there is a value for that variable in every environment. We shouldn't discover after a deploy that a particular environment doesn't have a variable set.

kellyi:
Not sure whether "code organization" implies changing the dir structure and project setup of the app container, but it appears that this is one way Create React App recommends for injecting server side variables: https://facebook.github.io/create-react-app/docs/title-and-meta-tags#injecting-data-from-the-server-into-the-page in a way that might minimize the number of changes to the client setup.

rajadain:
That looks really promising! I'd like to do this with a small footprint. Trying that out now.

rajadain:
Okay, I don't think the link above is enough, because we still need server infrastructure in place to replace the `__SERVER_DATA__` placeholder. That would require adding middleware / views to Django.

However, once we've done that, we'd need to switch to using the Django rendered endpoint in development (rather than hitting webpack-dev-server directly), so we have access to those environment variables.

This is accomplished by using [`django-webpack-loader`](https://github.com/owais/django-webpack-loader) and [`webpack-bundle-tracker`](https://github.com/owais/webpack-bundle-tracker), as done in https://github.com/project-icp/bee-pollinator-app/pull/289/commits/ee6791b21f7803c01159cef65126dc86704d7233 and https://github.com/project-icp/bee-pollinator-app/pull/289/commits/d89c46b30b6042d074130ca79758e332a8f511cc.

[...]

Issue #64, opened 2019-01-09
Add Rollbar for client


kellyi:
## Overview

We should configure Rollbar for the client, as we've done with a few other applications.

### Describe the solution you'd like

Configure Rollbar to report client errors and have sourcemaps.

### Additional Context

When we've configured this in other applications, it was done via adjusting the Webpack setup. However, since this is a Create React App created app, it may be more circuitous to set up Rollbar.

Depends on #62

PR #207, opened 2019-02-15
Support Front-end Environment Variables, Rollbar

rajadain:
## Overview

Adds a mechanism to allow for Django to inject environment variables into the front-end at run-time. These will override any set by react-scripts at compile-time. Usage will fallback to compile-time variables if a run-time version is not found. In the front-end, instead of using `process.env.REACT_APP_VARIABLE_NAME`, we should now use `env('REACT_APP_VARIABLE_NAME')`.

Also removes `django-spa`, which was a thin wrapper around an old version of `whitenoise`, with a more recent version of `whitenoise`. This is configured to serve static files found under `/static/` from the root URL `/`, mimicking the behavior of `django-spa`. Accessing the root directly `/` now goes through `oar.urls` and is handled by a new `web` app's `views.index`, which renders the react-scripts generated file with run-time variables injected. Accessing a URL not recognized by Django is assumed to be a front-end route, just as `django-spa` does, and is passed through to the front-end.

By convention, any environment variable available to Django that starts with `REACT_APP_` will be injected in to the front-end. Adding new environment variables is possible, although they have to be wired through many steps in Terraform, as done in the final commit.

Because these variables are populated via Django, to test them locally one must run `cibuild` and then go to [:8081/](http://localhost:8081/). When viewing the output of react-scripts at [:6543/](http://localhost:6543/) directly, these variables will not be available. However, their absence should not interfere with the development experience.

See commit messages for details.

Connects #64 
Connects #203 

## Demo

image

image

image

## Notes

Notably, the variables are only injected when the page is rendered via Django. In the current setup, WhiteNoise receives requests first, and if it can find a corresponding static file to return it will do so immediately. It is only those requests that do not match a static file that are sent on to Django. This means that while any request to `/` _will_ have the environment variables injected, any request to `/index.html` _will not_. I have not been able to think of a way around this short of removing WhiteNoise entirely, but perhaps this is an acceptable compromise at this stage.

## Testing Instructions

* Check out this branch
* Edit your `.env.backend` file to include the key `REACT_APP_ROLLBAR_POST_CLIENT_ITEM_ACCESS_TOKEN` with the `post_client_item` access token from Rollbar
* Run `scripts/update`, then `scripts/cibuild`, then `scripts/server`
* Go to [:8081/](http://localhost:8081/) and open the developer tools
* Trigger an error from the console

    ```
    window.onerror("TestError: Hello world", window.location.href)
    ```

* Log in to Rollbar and ensure you see the error there
* Go to [:6543/](http://localhost:6543/) and ensure development and hot-reloading work just as before, despite an extra error in the console for the invalid HTML of the template.


deployment/terraform/variables.tf
```diff
@@ -58,6 +58,8 @@ variable "google_geocoding_api_key" {}
 
 variable "rollbar_server_side_access_token" {}
 
+variable "react_app_rollbar_post_client_item_access_token" {}
```
hectcastro: I think that it would be good to consider renaming this Terraform variable to something like `rollbar_client_side_access_token` for consistency with `rollbar_server_side_access_token`. We can leave the environment variable itself with the `REACT_APP_` prefix.

rajadain:
Thank you for reviewing and all the suggestions! I've opened #208 with all the recommended changes, and ensured the problems mentioned here are addressed.

PR #208, opened 2019-02-15
Support Front-end Environment Variables, Rollbar 2

rajadain:
## Overview

An alternative to #207, that instead of replacing Django-SPA with WhiteNoise, renames the generated entry file to `main.html` instead of `index.html`. This (rather brilliant) suggestion by @hectcastro allows us to bypass Django-SPA's default behavior and use `oar.urls` to render the page.

We still need to add a custom 404 handler, as by default Django-SPA tries to send non-matching routes to `static/index.html` which no longer exists.

This also preserves the behavior of `/admin/` which was broken in #207.

Supersedes #207 
Connects #64
Connects #203

## Motivation

The main reason to do this instead of having a settings endpoint that is queried after page load is that anything that requires an environment variable will have to check if the variable has loaded before it can function. As the number of environment variables grows, so will the unnecessary asynchronous code in the front-end. By ensuring that all variables are present at load, we simplify current and future code.

## Notes

The replacement of a 404 handler with a Django view potentially exposes us to DDoS attacks. However, if there are CDNs in front of the Django container, we should be okay.

~React errors are not caught because of [Error Boundaries in React 16](https://rollbar.com/blog/react-error-boundaries/). I'm going to make a follow up issue for fixing those.~ Fixed in a10ff7e.

## Testing Instructions

* Run the steps as in #207
* Ensure that [:8081/admin/](http://localhost:8081/admin/) loads and works correctly


kellyi:
> React errors are not caught because of [Error Boundaries in React 16](https://rollbar.com/blog/react-error-boundaries/). I'm going to make a follow up issue for fixing those.

Out of curiosity, can you add a screenshot of them not being caught? The entire App is wrapped in an ErrorBoundary https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/App.jsx#L49 which I believe has been catching things previously, although it may be misconfigured.

rajadain:
Here's a screenshot:

image

Here's the error I introduced:

```diff
diff --git a/src/app/src/components/FilterSidebarSearchTab.jsx b/src/app/src/components/FilterSidebarSearchTab.jsx
index 07b6f21..c08d832 100644
--- a/src/app/src/components/FilterSidebarSearchTab.jsx
+++ b/src/app/src/components/FilterSidebarSearchTab.jsx
@@ -63,6 +63,7 @@ function FilterSidebarSearchTab({
     facilities,
     fetchingOptions,
 }) {
+    boom(); // eslint-disable-line
     if (fetchingOptions) {
         return (
             <div className="control-panel__content">
```

It is a little contrived, as that would actually be caught by ESLint had it not been for an explicit disabling, but still.

Here's a screenshot from Chrome with slightly more detail:

image

kellyi:
Took a look and I think you can just add the code from the Rollbar blog post...

```
componentDidCatch(error, errorInfo) {    
 Rollbar.error(error); // Send it to Rollbar!
}
```
here:

https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/components/ErrorBoundary.jsx#L26

rajadain:
That worked! Thanks @kellyi. Fixed in a10ff7e.

kellyi:
Now that https://github.com/open-apparel-registry/open-apparel-registry/commit/a10ff7e17c17d34cb5ff4bc47422101288511b82 just calls Rollbar directly within the existing code, and not in the HTML header, I still think the case is pretty strong for getting these env var dynamically from one or several API endpoints and putting them in Redux state or just on the window here.

Getting them dynamically from endpoints would circumvent needing to make changes the existing Django SPA setup. While it may be too early to project how this number would grow, the number of environment variables the client currently needs is just 2 in total.

That also matches how the previous error logging which was done via Raven and Sentry within the application code, which we removed in 40 in anticipation of adding Rollbar later: https://github.com/open-apparel-registry/open-apparel-registry/pull/40

They app already depends on quite a bit of asynchrony for its normal functioning, so it feels like one or two more async requests (along with the code to handle them) doesn't add too much complexity.

rajadain:
Alright, thanks for taking a look. I think we've reached the end of this line of implementation, and should switch to using a dynamic endpoint, as suggested above. Thankfully, most of the commits herein can be reused. I'll get a new PR up shortly.

Issue #149, opened 2019-01-24
Launch background processing tasks that run FacilityListItem processing functions


jwalgran:
## Overview

Once a background task processor has been provisioned, launch processing tasks to run FacilityListItems through parsing, geocoding, and matching.

### Describe the solution you'd like

* Queue an Array Job to `parse` each `FacilityListItem` created from an uploaded CSV
* Use the ID of the `parse` Job to queue a dependent Array Job to `geocode` each `FacilityListItem` 
* Use the ID of the `geocode` Job to queue a dependent Array Job to `match` each `FacilityListItem` 

Depends on #59
Depends on #158

PR #210, opened 2019-02-19
WIP Process with batch

jwalgran:
Connects #149


jwalgran:
I'm running into a chicken-and-egg problem with provisioning new ECR repositories.

The `Jenkinsfile` runs `cipublish` before `infra`, which fails because Terraform has not had a chance to create the new ECR repositories to which the images should be published. I am going to attempt to comment out the push to the new repos to see if it will allow Terraform to apply the infrastructure changes.

PR #212, opened 2019-02-20
Support Front-end Environment Variables, Rollbar 3

rajadain:
## Overview

Configures Rollbar, adds support of injecting environment variables at run-time. If there are variables in Django's environment that start with `REACT_APP_`, they will be added to front-end app on load.

Supersedes #208
Supersedes #207
Connects #64
Connects #203

### Demo

image

## Testing Instructions

Follow the instructions in #207 and #208.


jwalgran:
We should wait to merge this until after our client demo.

kellyi:
Setting this up to take a look!

rajadain:
Thanks for taking a look, and for giving me the room to come to this solution!

I [just added](https://github.com/open-apparel-registry/open-apparel-registry/compare/822209ea17dc29a832a34d6a15452934393cb675..d0cfb9c75d4c2a0b92241657f0c3cbfb8bc04021) an `Object.freeze` to the `ENVIRONMENT` object. Will merge once green.

Issue #211, opened 2019-02-19
Match processing attempts to process the CSV header row


kellyi:
## Overview

While working on #206 I noticed that the first row for any CSV I uploaded would end up in an error state if I ran the match processing all the way through on the list:

<img width="338" alt="screen shot 2019-02-19 at 4 57 45 pm" src="https://user-images.githubusercontent.com/4165523/53050782-1da35f80-3468-11e9-9f5b-b2b19509d75b.png">

### Expected Behavior

CSV parsing should discard the header for the match processing.

### Actual Behavior

CSV parsing doesn't always discard the header for match processing

### Steps to Reproduce

Upload a list through the app, then run the processing script through the parse, geocode, and match steps and verify that the first row in the UI is in the error state depicted above.

rajadain:
I notice that here, we add all lines in the CSV file as FacilityListItems:

https://github.com/open-apparel-registry/open-apparel-registry/blob/76a1ea9b20b199ddb9ebc658d624964adc41818a/src/django/api/views.py#L338-L344

This causes the header row to be added to `api_facilitylistitem` table:

```sql
SELECT id, row_index, raw_data, status, facility_list_id
FROM api_facilitylistitem
WHERE row_index = 0;

 id | row_index |       raw_data       |  status  | facility_list_id
----+-----------+----------------------+----------+------------------
  1 |         0 | country,name,address | ERROR    |                1
 83 |         0 | country,name,address | UPLOADED |                2
(2 rows)
```

I can skip that one line like this:

```python
for idx, line in enumerate(csv_file):
    if idx > 0:
        new_item = FacilityListItem(
            row_index=idx,
            facility_list=new_list,
            raw_data=line.decode().rstrip()
        )
        new_item.save()
```

One effect of this will be that the `row_index` field will start at `1`. This can be adjusted to start at `0`, but then it won't directly match the line number of the input CSV, which will continue to include a header row. Which is preferable?

jwalgran:
I think we should adjust the value so that `row_index` continues to start from zero. AWS Batch array jobs feed us a zero based index.

kellyi:
> I think we should adjust the value so that `row_index` continues to start from zero. AWS Batch array jobs feed us a zero based index

[...]

PR #213, opened 2019-02-21
Process line items with AWS Batch

jwalgran:
## Overview

Add a second container for running AWS Batch jobs and add related provisioning and application code to submit jobs that process uploaded CSVs using our existing `batch_process` management command.

Connects #149 

## Demo

screen shot 2019-02-20 at 6 59 59 pm

screen shot 2019-02-20 at 7 00 11 pm

screen shot 2019-02-20 at 6 59 44 pm

## Notes

Batch does not allow us to override the `ENTRYPOINT` of a container, so we have created a thin wrapper that strips off the gunicorn `ENTRYPOINT` inherited from our Django base container. It is a little wasteful storage-wise to push 2 copies of the same container but we decided it was preferable to do that rather than not use our Azavea base Django container.

We are opting to not rename the `openapparelregistry` repo/image to `openapparelregistry-app` to avoid having terraform destroy the existing repository while this is change is being tested in a shared staging environment.

When the container is run as a normal Django web app process in Staging or Production we check the instance metadata to properly set the `ALLOWED_HOSTS` Django setting. This instance metadata check fails when the container is run in AWS Batch so we have added a the `BATCH_MODE` environment variable which acts as a switch to disable the instance metadata check.

Jobs are only submitted in staging an production because AWS Batch cannot run in development.

This code assumes that we have a single job queue and we will always want to run the latest version of the job definition. This may change in the future if we expand our processing but for now those assumptions make for simpler code.

We record the job IDs in the `FacilityListItem.processing_results`. It is a bit redundant to log the same ID on all the line items but it is consistent with the fact that all processing is done at the item level and allows us to display the entire processing history of an item without needing to collate detail from the item and its parent list.

There is no way to resubmit jobs if the process fails. In this initial version the user can work around that type failure by reuploading the same CSV and using the "replaces" feature.

We needed to add an additional security policy to the app IAM role to allow listing Batch resources and submitting jobs.

## Testing Instructions

This is only testable on staging. I pushed a copy of this branch as `test/process-with-batch`. Jenkins will deploy a branch prefixed with `test/` (At the time this PR was opened, Jenkins jobs are failing)

* Log in as a user other than the admin included in the fixture data.
* Browse https://staging.openapparel.org/contribute and upload a CSV
* Wait (it can sometimes take a while for compute resources to start processing)
  * You can browse the processing status in the AWS console https://console.aws.amazon.com/batch/home?region=us-east-1#/jobs
* Browse the uploaded list at https://staging.openapparel.org/lists/{list-id}. Verify that as items are processed, their status changes.

### DIFF
--- a/deployment/terraform/batch.tf
+++ b/deployment/terraform/batch.tf
@@ -51,7 +51,7 @@ data "template_file" "default_job_definition" {
   template = "${file("job-definitions/default.json")}"
 
   vars {
-    image_url = "${module.ecr_repository.repository_url}:${var.image_tag}"
+    image_url = "${module.ecr_repository_batch.repository_url}:${var.image_tag}"
 
     postgres_host     = "${aws_route53_record.database.name}"
     postgres_port     = "${module.database.port}"
--- a/deployment/terraform/container_service.tf
+++ b/deployment/terraform/container_service.tf
@@ -1,5 +1,5 @@
 locals {
-  app_image = "${module.ecr_repository.repository_url}:${var.image_tag}"
+  app_image = "${module.ecr_repository_app.repository_url}:${var.image_tag}"
 }
 
 #
--- a/deployment/terraform/iam.tf
+++ b/deployment/terraform/iam.tf
@@ -46,6 +46,28 @@ resource "aws_iam_role_policy" "ses_send_email" {
   policy = "${data.aws_iam_policy_document.ses_send_email.json}"
 }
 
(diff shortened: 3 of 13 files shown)


jwalgran:
>One thing that didn't quite work as I expected was that one of that Batch jobs tried to parse the CSV header and failed.

Thanks. Kelly spotted this and opened up #211 to fix it.

PR #222, opened 2019-02-21
Get mapbox token at compile time

kellyi:
## Overview

The mapbox token was unset when trying to get it at runtime, so return
to getting it at compile time.

Note: this essentially just reverts the mapbox key part of the Rollbar changes.

Connects #223 

## Testing Instructions

- get this branch, then test that the mapbox map works on both server and cibuild versions

### DIFF
--- a/src/app/src/util/constants.oarmap.js
+++ b/src/app/src/util/constants.oarmap.js
@@ -1,6 +1,4 @@
-import env from './env';
-
-export const MAPBOX_TOKEN = env('REACT_APP_MAPBOX_TOKEN');
+export const MAPBOX_TOKEN = process.env.REACT_APP_MAPBOX_TOKEN;
 
 export const CIRCLE_COLOR = '#1A237E';
 export const CIRCLE_TEXT_COLOR = '#fff';
--- a/src/app/src/util/env.js
+++ b/src/app/src/util/env.js
@@ -1,7 +1,0 @@
-export default function (v) {
-    if (window.ENVIRONMENT && v in window.ENVIRONMENT) {
-        return window.ENVIRONMENT[v];
-    }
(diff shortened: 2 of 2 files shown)


jwalgran:
I was not able to replicate #223 on develop. I added this `console.log` statement
 to the main map component

```diff
diff --git a/src/app/src/components/OARMap.jsx b/src/app/src/components/OARMap.jsx
index c2c74dd..36df095 100644
--- a/src/app/src/components/OARMap.jsx
+++ b/src/app/src/components/OARMap.jsx
@@ -30,6 +30,7 @@ import {
 } from '../util/constants.oarmap';
 
 mapboxgl.accessToken = MAPBOX_TOKEN;
+window.console.log('MAPBOX_TOKEN', MAPBOX_TOKEN);
 
 const OARMapStyles = Object.freeze({
     copySearchButtonStyle: Object.freeze({
```

After `./scripts/update` and `./scripts/server` I was seeing the key correctly logged on both http://localhost:6543/ and http://localhost:8081

Were you able to reproduce the original issue?

kellyi:
> Were you able to reproduce the original issue?

Interesting. I thought @rajadain and I had replicated it yesterday but maybe we were doing something incorrectly.

If it's not actually unset, though, I can just close this PR rather than trying to take more time investigating. If it turns out to be a problem the change will be easy to make.

jwalgran:
The app crashes if the key is not set

screen shot 2019-02-22 at 3 24 58 pm

Since staging is working correctly, I do think we should just close this PR.

rajadain:
Huh. This was a wild ride. Still not sure what's exactly going on, but happy that the environment variable fallback technique is working.

PR #229, opened 2019-02-25
Update bug report template to apply bug label

jwalgran:
## Overview

Having the label makes it easier to filter the issue list.

## Testing Instructions

I don't think there is a way to test this without merging the change and then testing through the Github UI.


jwalgran:
It worked.

screen shot 2019-02-25 at 12 01 38 pm

Issue #187, opened 2019-02-07
Handle multiple facilities at the same point/address with a disambiguator and clustered marker


kellyi:
## Overview

Handle multiple facilities at the same point/address with a disambiguator and clustered marker.

### Describe the solution you'd like

Previously multiple facilities which have been geocoded to the same point but deemed different facilities had a clustered marker and a way to disambiguate them. That functionality was not yet ported over when connecting the client to the new API, so we should restore it.

kellyi:
Here's the legacy Map.jsx component code:

https://gist.github.com/kellyi/2d45df6297e40314663afe8a13a37ed6

I think the code for creating the disambiguation popup starts here:

https://gist.github.com/kellyi/2d45df6297e40314663afe8a13a37ed6#file-map-jsx-L144

PR #230, opened 2019-02-25
Add disambiguation popup for multiple facilities

kellyi:
## Overview

Implement a disambiguation popup which is displayed when a marker from a
set of search results has multiple facilities at the same point.

Connects #187 

## Demo

facilities-popup

## Notes

~I still need to set up some viable test data for this, so I'll push another commit to do so~ I updated the `resetdb` command to do this at the end.

## Testing Instructions

- get this branch, then run `./scripts/manage resetdb`
- search for facilities, then click on the facilities at `Point(0, 0)` and verify that it zooms in and displays the popup as depicted in the demo gif
- selecting a facility from the list should link you to its details page
- verify that you can toggle through each of the multiple facilities and be linked to their details page
- verify that you can dismiss and re-create the marker
- verify that the markers for single facilities continue to work as before

### DIFF
--- a/src/app/src/components/OARMapPopup.jsx
+++ b/src/app/src/components/OARMapPopup.jsx
@@ -1,0 +1,128 @@
+import React from 'react';
+import { createPortal } from 'react-dom';
+import { arrayOf, shape, string } from 'prop-types';
+import { Link } from 'react-router-dom';
+import List from '@material-ui/core/List';
+import ListItem from '@material-ui/core/ListItem';
+import ListSubheader from '@material-ui/core/ListSubheader';
+import ListItemIcon from '@material-ui/core/ListItemIcon';
+import LinkIcon from '@material-ui/icons/Link';
+import ListItemText from '@material-ui/core/ListItemText';
+
+import { makeFacilityDetailLink } from '../util/util';
+import COLOURS from '../util/COLOURS';
+
+const OARMapPopupStyles = Object.freeze({
+    selectedListItemStyles: Object.freeze({
+        backgroundColor: COLOURS.NAVY_BLUE,
+        maxWidth: '500px',
+    }),
+    unselectedListItemStyles: Object.freeze({
+        maxWidth: '500px',
+    }),
+    linkStyles: Object.freeze({
+        textDecoration: 'none',
+        display: 'flex',
+    }),
+    unselectedLinkStyles: Object.freeze({
+        textDecoration: 'none',
+        display: 'flex',
+    }),
+    selectedItemTextStyles: Object.freeze({
+        color: COLOURS.WHITE,
+    }),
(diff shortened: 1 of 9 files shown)


src/app/src/components/OARMapPopup.jsx
```diff
@@ -0,0 +1,125 @@
...
+                            .map(({
+                                properties: {
+                                    address,
+                                    name,
+                                    oar_id: oarID,
+                                },
+                            }) => (
+                                <ListItem
+                                    className="popup-item display-flex notranslate"
+                                    key={oarID} // es
```
kellyi: Reminder to myself to remove the rest of this comment.

src/django/api/management/commands/resetdb.py
```diff
@@ -9,3 +12,11 @@ def handle(self, *args, **options):
         call_command('reset_schema', '--noinput')
         call_command('migrate')
         call_command('loadfixtures')
+
+        facility_one = Facility.objects.first()
```
jwalgran: Consider moving this code (or something similar to `makefixtures`. `resetdb` was intended to just be a helper that runs 3 commands in sequence. Ensuring that we have test data that allows us to exercise the UI is more the responsibility of the fixture.

jwalgran: Django querysets are lazy, so you could optionally use a slightly shorter version

```python
for f in Facility.objects.all().order_by('id')[:2]:
    f.location = Point(0, 0)
    f.save()
```

kellyi: Cool, I'll move this out of resetdb and into makefixtures.

kellyi:
Going to give the popup list items a `maxWidth` property to ensure long strings get wrapped:

screen shot 2019-02-26 at 2 10 11 pm

src/django/api/management/commands/resetdb.py
kellyi: I ended up changing the mock facility creation code to:

```
def make_facility(facility_item):
    (created_at, updated_at) = make_created_updated()
    fields = facility_item['fields']
    # A workaround to have the CSV parser work on a single row
    for row in csv.reader([fields['raw_data']]):
        parsed = row

    pk = facility_item['pk']
    point_is_fixed = pk % 70 == 0

    return {
        'model': 'api.facility',
        'pk': pk,
        'fields': {
            'name': parsed[1],
            'address': parsed[2],
            'country_code': COUNTRY_CODES[parsed[0].upper()],
            'location': make_point(point_is_fixed),
            'created_from': pk,
            'created_at': created_at,
            'updated_at': updated_at,
        }
    }
```

and also changed `make_point` to make it return a point at 0 0 if the boolean True was passed in as an argument.

Issue #231, opened 2019-02-25
Password reset links use http rather than https on staging


jwalgran:
## Overview

Password reset links use `http://` rather than `https://` on staging. There is a redirect in place, so these links are not broken, but we want to avoid sending authentication related requests in the clear.

### Expected Behavior

URLs in password reset email use the `https://` protocol.

### Actual Behavior

URLs in password reset email use the `http://` protocol.

### Steps to Reproduce

* Ensure you have a registered account with an email account you can check.
* Log out.
* Request a password reset email.

### Demo

<img width="922" alt="screen shot 2019-02-25 at 4 43 57 pm" src="https://user-images.githubusercontent.com/17363/53376757-b597c200-391c-11e9-9350-98c012958648.png">

kellyi:
Shares points with #97

Issue #97, opened 2019-01-17
Add "Forgot Password" workflow


kellyi:
## Overview

Following #87 we should implement a "Forgot Password" workflow.

### Describe the solution you'd like

Currently the login screen has a button to open a dialog for entering an email address to which to send a "forgot password" link. Likewise, the Redux action for sending an email exists in actions/auth.

For this feature, we should

- wire up the dialog & Redux action to make a request to django-rest-auth's provided forgot password link
- have Django send an email with a reset link to the user's email address
- add a react router path, like `/auth/forgot?token=xyz...`, which will display a mini form component for resetting the password, with password and confirmation inputs
- add a Redux action to make the request back to Django to reset the password
- on successfully resetting, redirect the user to `/auth/login`

PR #232, opened 2019-02-26
Set `use_https` correctly for password reset email

kellyi:
## Overview

Adjust the `use_https` setting for the password reset email to False
in development and True in production. This effectively changes the
`protocol` setting used in the base PasswordResetForm's `save` method:

https://github.com/django/django/blob/master/django/contrib/auth/forms.py#L290

Connects #231
Connects #97 

## Demo

Using `'use_https': settings.ENVIRONMENT != 'Development',`:

screen shot 2019-02-26 at 10 49 47 am

Using `'use_https': settings.ENVIRONMENT == 'Development',`:

screen shot 2019-02-26 at 10 48 54 am

## Testing Instructions

- serve this branch then use the password reset form to send a reset email and verify that it uses http
- adjust the `use_https` line updated in this changeset to be

```
'use_https': settings.ENVIRONMENT == 'Development',
```

... then reload the server and send another password reset email to verify that the email now uses https

### DIFF
--- a/src/django/api/serializers.py
+++ b/src/django/api/serializers.py
@@ -205,7 +205,7 @@ class UserPasswordResetSerializer(PasswordResetSerializer):
             domain_override = request.get_host()
 
         opts = {
-            'use_https': request.is_secure(),
+            'use_https': settings.ENVIRONMENT != 'Development',
             'domain_override': domain_override,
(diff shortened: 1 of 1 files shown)


kellyi:
Cool, thanks for taking a look!

Issue #215, opened 2019-02-21
Show on the list detail page when an item is the source of a facility


jwalgran:
## Overview

There is currently no distinction between a list item being matched to an existing facility and a list item creating a new facility. Add a designation to list items from which facilities were created.

### Describe the solution you'd like

* Display "New Facility" in the Status column
* The expanded view is redundant. Instead, render the facility name as a link on the main row.

PR #235, opened 2019-02-27
Indicate when facility list item is facility source

kellyi:
## Overview

- update the list item table to indicate whether a facility list item is
a facility source, displaying a `NEW_FACILITY` status if so

Connects #215 

## Demo

screen shot 2019-02-27 at 10 14 53 am

## Testing Instructions

- serve this branch, then upload and process a facility list using the instructions in https://github.com/open-apparel-registry/open-apparel-registry/pull/206
- refresh the list items page for the now-processed list and verify that facility list items which have led to newly created facilities have static rows with `NEW_FACILITY` displayed as their status and that they are linked to the facilities detail page for the facility

### DIFF
--- a/src/app/src/components/FacilityListItemsTable.jsx
+++ b/src/app/src/components/FacilityListItemsTable.jsx
@@ -105,6 +105,24 @@ function FacilityListItemsTable({
                 );
             }
 
+            if (item.status === facilityListItemStatusChoicesEnum.MATCHED
+                && item.id === item.matched_facility.created_from_id) {
+                return (
+                    <FacilityListItemsTableRow
+                        key={item.row_index}
+                        rowIndex={item.row_index}
+                        countryName={item.country_name}
+                        name={item.name}
+                        address={item.address}
+                        status={item.status}
+                        handleSelectRow={handleSelectRow}
+                        hover={false}
+                        newFacility
+                        oarID={item.matched_facility.oar_id}
+                    />
+                );
+            }
+
             if (item.row_index !== selectedFacilityListItemsRowIndex) {
                 return (
                     <FacilityListItemsTableRow
--- a/src/app/src/components/FacilityListItemsTableRow.jsx
+++ b/src/app/src/components/FacilityListItemsTableRow.jsx
@@ -1,5 +1,6 @@
 import React, { memo } from 'react';
 import { bool, func, number, oneOfType, string } from 'prop-types';
+import { Link } from 'react-router-dom';
 import TableRow from '@material-ui/core/TableRow';
 import TableCell from '@material-ui/core/TableCell';
 import isEqual from 'lodash/isEqual';
@@ -8,13 +9,17 @@ import { facilityListItemStatusPropType } from '../util/propTypes';
 
 import { listTableCellStyles } from '../util/styles';
 
+import { makeFacilityDetailLink } from '../util/util';
+
 const propsAreEqual = (prevProps, nextProps) =>
     isEqual(prevProps.rowIndex, nextProps.rowIndex)
     && isEqual(prevProps.countryCode, nextProps.countryCode)
     && isEqual(prevProps.name, nextProps.name)
     && isEqual(prevProps.address, nextProps.address)
     && isEqual(prevProps.status, nextProps.status)
-    && isEqual(prevProps.hover, nextProps.hover);
+    && isEqual(prevProps.hover, nextProps.hover)
+    && isEqual(prevProps.newFacility, nextProps.newFacility)
+    && isEqual(prevProps.oarID, nextProps.oarID);
 
 const FacilityListItemsTableRow = memo(({
     rowIndex,
@@ -24,6 +29,8 @@ const FacilityListItemsTableRow = memo(({
     status,
     hover,
     handleSelectRow,
+    newFacility,
+    oarID,
 }) => (
     <TableRow
         hover={hover}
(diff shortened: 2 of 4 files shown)


src/app/src/components/FacilityListItemsTable.jsx
```diff
@@ -105,6 +105,24 @@ function FacilityListItemsTable({
...
+                return (
+                    <FacilityListItemsTableRow
+                        key={item.row_index}
+                        rowIndex={item.row_index}
+                        countryName={item.country_name}
+                        name={item.name}
+                        address={item.address}
+                        status={item.status}
+                        handleSelectRow={handleSelectRow}
+                        hover
```
kellyi: Might be worth dropping `hover` here since the other places that's set are for rows which are interactive, and can be expanded on click.

src/app/src/components/FacilityListItemsTableRow.jsx
```diff
@@ -64,14 +81,20 @@ const FacilityListItemsTableRow = memo(({
...
+                    : status
+            }
         </TableCell>
     </TableRow>
 ), propsAreEqual);
 
 FacilityListItemsTableRow.defaultProps = {
     hover: false,
     handleSelectRow: null,
+    newFacility: false,
```
rajadain: Is `{ newFacility: true, oarID: null }` a legitimate possibility? If not, we can probably just use `oarID` as the only field, with `oarID === null` implying `newFacility === true`.

src/app/src/components/FacilityListItemsTableRow.jsx
```diff
@@ -52,7 +59,17 @@ const FacilityListItemsTableRow = memo(({
             padding="none"
             style={listTableCellStyles.nameCellStyles}
         >
-            {name}
+            {
+                (newFacility && oarID)
+                    ? (
+                        <Link
+                            to={makeFacilityDetailLink(oarID)}
+                            href={makeFacilityDetailLink(oarID)}
```
rajadain: Is `href` here specified as a fallback?

src/app/src/components/FacilityListItemsTableRow.jsx
kellyi: If it does happen it would be due to a bug in the API, but I split these props to ensure that the view would be resilient to having a small problem in the data.

However, presumably every facility should always have an OAR ID.

src/app/src/components/FacilityListItemsTableRow.jsx
```diff
@@ -64,14 +81,20 @@ const FacilityListItemsTableRow = memo(({
             padding="none"
             style={listTableCellStyles.statusCellStyles}
         >
-            {status}
+            {
+                newFacility
+                    ? 'NEW_FACILITY'
```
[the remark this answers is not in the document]
kellyi: cc: @jwalgran but my impression was that this issue was entirely about displaying the data and not about changing the underlying model.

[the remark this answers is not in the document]
kellyi: (At present the backend can always figure out `NEW_FACILITY` status on its own through the same comparisons btw).

src/app/src/components/FacilityListItemsTableRow.jsx
kellyi: `href`'s added here because Eslint requires placing it on these `Link` tags, although does feel redundant to have both props.

src/app/src/components/FacilityListItemsTableRow.jsx
rajadain: > However, presumably every facility should always have an OAR ID.

Is there any way to enforce that?

kellyi: I suppose it's conceivably enforced by being the primary key of the `Facility` model:

https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/django/api/models.py#L321

The rhyming prop type for the list items marks it as `isRequired` but that's only checked at runtime: https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/util/propTypes.js#L81

I guess I'd imagine that the API would 500 on the serializer crashing here, too:

https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/django/api/serializers.py#L177

rajadain: Given the above enforcement, should we still keep the two fields? On one hand, it slightly improves readability. On the other, it implies a connection that could be missed when touching this code in the future.

kellyi: Happy to do either. My intention here was to separate out two things:

- we only want the view to create a link to the facility when it's both a new facility and there's an OAR ID to link to
- we only want to display NEW_FACILITY status when it's a new facility (but it doesn't matter if there's an OAR ID)

Currently both of these will be in place, so we could rely on both of them being there, but fusing them would overload what OAR ID is supposed indicate for the view and also couple the view more closely to the API, which is I think what informed deciding to split them.

However, on balance it's not a huge deal either way.

src/app/src/components/FacilityListItemsTableRow.jsx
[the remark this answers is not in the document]
jwalgran: It is just a display issue. We do not want to add a new status.

src/app/src/components/FacilityListItemsTableRow.jsx
rajadain: Ah that makes sense. By having a separate field we ensure that if OAR ID is changed in the future, this behavior won't be affected. Let's keep the split then.

kellyi:
> All my questions have been answered above. This is good to go!

Cool, thanks for helping me clarify how this is supposed to work!

Before I merge this I'm going to squash in a new commit to change these `NEW_FACILITY` rows to `hover={false}` to make them appear non-interactive.

Issue #185, opened 2019-02-06
Rename Organization model to Contributor


kellyi:
Edited after discussion

## Overview

We currently have an `Organization` model but the UI refers to this entity as a "Contributor." Contributors to the OAR are not always organizations. Rename the `Organization` model to `Contributor`.

~Currently the React client uses `contributor` to refer what we've adapted into the `organization` model, but the API uses `organization`. We should normalize on organization.~

### Describe the solution you'd like

- Rename the `Organization` model to `Contributor`.
- Rename related attributes.
- Rebuild related fixtures
- Update client to use `contributor` rather than `organization` where it isn't already

~Update client-side form field labels, variable names, et cetera to use `organization` rather than `contributor` when necessary. Adjust the API to match.~

jwalgran:
@katieashaw I would like to get your opinion on this. Internally, we have started using "Organization" to refer to a brand, retailer, MSI, etc. The application interface currently uses "Contributor" when referring to these entities. Do you anticipate any our stakeholders having a problem with using the word "Organization?"

katieashaw:
My only slight hesitation is that individuals, in certain circumstances, might upload to the site - e.g. a researcher, student or journalist. @nataliegrillon Do you have a view on this?

nataliegrillon:
agreed @katieashaw - data contributors will be potentially more than just organizations....

jwalgran:
Thanks for the feedback. It sounds like we should reverse this issue and change our code to refer to Contributor internally so we are consistently using the same name.

PR #245, opened 2019-03-04
Rename "Organization" to "Contributor"

rajadain:
## Overview

Since individuals such as researchers, students, or journalists may contribute to the site, in addition to organizations such as a brand, retailer, manufacturer, etc., we switch to using the more inclusive term "Contributor" throughout the code.

This also brings the back-end terminology in line with what the UI already uses.

organization → contributor
org → contrib

Connects #185 

## Demo

```sql
\d api_contributor;
                                           Table "public.api_contributor"
       Column       |           Type           | Collation | Nullable |                   Default
--------------------+--------------------------+-----------+----------+---------------------------------------------
 id                 | integer                  |           | not null | nextval('api_contributor_id_seq'::regclass)
 name               | character varying(200)   |           | not null |
 description        | text                     |           | not null |
 website            | character varying(200)   |           | not null |
 contrib_type       | character varying(200)   |           | not null |
 other_contrib_type | character varying(200)   |           |          |
 created_at         | timestamp with time zone |           | not null |
 updated_at         | timestamp with time zone |           | not null |
 admin_id           | integer                  |           | not null |
Indexes:
    "api_contributor_pkey" PRIMARY KEY, btree (id)
    "api_contributor_admin_id_key" UNIQUE CONSTRAINT, btree (admin_id)
Foreign-key constraints:
    "api_contributor_admin_id_3bfe88da_fk_api_user_id" FOREIGN KEY (admin_id) REFERENCES api_user(id) DEFERRABLE INITIALLY DEFERRED
Referenced by:
    TABLE "api_facilitylist" CONSTRAINT "api_facilitylist_contributor_id_c09fdaf8_fk_api_contributor_id" FOREIGN KEY (contributor_id) REFERENCES api_contributor(id) DEFERRABLE INITIALLY DEFERRED
```

## Notes

Civil Society Organizations are still called such, and were not switch to Civil Society Contributors.

## Testing Instructions

* Checkout this branch and `manage resetdb`
* Make an account and upload a file. Ensure it works correctly.
* Browse for location on the map. Ensure it works correctly.


src/django/api/models.py
```diff
@@ -42,16 +42,16 @@ def create_superuser(self, email, password, **extra_fields):
...
-class Organization(models.Model):
+class Contributor(models.Model):
     """
     A participant in or observer of the supply chain that will
     upload facility lists to the registry.
     """
     # These choices must be kept in sync with the identical list kept in the
     # React client's constants file
-    OTHER_ORG_TYPE = 'Other'
+    OTHER_CONTRIB_TYPE = 'Other'
```
kellyi: Originally I had thought we might have to make a corresponding change in the React client's constants file, but it turns out the only thing sent back here is `('Other', 'Other')`. I wonder if it might make sense just to call this `OTHER_TYPE` or something even more vague.

rajadain: I think this does some minor typing for us, so we don't mix-up `OTHER_CONTRIB_TYPE` with `OTHER_SOMETHINGELSE_TYPE`, even though they may both be `'Other'` as strings.

rajadain:
Thanks for taking a look! Going to merge this shortly.

Issue #104, opened 2019-01-18
Enable viewing and editing User profiles and related Organization details


kellyi:
## Overview

#87 temporarily replaced the existing Profile component with a placeholder UserProfile component. However, we should re-enable profile functionality.

The legacy component was the `Profile` component, which we should delete as part of this work. The app currently uses the new `UserProfile` component instead.

### Describe the solution you'd like

If the profile being viewed is not the currently-logged-in-user, then we should show the "View Profile" screen as depicted here: https://github.com/open-apparel-registry/open-apparel-registry/pull/87#discussion_r248833234

If the profile being viewed is the currently-logged-in-user, then the form should be editable and updates should be sent to the API. The auth reducer set up in #87 has a model for how we've set up other forms.

I've alreadty set up a `ProfileReducer` and `profile` actions, so we can use those as the basis for implementing this.

Design for this should be:

- create a new serializer and endpoint for sending non-logged-in user profile data by ID
- on visiting the user profile route declared here --  https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/App.jsx#L79 -- the application should make a request to the new non-logged-in user API endpoint. (Use the react router parameter for the user id as the `pk` to send to the API)
- when that data returns, the application should check whether the non-logged-in-user data is for the same user as the currently-logged-in user and set `isEditable` here: https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/reducers/ProfileReducer.js#L39
- if it's not the same user, render the profile form with all of the inputs disabled and don't show the API token section
- if it is the same user, render the profile form with the inputs enabled
- adjust the actions and Redux updates for the ProfileReducer to save the inputs; the actions have been set up here https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/components/UserProfile.jsx#L179 but they're currently essentially no-ops: https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/reducers/ProfileReducer.js#L89
- enable a `saveUserProfile` type action, create a corresponding API endpoint implementation which sends back the updated profile data on successfully saving
- place the updated data sent from the API back into the Redux store so that the profile form will update correctly

kellyi:
Marking this as blocked for now while I work on API tokens, which will make some adjustments to the profile.

PR #246, opened 2019-03-04
Enable viewing and editing profiles

kellyi:
## Overview

This PR enables viewing and editing profile data.

For logged in users, their profiles are now editable as depicted below. Edits required a password for confirmation and are limited to the following fields:

- name
- description
- contributor type
- other contributor type
- website

... all of which live on the contributor model

Profiles for other users are view-only.

I also updated the Facility Details page to have links to the contributor profiles.

Connects #104

## Demo

view-edit-profile
facility-profile-link
edit-profile-errors

## Notes

0f6e4cd  is an unrelated commit to fix a map loading bug, but I have it here so I could test it out in the course of working on this.

## Testing

- get this branch, then resetdb and register for two accounts
- sign in to one of the accounts and visit your profile page, verifying that it shows up as editable
- search for some facilities and then visit a details page. verify that there is now a link from any contributors to their profile pages and that those are view-only
- edit your profile and attempt to save it. verify that
   - not including a required field leads to an error message
   - an incorrect password shows an error message
   - successfully saving shows the toast and updates your profile data and, if it's different, the name in the app header
- run `./scripts/test` and verify that they pass


kellyi:
Rebased on develop.

rajadain:
Can we hide the "Website (Optional)" field title in read-only mode if the website is not specified? It looks a little odd.

image

kellyi:
> Can we hide the "Website (Optional)" field title in read-only mode if the website is not specified? It looks a little odd.

Sure!

kellyi:
> Can we hide the "Website (Optional)" field title in read-only mode if the website is not specified? It looks a little odd.

Handled in 282112ba604ddd83e463415b8767f3bc5aec18d2

kellyi:
2ce21d8352510244c6327f2465c0ec19bc6cb31a fixes a bug in which the facility details data was sending back the contributor IDs rather than the contributor user IDs, which meant that the profile links created from those IDs were incorrect.

Issue #221, opened 2019-02-21
Use trained gazetteer model to match uploaded records.


jwalgran:
Depends on the model built in #220 

* Add `dedupe` to container provisioning.
* Replace the placeholder match logic with the trained gazetteer model. Dedupe has a [StaticGazetteer](https://docs.dedupe.io/en/latest/API-documentation.html#StaticGazetteer) class for loading pretrained models.
* Determine if we should be matching records as a group (as all the dedupe examples do) or one at a time. Alter the `batch_process` command and AWS Batch job submission if needed.

Depends on #220

PR #248, opened 2019-03-05
Add dedupe-based facility list item matching 

jwalgran:
## Overview

Replaces the placeholder facility list matching process with a real, working process based on a trained [dedupe](https://github.com/dedupeio/dedupe) model.

Connects #221

## Notes

The provisioning time has increased a bit because dedupe depends on numpy which needs the `build-essential` apt package to be installed.

The initial version of this matcher used a pre-trained model downloaded from s3. Experiments with the pre-trained model revealed that the IDs of the canonical items used while training are returned as match results. This means that the model must be trained using actual data from the facilities table in order to have the proper IDs.

The `training.json` file that is read in during model training was created by running through an active learning session using the [oar_gazetteer_example.py](https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/research/dedupe/gazetteer/oar_gazeteer_example.py) research script against the 42 example public facility CSVs used to populate the previous version of the application.

We have updated the Vagrantfile and docker-compose.yml file to ensure that our development environment is provisioned with 2 cores. This avoid a bug uncovered during our gazetteer research that only appears when there is a single core available. See dedupeio/dedupe#734

The new match process is designed to process the full list at once, and does so efficiently. We have refactored the `batch_process` code to separate "item" actions from "list" actions. "match" is the only a "list" action at the moment, and possibly ever, so we did not generalize the calling of the function that does the actual work. .

We moved the handling of the `parse` and `geocode` actions to a `process_items` method to keep the length of the `handle` method under control.

Since the matching process only works on an entire list at once we have also switched to using a normal job in AWS Batch for matching rather than an array job

We added the Unidecode PyPI package following the lead of the dedupe examples.

    Unidecode provides [a function that] takes Unicode data and tries to
    represent it in ASCII characters (i.e., the universally displayable
    characters between 0x00 and 0x7F), where the compromises taken when mapping
    between two character sets are chosen to be near what a human with a US
    keyboard would choose.

## Testing Instructions

* ./scripts/manage resetdb
* Register an account
* Fully process the CSV
* Browse the list http://localhost:6543/lists/15?page=1&rowsPerPage=100 and verify that the match process completed successfully and produced a mix of automatic matches, new facilities, and a pending match.


jwalgran:
I'm moving the review assignment over to Kelly because Terrence ended up with three pending reviews.

jwalgran:
When a list is uploaded without any matched and exception is raised that crashes the matching job.

```
File "./manage.py", line 15, in <module>
execute_from_command_line(sys.argv)
File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 371, in execute_from_command_line
utility.execute()
File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 365, in execute
self.fetch_command(subcommand).run_from_argv(self.argv)
File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 288, in run_from_argv
self.execute(*args, **cmd_options)
File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 335, in execute
output = self.handle(*args, **options)
File "/usr/local/src/api/management/commands/batch_process.py", line 70, in handle
result = match_facility_list_items(facility_list)
File "/usr/local/src/api/processing.py", line 341, in match_facility_list_items
threshold = gazetteer.threshold(messy, recall_weight=recall_weight)
File "/usr/local/lib/python3.7/site-packages/dedupe/api.py", line 975, in threshold
return self.thresholdBlocks(blocked_pairs, recall_weight)
File "/usr/local/lib/python3.7/site-packages/dedupe/api.py", line 74, in thresholdBlocks
self.num_cores)['score']
File "/usr/local/lib/python3.7/site-packages/dedupe/core.py", line 214, in scoreDuplicates
raise BlockingError("No records have been blocked together. "
dedupe.core.BlockingError: No records have been blocked together. Is the data you are trying to match like the data you trained on?
```

We should catch this `BlockingError` and make new facilities out of all of the items.

kellyi:
Setting this up!

kellyi:
Tested this out and it worked as described!

I noticed that on list 8 the last item generated an error, which must have happened at the geocoding step? Is this the correct outcome, based on the address?

screen shot 2019-03-06 at 10 57 28 am

kellyi:
Question! Is it possible/desirable to write tests specifically for this functionality? That might help surface edge cases, but I don't know whether the dedupe model is "testable" in the usual way.

jwalgran:
I would like to add some tests to exercise this. Part of the reason for splitting up the matching and model creating/updating was to make testing easier (the other part was to not have to wrap the transaction around the model training unnecessarily)

jwalgran:
Thanks for the review.

Issue #253, opened 2019-03-05
Rejected match shows a status of "CONFIRMED_MATCH"


jwalgran:
## Overview

When rejecting a single match match, the status confusingly changes to "CONFIRMED_MATCH"

### Expected Behavior

After rejecting a single match, the status changes to "NEW_FACILITY".

### Actual Behavior

After rejecting a single match, the status changes to "CONFIRMED_MATCH".

### Steps to Reproduce

NOTE: These specific steps depend on #248

* ./scripts/manage resetdb
* Register an account
* Upload  ⁨src⁩/django⁩/api⁩/management⁩/commands⁩/facility_lists⁩/13.csv
* Fully process the CSV
  * `./scripts/manage batch_process --list-id 15 --action parse`
  * `./scripts/manage batch_process --list-id 15 --action geocode`
  * `./scripts/manage batch_process --list-id 15 --action match`
* Browse the list http://localhost:6543/lists/15?page=1&rowsPerPage=100
* Reject the "Workfield Knitwears" match and confirm the rejection

### Demo

![2019-03-05 11 22 53](https://user-images.githubusercontent.com/17363/53828095-ed64c200-3f39-11e9-8463-c7c570598578.gif)

PR #254, opened 2019-03-05
Display NEW_FACILITY when all matches are rejected

kellyi:
## Overview

This PR fixes a bug whereby list item rows for new facilities which were created by rejecting all potential matches were displayed in the `CONFIRMED_MATCH` style rather than in the `NEW_FACILITY` style.

Connects #253

## Demo

Potential matches:

screen shot 2019-03-05 at 2 06 16 pm

After rejecting all potential matches:

screen shot 2019-03-05 at 2 06 33 pm

## Testing Instructions

- get this branch, register for an account, and upload then process a list
- for a list item with a set of proposed potential matches, reject each potential match
- verify that once all the matches are rejected, the row becomes a `NEW_FACILITY` row

### DIFF
--- a/src/app/src/components/FacilityListItemsTable.jsx
+++ b/src/app/src/components/FacilityListItemsTable.jsx
@@ -105,7 +105,8 @@ function FacilityListItemsTable({
                 );
             }
 
-            if (item.status === facilityListItemStatusChoicesEnum.MATCHED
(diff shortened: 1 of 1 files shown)


kellyi:
> It can either be addressed as an additional commit to this PR or a new PR.

Oh bummer. I think we should do that as a new PR because I think it'd be worthwhile to write a test for it.

kellyi:
Thanks for the quick review!

Issue #255, opened 2019-03-05
A FacilityMatch is not created when all pending matches are rejected


jwalgran:
## Overview

When all pending matches for a FacilityListItem are rejected, a new Facility is created, however a FacilityMatch row is not being created, resulting in the contributor not appearing on the facility detail page.

### Expected Behavior

The facility detail page for a facility created after rejecting all pending matches displays the name of the contributor.

The tests are updated to verify that this works.

### Actual Behavior

The facility detail page for a facility created after rejecting all pending matches does not show the name of the contributor.

### Steps to Reproduce

NOTE: These specific steps depend on #248 and  #253

* ./scripts/manage resetdb
* Register an account
* Upload  ⁨src⁩/django⁩/api⁩/management⁩/commands⁩/facility_lists⁩/13.csv
* Fully process the CSV
  * `./scripts/manage batch_process --list-id 15 --action parse`
  * `./scripts/manage batch_process --list-id 15 --action geocode`
  * `./scripts/manage batch_process --list-id 15 --action match`
* Browse the list http://localhost:6543/lists/15?page=1&rowsPerPage=100
* Reject the "Workfield Knitwears" match and confirm the rejection
* Click the "Workfield Knitwears" link to browse the facility detail page

### Demo

<img width="1036" alt="screen shot 2019-03-05 at 12 35 30 pm" src="https://user-images.githubusercontent.com/17363/53831982-4d139b00-3f43-11e9-9dbd-d1b4d26ce443.png">

kellyi:
I updated the issue to include a line about having a test to verify that the code works.

PR #256, opened 2019-03-05
Create new FacilityMatch when all matches rejected

kellyi:
## Overview

- create a new FacilityMatch record along with a new Facility record
when all other potential matches have been rejected through the UI
- add test to verify the FacilityMatch count increments on rejecting
all matches

Connects #255 

## Demo

screen shot 2019-03-05 at 5 13 02 pm

## Testing Instructions

- get this branch, then run the tests to verify that they pass
- process a list, then reject all matches for one item in the UI. verify that it changes to `NEW_FACILITY` status and that it now has one more `FacilityMatch` than it did before: one created for the new facility


kellyi:
Great, thanks!

Issue #257, opened 2019-03-05
App crashes when "Style is not done loading"


kellyi:
## Overview

The app will crash if the OARMap component attempts to update the map before its styles are finished loading. We'll need to conditionally have the app delay trying to load styles until the map is loaded, but also make sure that once the styles are loaded the data loads and is turned into markers without any additional interaction.

![screen shot 2019-03-05 at 3 45 26 pm](https://user-images.githubusercontent.com/4165523/53840861-4730b080-3f69-11e9-9402-1f98f1a3ae7d.png)

### Expected Behavior

App does not crash

### Actual Behavior

App crashes when it tries to manipulate the map before styles have loaded

Issue #252, opened 2019-03-05
The "View my facilities" link on /lists sets the contributor ID in the contributor select list not the contributor name


jwalgran:
## Overview

"View my facilities" link on /lists is setting an id in the "Filter by contributor" box on the map page sidebar rather than the name of the contributor

### Expected Behavior

Clicking "View my facilities" navigates to the map page and sets the name of the contributor in the "Filter by contributor" box.

### Actual Behavior

Clicking "View my facilities" navigates to the map page but sets the ID of the contributor in the "Filter by contributor" box. Refreshing the page results in the contributor name being correctly selected.

### Steps to Reproduce

* Log in
* Upload and process a list
* Browse `/lists`
* Click "View my facilities"

### Demo

![2019-03-05 10 57 51](https://user-images.githubusercontent.com/17363/53826492-f18ee080-3f35-11e9-96cc-9396d01ef573.gif)

kellyi:
If I remember correctly it should flip over to "name" once the contributor name/id tuples have returned from the API? The bug here might be that we're displaying the search tab too early, before those results have returned. In other cases we show a loading spinner until the options have come back from the API.

rajadain:
Also, there seems to be crossover between users.

If I log in as a user TT1, and upload a list, and view my facilities I'm taken to `/facilities/?contributors=100`. Then, if I log out, and register / log in as new user TT2, and upload a new list, and view my facilities, I'm taken _again_ to `/facilities/?contributors=100`, instead of `/facilities/?contributors=101`. If I refresh the page somewhere in this sequence, it loads the correct facilities.

rajadain:
It seems like when `UPDATE_ALL_FILTERS` fires _before_ `COMPLETE_FETCH_CONTRIBUTOR_OPTIONS`, it behaves **correctly**, i.e. we see the contributor name, not ID, in the filter field. This is the case in a full page reload.

In a front-end routing case, `UPDATE_ALL_FILTERS` fires _after_.

kellyi:
> Also, there seems to be crossover between users.

Interesting. I wonder if we are using `memo` somewhere accidentally that's not getting cleared?

PR #258, opened 2019-03-05
Delay loading map data until styles have loaded

kellyi:
## Overview

- fixes a bug whereby the map would crash the app if the OARMap
component attempted to load data before the map's style was loaded

Connects #257

## Testing Instructions

- rebase work for #252 on this branch
- try this out for a bit and check that the app doesn't crash as it did before

### DIFF
--- a/src/app/src/components/OARMap.jsx
+++ b/src/app/src/components/OARMap.jsx
@@ -44,7 +44,7 @@ const OARMapStyles = Object.freeze({
     }),
 });
 
-const loadEvent = 'load';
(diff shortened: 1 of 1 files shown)


kellyi:
This one's probably low priority make work perfectly right since we're going to change the map library.

rajadain:
For the record, the way it would crash before was:

- Log in
- Upload a list
- Go to my lists
- Go to my facilities
- App would crash

kellyi:
Thanks for testing!

Issue #278, opened 2019-03-07
Facilities List is Incorrectly Cached


rajadain:
## Overview

If, after logging out from an account A, I login as a second account B, go to my Lists and click My Facilities, I am shown the facilities contributed by A, not by B.

### Expected Behavior

When My Facilities is clicked, the facilities contributed by the currently active account are shown.

### Actual Behavior

Facilities cached from last time are shown.

### Steps to Reproduce

* Have two accounts, A and B.
* Log in as A. Go to My Lists, then My Facilities. Note the URL and the contributor filter value.
* Log out. Log in as B. Go to My Lists, then My Facilities. Note the URL and the contributor filter correspond to A, instead of B.

### Demo

![2019-03-07 17 46 36](https://user-images.githubusercontent.com/1430060/53994889-7ed26080-4101-11e9-8968-407c1d95f3d3.gif)

Note the URL and the value in the filter.

### Additional context

The link from My Facilities is correct, but the change happens somewhere during routing. Thus, if the link for A is `/facilities/?contributor=100` and the list for B is `/facilities/?contributor=101`. The My Facilities link for B will point to `101`, and then navigate to `101`, but then somehow everything will shift to `100` and the URL will be updated to match.

kellyi:
I wonder if it makes sense to add a listener for the `COMPLETE_SUBMIT_LOGOUT` action in every reducer -- besides maybe the one for filter options -- which would reset that reducer back to its initial state. That should cure this and other similar bugs related to stale data.

PR #279, opened 2019-03-07
Correct Contributor Filter Value

rajadain:
## Overview

Previously, the labels shown in the filter selection box would match the contributor name only if UPDATE_ALL_FILTERS fired before COMPLETE_FETCH_CONTRIBUTOR_OPTIONS. The former set the list of active filters from the query string, and the latter updated that list with labels after fetching the data from the API.

In cases when COMPLETE_FETCH_CONTRIBUTOR_OPTIONS fired before UPDATE_ALL_FILTERS, such as loading the home page, then going to "My Lists", then selecting "My Facilities", the filter selection box would show the contributor's ID, not the label.

By matching the query string filters to the list of saved contributors in the state before firing UPDATE_ALL_FILTERS, we ensure that correct label is used in the case when it fires after COMPLETE_FETCH_CONTRIBUTOR_OPTIONS.

This ensures that the filter selection box has the correct value when the page is loaded for the first time (e.g. full page refresh with the query string params), or if it is navigated to from within the app (e.g. clicking "My Facilities").

Connects #252 

## Notes

I discovered #278 while working on this, but couldn't fix it easily, so spun it off into its own card.

## Testing Instructions

* Go to [:6543/](http://localhost:6543/) and log in to an account
* Go to My Lists, then My Facilities
* Inspect the contributor filter box. Ensure it shows your account name in its label, and not its ID.
* Do a full page refresh. Ensure it still shows the correct thing.

### DIFF
--- a/src/app/src/util/util.js
+++ b/src/app/src/util/util.js
@@ -407,3 +407,23 @@ export const sortFacilitiesAlphabeticallyByName = data => data
 
         return (a < b) ? -1 : 1;
     });
+
+// Given a list where each item is like { label: 'ABCD', value: 123 }, and
+// a payload which is a list of items like { label: '123', value: 123 },
+// returns a list of items from the payload with their labels replaced with
+// matching items found in the list.
+export const updateListWithLabels = (list, payload) => list
+    .reduce((accumulator, { value }) => {
+        const validOption = payload
+            .find(({ value: otherValue }) => value === otherValue);
+
+        if (!validOption) {
+            return accumulator;
+        }
+
+        return accumulator
+            .concat(Object.freeze({
+                value,
+                label: validOption.label,
+            }));
+    }, []);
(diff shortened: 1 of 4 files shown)


src/app/src/util/util.js
```diff
@@ -407,3 +407,23 @@ export const sortFacilitiesAlphabeticallyByName = data => data
 
         return (a < b) ? -1 : 1;
     });
+
+// Given a list where each item is like { label: 'ABCD', value: 123 }, and
+// a payload which is a list of items like { label: '123', value: 123 },
+// returns a list of items from the payload with their labels replaced with
+// matching items found in the list.
+export const updateListWithLabels = (list, payload) => list
```
kellyi: Cool! How would you feel about adding a test for this? I've been trying to keep this utils file at near 100% coverage in the utils.test.js suite.

rajadain: That's a great idea. I was a little nervous about the tenuousness of this code, adding tests will help with that.

rajadain: Add a test in a fixup and squashed it for simpler review.

rajadain:
Ready for another look.

rajadain:
Thanks for taking a look!

Issue #276, opened 2019-03-07
Implement OAR ID generator


jwalgran:
## Overview

Remove the placeholder generator and replace it with a final one.

### Describe the solution you'd like

- Incorporate ideas from email discussion.
- Incorporating the country code is a popular idea.
- Evaluate https://permid.org/

PR #282, opened 2019-03-08
Add OAR ID generation

jwalgran:
## Overview

This commit implements a new process for generating OAR IDs

Connects #276

### Notes

The new IDs have the following structure:

Position   Value
1-2        Country code
3-9        When OAR ID was generated as 4-digit year and 3-digit day of the year
10-14       A random value encoded from a restricted character set
15         A check digit

An example, with optional hyphens separating the tokens:

CN-2019067-NZ95A-M

This schema has the following desirable qualities:

- We don't know much about a facility when it is added to the Facility table. Country code and the date the ID was created are two useful, immutable pieces of information that we can use to build a meaningful ID.
- The random portion of the ID inherits these desirable qualities from the base32 implementation used to create them (https://www.crockford.com/base32.html):
  - Human readable and machine readable.
  - Compact. Humans have difficulty in manipulating long strings of arbitrary symbols.
  - Error resistant. Entering the symbols must not require keyboarding gymnastics.
  - Pronounceable. Humans should be able to accurately transmit the symbols to other humans using a telephone.
- The check digit allows the value to be algorithmically verified for errors.
- It is shorter than the GUID values used in the previous version of the
  application.

A 5-digit base32 number can encode around 32 million values. This combined with the date and country means that we are unlikely to exhaust the pool of available IDs.

## Testing Instructions

* `./scripts/update` to install the new Python dependency
* Verify that the unit test pass
* Run `manage.py resetdb` and verify that the fixtures load correctly.
* Register and log in
* Fully process the CSV
* Verify that new facilities are created with new IDs

### DIFF
--- a/src/django/api/oar_id.py
+++ b/src/django/api/oar_id.py
@@ -1,0 +1,78 @@
+import base32_crockford
+import random
+
+from datetime import datetime
+from api.countries import COUNTRY_NAMES
+
+
+def checksum(short_id):
+    """
+    Given an OAR ID string without the final digit, calculate and return the
+    final checksum digit.
+    """
+    country_val = ord(short_id[0]) + ord(short_id[1])
+    rest_val = sum([base32_crockford.symbols.index(x) for x in short_id[2:]])
+    return base32_crockford.symbols[(country_val + rest_val) % 32]
+
+
+def random_base_32_string(length):
+    """
+    Return a random number encoded as a zero-padded base 32 string of the
+    specified length.
+    """
+    maximum = pow(32, length) - 1
+    return base32_crockford.encode(random.randint(0, maximum)).zfill(length)
+
+
+def make_oar_id(country_code, hyphenate=False):
+    """
+    Return a new, semi-random OAR ID string with the following schema
+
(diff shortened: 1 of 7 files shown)


src/django/api/oar_id.py
```diff
@@ -0,0 +1,78 @@
...
+    r = random_base_32_string(5)
+    short_id = country_code + day + r
+    if hyphenate:
+        separator = '-'
+    else:
+        separator = ''
+    return separator.join([country_code, day, r, checksum(short_id)])
+
+
+def validate_oar_id(raw_id, raise_on_invalid=True):
```
kellyi: Wonder if we could add a similar function on the client that would just immediately fail requests for invalid OAR IDs without hitting the server again?

jwalgran: We could, but I'm not sure how valuable it would be. Failing to fetch a single record by primary key is always going to be pretty fast.

kellyi:
Discovered https://github.com/open-apparel-registry/open-apparel-registry/issues/292 while testing this.

jwalgran:
Thanks for the review.

PR #283, opened 2019-03-11
Reset filter options state on logout

kellyi:
## Overview

Resets the filter options state on logging out to fix a bug whereby the
facilities list link was being incorrectly set.

Connects #278

## Testing Instructions

- try to recreate the bug described in #278 and verify that it's fixed

### DIFF
--- a/src/app/src/reducers/FiltersReducer.js
+++ b/src/app/src/reducers/FiltersReducer.js
@@ -16,6 +16,8 @@ import {
     completeFetchCountryOptions,
 } from '../actions/filterOptions';
 
+import { completeSubmitLogOut } from '../actions/auth';
+
 import {
     updateListWithLabels,
 } from '../util/util';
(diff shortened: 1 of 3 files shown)


rajadain:
Taking a look now.

src/app/src/reducers/FiltersReducer.js
```diff
@@ -16,6 +16,8 @@ import {
     completeFetchCountryOptions,
 } from '../actions/filterOptions';
 
+import { completeSubmitLogOut } from '../actions/auth';
```
rajadain: Ah, great work. I see some other reducers also import and use this, but some others don't:

```bash
$ ag -L completeSubmitLogOut src/app/src/reducers/

src/app/src/reducers/index.js
src/app/src/reducers/UIReducer.js
src/app/src/reducers/UploadReducer.js
src/app/src/reducers/FilterOptionsReducer.js
src/app/src/reducers/FacilitiesReducer.js
src/app/src/reducers/OARMapReducer.js
```

Is it alright for them not to have it? Should this be something "higher up" that wraps the other reducers so we don't have to keep doing it manually?

kellyi: Yeah, those other reducers don't store any user-related data so I think they're good. However, I think the `FacilitiesReducer` should be dumped because it needs to stay harmonized with the set of selected filters in the FiltersReducer.

kellyi:
Thanks for reviewing! I added similar clear functionality to the Facilities reducer and also to the UI Reducer just to ensure that data is cleared on log out, too.

Issue #275, opened 2019-03-07
Allow facilities that have no geocoder results to proceed to matching


jwalgran:
## Overview

Currently, Google fails to geocode a facility, we do not allow it to be matched. It may be possible to match a malformed address to an existing facility, so we should not require a successful geocode to proceed.

### Describe the solution you'd like

  - Add new `ERROR_PARSING`, `ERROR_GEOCODING` and `ERROR_MATCHING` states.
  - Update the processing code to use these more detailed error states
  - Add a new `GEOCODED_NO_RESULTS` state
  - If the geocoder returns no results, use this new state rather than `ERROR`. If there is a more severe geocoding error, use `ERROR_GEOCODING`
  - Modify the `batch_process` command to process records that are either `GEOCODED` or `GEOCODED_NO_RESULTS`
  - If a match is not found for a facility that fails geocoding _do not_ attempt to create a new facility. We can't, because a lat/lng is required. Instead, set the state to `ERROR_MATCHING` and record the reason for the failure in the processing results.

PR #291, opened 2019-03-12
Allow list items to be matched if there are no geocode results

jwalgran:
## Overview

The dedupe matcher may be able to match a list item to an existing, geocoded facility even if the geocoder returns no results. To allow for this we have added a new `GEOCODED_NO_RESULTS` status and modified the workflow.

Connects #275

### Demo

Screen Shot 2019-03-12 at 8 58 46 AM

Screen Shot 2019-03-12 at 10 02 18 AM

### Notes

We add a `result_count` to the response from the geocoder so we can easily distinguish a successful geocode from one that did not fail, but also did not produce any results.

Items that cannot be geocoded and also fail to be matched end up in the `ERROR_MATCHING` status because we cannot create a new Facility without a lat/lng.

While we were changing the list of available status codes we took the opportunity to add more specific error states. Creating a separate error state for each stage of processing makes the status column displayed on the facility list item page less ambiguous and prepares us for easily resubmitting items for processing at the correct point in the workflow.

## Testing Instructions

 * Verify that the tests pass
 * Apply a diff to introduce some intermittent failures
```diff
diff --git a/src/django/api/processing.py b/src/django/api/processing.py
index c791d35..23450b0 100644
--- a/src/django/api/processing.py
+++ b/src/django/api/processing.py
@@ -42,6 +42,8 @@ def parse_facility_list_item(item):
     if item.status != FacilityListItem.UPLOADED:
         raise ValueError('Items to be parsed must be in the UPLOADED status')
     try:
+        if item.id % 5 == 0:
+            raise Exception('boom')
         fields = [f.lower() for f in parse_csv_line(item.facility_list.header)]
         values = parse_csv_line(item.raw_data)
         if CsvHeaderField.COUNTRY in fields:
@@ -77,6 +79,8 @@ def geocode_facility_list_item(item):
     if item.status != FacilityListItem.PARSED:
         raise ValueError('Items to be geocoded must be in the PARSED status')
     try:
+        if item.id % 4 == 0:
+            raise Exception('boom')
         data = geocode_address(item.address, item.country_code)
         if data['result_count'] > 0:
             item.status = FacilityListItem.GEOCODED
```
* Parse and geocode the CSV
* Verify that some items are in `ERROR_PARSING`, `ERROR_GEOCODING` and  `GEOCODED_NO_RESULTS` statuses.
* Match the CSV
* Verify that there is an item in the `ERROR_MATCHING` status and that it has an informative message


kellyi:
General question about this: can a facility with no geocoded location that makes it to matching in the UI end up becoming a facility on its own? The current serializer for facilities sends back geojson (and the client expects there to be coordinates for facilities), so I don't know how that would break down for facilities without geometries.

kellyi:
Nevermind, you answered it in the commit:

> Items that cannot be geocoded and also fail to be matched end up in the
`ERROR_MATCHING` status because we cannot create a new Facility without a
lat/lng.

kellyi:
I ran this according to the test instructions above and everything seems to be working as described except that I don't see any items in `ERROR_GEOCODING` status. I applied the patch and then uploaded lists 13, 2, and 10 and it's either `ERROR_PARSING` or `ERROR_MATCHING`. Not sure whether this is expected based on the state of the data here?

kellyi:
(Otherwise this seems to be working!)

jwalgran:
That is curious. Raising the exception in `geocode_facility_list_item` on every fifth record should definitely be creating some `ERROR_GEOCODING` rows. I'll repeat the test steps on my machine.

jwalgran:
Of course. My exception throwing diff incorrectly used `id % 5` for both tasks. Switching the geocode one to `id % 4` exercises the error handling properly

```diff
diff --git a/src/django/api/processing.py b/src/django/api/processing.py
index c791d35..8cba493 100644
--- a/src/django/api/processing.py
+++ b/src/django/api/processing.py
@@ -42,6 +42,8 @@ def parse_facility_list_item(item):
     if item.status != FacilityListItem.UPLOADED:
         raise ValueError('Items to be parsed must be in the UPLOADED status')
     try:
+        if item.id % 5 == 0:
+            raise Exception('boom')
         fields = [f.lower() for f in parse_csv_line(item.facility_list.header)]
         values = parse_csv_line(item.raw_data)
         if CsvHeaderField.COUNTRY in fields:
@@ -77,6 +79,8 @@ def geocode_facility_list_item(item):
     if item.status != FacilityListItem.PARSED:
         raise ValueError('Items to be geocoded must be in the PARSED status')
     try:
+        if item.id % 4 == 0:
+            raise Exception('boom')
         data = geocode_address(item.address, item.country_code)
         if data['result_count'] > 0:
             item.status = FacilityListItem.GEOCODED
```

Screen Shot 2019-03-13 at 9 51 54 AM

jwalgran:
Ready for another look now that I fixed the testing diff.

kellyi:
Setting this up again!

kellyi:
Screen Shot 2019-03-13 at 2 07 17 PM

jwalgran:
Thanks for the review.

Issue #272, opened 2019-03-07
The password complexity requirements are different for registration and password reset


jwalgran:
## Overview

User registration does not enforce any password requirements, but password reset requires a minimum length and that the password is not "common."

### Expected Behavior

User registration displays and enforces the same password requirements as password reset. Both pages display the password requirements.

### Actual Behavior

Any non-empty password value is accepted during user registration. Neither the user registration form nor the password reset form lists the password requirements.

### Steps to Reproduce

* Browse http://localhost:6543/auth/register
* Register an account with "p" as the password
* Browse http://localhost:6543/auth/login and click the password reset link
* Submit the password reset and browse the link printed to the server console
* Attempt to reset the password to "a" or "password".

PR #294, opened 2019-03-12
Enforce password requirements at registration

kellyi:
## Overview

- add a validate method to the user serializer to ensure it has the same
password length and complexity requirements as are enforced for the
reset password form

Connects #272 

## Demo

Screen Shot 2019-03-12 at 4 33 26 PM

Screen Shot 2019-03-12 at 4 33 56 PM

## Testing Instructions

- ./scripts/server then try to register an account with the following passwords:
   - `h`
   - `hello`
   - `password`
... and verify that they each fail
- choose a different password, then register and verify that registration succeeds
- on the sign in page, send the reset password email then get the link from the terminal and open it in the browser
- try to set a new password with the following passwords and verify that they all fail and have similar error messages:
   - `h`
   - `hello`
   - `password`
- choose a different password and verify that you are able to reset it and then sign in

### DIFF
--- a/src/django/api/serializers.py
+++ b/src/django/api/serializers.py
@@ -1,6 +1,8 @@
 from django.conf import settings
+from django.core import exceptions
 from django.db import transaction
 from django.contrib.auth.forms import PasswordResetForm
+from django.contrib.auth import password_validation
 from rest_framework.serializers import (CharField,
                                         EmailField,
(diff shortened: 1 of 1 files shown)


rajadain:
Taking a look now

kellyi:
Thanks!

Issue #183, opened 2019-02-06
Pressing the enter key in a form field should submit the form


jwalgran:
## Overview

The default behavior for a web form is to submit the form if you press <kbd>Enter</kbd>. We should provide this behavior.

### Describe the solution you'd like

- Pressing <kbd>Enter</kbd> with the focus on any field should submit the form. 
  - Login
  - Registration
  - Search
    - This one may not be possible if our select library does not support the proper event handling. If I focus the country field and type <kbd>c</kbd> <kbd>h</kbd> <kbd>i</kbd> <kbd>n</kbd>  <kbd>Enter</kbd>, "China" should be selected, as it is now. If I immediately press  <kbd>Enter</kbd> again we would prefer that the form is submitted.

kellyi:
I opted not to do this:

> This one may not be possible if our select library does not support the proper event handling. If I focus the country field and type c h i n Enter, "China" should be selected, as it is now. If I immediately press Enter again we would prefer that the form is submitted.

here.

I did update the facility name search to submit on pressing enter because it is a traditional text input:

![Screen Shot 2019-03-14 at 11 38 07 AM](https://user-images.githubusercontent.com/4165523/54370256-b2574280-464d-11e9-84e3-fbcda1ddd274.png)

However, the other filters on that tab are select boxes with text search and the built-in react-select behavior is to have it choose an option from the list on pressing enter.

PR #296, opened 2019-03-12
Submit forms on enter press

kellyi:
## Overview

Create utility function for enter key press handlers & add tests for
utility function.

Add handlers to the following forms to call a given submit form function
on pressing enter in a text input:

- register
- login
- request reset password
- confirm reset password
- profile
- search tab text search filter

This does not add the same handler to the search tab's multiselect
dropdowns, as their idiom is that an enter press selects one of the
options from the input. In other words: they're more like select
dropdowns searchable via text input than regular text inputs.

Connects #183

## Testing

Get this branch, then try out the following forms to verify that pressing enter in any of their text input fields submits the form:

- register
- login
- request reset password
- confirm reset password
- profile
- search tab text search filter

Also run the tests and verify that they pass.

### DIFF
--- a/src/app/src/__tests__/utils.tests.js
+++ b/src/app/src/__tests__/utils.tests.js
@@ -52,6 +52,7 @@ const {
     caseInsensitiveIncludes,
     sortFacilitiesAlphabeticallyByName,
     updateListWithLabels,
+    makeSubmitFormOnEnterKeyPressFunction,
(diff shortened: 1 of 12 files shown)


rajadain:
Taking a look now.

kellyi:
Great, thanks! I'm going to hold off on merging this one until tomorrow afternoon.

PR #293, opened 2019-03-12
Restyle lists table

kellyi:
## Overview

- hide the "Public" column on the lists table
- highlight the active list table row

Connects #274 

## Demo

Screen Shot 2019-03-12 at 4 02 33 PM

## Testing Instructions

- get this branch, then register and upload a list
- upload another list which replaces the first list
- visit the lists page and verify that you see what's depicted above

Issue #274, opened 2019-03-07
Hide the public column on /lists / make the values of the active column Yes or No / Use style to emphasize active lists


jwalgran:
## Overview

Clean up the display of the /lists page by

- Hiding the "Public" column as we are not yet implementing private lists
- Changing the values of the "Active" column to the more person-friendly, "Yes" or "No."
- Using styling to make active lists more prominent than inactive lists (Would appreciate a suggestion rom @designmatty)

designmatty:
You might consider doing something along these lines:
![image](https://user-images.githubusercontent.com/1928955/54045347-5c8f2000-419f-11e9-9c78-af3e37a1bdad.png)

dboyer:
In conversation with the client, we decided to go  with Active/Inactive in the Active column rather than Yes/No in order to be very clear.

kellyi:
> we decided to go with Active/Inactive in the Active column rather than Yes/No in order to be very clear

This makes sense. I will create another pull request for this issue to do that work rather than having to create another issue.

PR #299, opened 2019-03-13
Replace yes/no with active/inactive on lists table

kellyi:
## Overview

Replace yes/no with active/inactive on lists table

Connects #274

## Demo

Screen Shot 2019-03-13 at 10 55 55 AM

## Testing Instructions

- use the testing instructions from #293 to test this branch, but this time verify that the lists table's "Active" column now displays "Active" and "Inactive" instead of "Yes" and "No" to indicate which lists are active.


jwalgran:
This looks good. Ready to merge.

Issue #297, opened 2019-03-12
Site not loading in IE 11


jwalgran:
## Overview

A script error is preventing the application from loading in IE11.

### Expected Behavior

Browsing the app working in IE 11

### Actual Behavior

An unhandled script error causes the app to fail

```
SCRIPT5009: 'Promise' is undefined
1.chunk.js (59631,34)
```

### Steps to Reproduce

* Browse the root url with IE11

kellyi:
Promise polyfill was quick but now it's encountering this: https://github.com/zloirock/core-js/issues/193

Going to try adding the function name polyfill too.

PR #304, opened 2019-03-13
Use all of `core-js` to polyfill for IE11

kellyi:
## Overview

To prevent having to add & debug polyfills piecemeal, just import all of 'core-js'

Connects #297 

## Demo

Screen Shot 2019-03-13 at 4 38 27 PM

## Notes

If it's necessary to try to reduce the bundle size later, we can try to prune this polyfill down to only the imports we need. This also potentially makes available some ES7+ syntax we don't habitually use, which I think we can just avoid introducing through a combination of restraint and code review:

https://github.com/zloirock/core-js/tree/v2#ecmascript-7-proposals

## Testing Instructions

- get this branch, then serve the app and visit in in IE11, FF, Chrome, and Safari
- verify that it works in each

### DIFF
--- a/src/app/src/index.js
+++ b/src/app/src/index.js
@@ -1,10 +1,5 @@
 // polyfills for IE11
-// catalog of other is available at https://github.com/zloirock/core-js/tree/v2
-import 'core-js/fn/string/starts-with';
-import 'core-js/fn/object/values';
-import 'core-js/fn/array/find';
-import 'core-js/fn/array/find-index';
-import 'core-js/fn/object/assign';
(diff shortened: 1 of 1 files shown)


kellyi:
I guess the minifier's doing its job correctly because it doesn't appear that the polyfill caused the bundle to get much larger:

Screen Shot 2019-03-13 at 4 43 18 PM

kellyi:
The build is failing because of a problem with how one of the geocoding tests is working.

kellyi:
Great! Thanks for reviewing.

Issue #302, opened 2019-03-13
Change map markers and clusters to match the branding


kellyi:
## Overview

#284 implements using Google Maps and markerclusterer using the default markers from Google Maps for single facilities and a set of quickly-made icon images for the clusters.

We should update these markers to match the app's branding.

### Describe the solution you'd like

Update the facility markers and the cluster markers to match the branding.

### Additional Context

The images for markerclusterer are m1-m5.png in this directory: https://github.com/open-apparel-registry/open-apparel-registry/tree/develop/src/app/public/images

Updating those just requires dropping new png files in the directory with the same names. 1-5 are used differently by the library for clusters with different numbers of elements. The library will print the number of items in the cluster on top of the marker:

![Screen Shot 2019-03-13 at 1 38 47 PM](https://user-images.githubusercontent.com/4165523/54301474-65179a00-4595-11e9-8fd8-aa21b2311871.png)

![Screen Shot 2019-03-13 at 1 38 51 PM](https://user-images.githubusercontent.com/4165523/54301475-65b03080-4595-11e9-989e-b71871826138.png)

For the other markers, Google Maps affords using custom symbols: https://developers.google.com/maps/documentation/javascript/examples/marker-symbol-custom

This is where the app creates the markers: https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/components/FacilitiesMap.jsx#L130

jwalgran:
We will want to expedite this so that it is in place before the screen recording session at 9:30am Friday March 15

kellyi:
Just realized there is one other place where we need to update the marker style:

https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/components/FacilityDetailsStaticMap.jsx#L13

The static maps api allows you to request a marker to be placed in the image from the image url, which is what:

```
&markers=||${lat},${lng}
```

...is doing. The initial `||` is an empty placeholder which leads it to use the default marker but one is also able to pass in optional marker styles there, per: https://developers.google.com/maps/documentation/maps-static/dev-guide#Markers

Issue #244, opened 2019-03-03
Zoom map to facility location on selecting its details page from the facilities tab text list


kellyi:
## Overview

#237 restores a text listing of facilities search results in a new "Facilities" tab in the sidebar. Each list item links to its details page.

In the legacy implementation, clicking on a facility to see its details would also cause the map to zoom in to its marker. We should restore that functionality.

We should also have the map zoom to the facility location in cases when the user has visited the facility route directly, arriving at the application via the `/facilities/oar_id` route. In this case, the facility location data loads after the component has loaded, so the zoom will need to happen when the facility data returns successfully from the API.

### Describe the solution you'd like

On clicking the facility link in the sidebar's facilities tab, have the map zoom to the facility marker on the map.

On visiting the `/facilities/oar_id` route directly, the map should zoom to the facility marker when the facility data has returned from the API.

When clicking on a marker on the map to see facility data, the map should *not* zoom at all.

Requires #209 
Requires #264

kellyi:
See also #128

Issue #128, opened 2019-01-22
Clicking on a facility from the list of results on the left hand side of the page does not prompt the map to zoom to that facility


katieashaw:
## Overview

When viewing a list of facilities that has been presented from a search, they appear on the left hand side of the page. When you click on one of the facilities, it does not simultaneously zoom to that facility on the map on the right. 

### Expected Behavior

Upon selecting a facility from the list, I would expect the map to zoom to that facility. 

### Actual Behavior

When you click on a facility from the list on the left hand side of the page, the map remains in the same position. 

### Steps to Reproduce

* Run a search on the site
* Click on one of the facilities listed on the left hand side of the page

### Demo

N/A

### Additional context

N/A

PR #308, opened 2019-03-14
Adjust map style and behaviors

kellyi:
## Overview

- add custom markers and styles for clusters
- add and use custom marker for static image API
- use custom markers for facilities markers
- zoom map to marker on selecting it from the list or visiting it
directly

Connects #302 
Connects #244 
Connects #128 

## Demo

markers
markers-2

## Notes

As discussed in Slack we opted to keep the static endpoint and send a marker URL location, which will be the images dir from staging and production. In development, however, this will not work. To test this locally:

- run a static server from `src/app/public` in this repository
- ngrok http <whatever port>
- replace `icon:${window.location.origin}` with `icon:https://whatever.ngrok.io` in the FacilityDetailsStaticMap component, leaving `/images/static-marker.png.....` in place.

## Testing Instructions

- serve this branch
- set up ngrok as described in the notes field above
- visit the map and interact with it. verify that you see the new markers for clusters, the JS API, and the static API, and that interacting with the facilities list through the filter sidebar zooms the map to a location, but clicking on a marker on the map main does not zoom


kellyi:
> I would like to make 2 follow up issues

Sounds good. These are #312 and #313. I labeled them as design but we could also fix them too.

Issue #309, opened 2019-03-14
Split facility and facility details serializers


kellyi:
## Overview

The existing FacilitySerializer included some expensive queries for data which was only needed on the details page: other names, other address, other contributors.

Since these are only needed on the details page, we can speed up the main facilities endpoint by separating the serializer into a main serializer and a details serializer, the latter of which has more data.

### Describe the solution you'd like

Split the facilities list serializer into one for the main page and one for the details page.

PR #310, opened 2019-03-14
Split facility and facilitydetails serializers

kellyi:
## Overview

To improve performance on the full facilities query, remove some
unrelated and expensive queries for other_name, other_contributors, and
other_addresses which are only needed for the details page.

Create a new FacilityDetailsSerializer and update the React app's
propTypes in tandem.

Update fetchSingleFacility Redux action to get data from the API
unconditionally rather than trying to get it from the existing
facilities data -- since the latter no longer containers some required
details data.

Remove unused utility function and test.

Connects #309

## Testing Instructions

- serve this branch then search from the main map and verify that it still works
- visit a facility details page and verify that that still works too

### DIFF
--- a/src/app/src/__tests__/utils.tests.js
+++ b/src/app/src/__tests__/utils.tests.js
@@ -19,7 +19,6 @@ const {
     createQueryStringFromSearchFilters,
     allFiltersAreEmpty,
     createFiltersFromQueryString,
-    getFeaturesFromFeatureCollection,
     getValueFromEvent,
     getCheckedFromEvent,
     getFileFromInputRef,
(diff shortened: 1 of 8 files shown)


jwalgran:
Will this have any effect on the `/list/{id}` page?

kellyi:
> Will this have any effect on the /list/{id} page?

Great question. Let me check.

kellyi:
> Will this have any effect on the /list/{id} page?

I think the answer is no because the serializers there are not using the geojson serializer, but we may have reused the prop types in some way.

jwalgran:
Looking at this now.

kellyi:
https://github.com/open-apparel-registry/open-apparel-registry/issues/314 is for the map height adjustments -- I gave it a design label.

Issue #301, opened 2019-03-13
Update fixture generator and committed fixtures to be more realistic


jwalgran:
## Overview

We will soon be taking screen captures for promotional videos and asking for design feedback from some trusted contributors. Attempt to avoid questions and confusion by making the fixture data more closely resemble actual data.

### Describe the solution you'd like

- Geocode the facility address so that they show up at real locations rather than randomized points around Philladelpha.
- Generate contributor names that match the type of the contributor, rather than random names i.e. Brand A, Brand B, Manufacturer Group A, Service Provider B 
   - Create a roughly equal distribution of contributors in each type.

katieashaw:
Adding an additional request to this task for the screen record for the launch animation, scheduled for this Friday 15th at 9.30am EDT.

I've shared the storyboard for the animation with @jwalgran and @dboyer We will want to capture screen records for the following slides in the storyboard:

_Possibly_ slide 15: Easily upload your data (happy to visualise this differently, if you prefer not to have a screen record of this yet)
Slide 16: Search by facility
Slide 17: Filter by contributor
Slide 18: Identify facility affiliations

For the purposes of the screen record, I only need to be able to run one search using dummy, but logical, data - i.e. addresses plotting in the correct country, showing multiple versions of the address that have been matched, and a few affiliations for the facility.

The video production company has said that animating fake data over the top of the screen record will be difficult for them, so is there a way you can create a search for me to carry out that matches this criteria, and let me know what to search by on Friday?

jwalgran:
I don't think there is any problem showing the data upload page (/contribute)

Our demo data already has associations built in so we will be able to provide you with a search that you can run and a facility that you can select that will show the relationships mentioned in the slite 18.

katieashaw:
Wonderful, thank you!

kellyi:
I think we should drop the `fake.url()` from the website part of the fixture generator -- https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/django/api/management/commands/makefixtures.py#L147 -- and instead just have it be a constant value, like `"https://info.openapparel.org"`.

jwalgran:
After deploying #315 I ran commands to reload the database and run the demo data through the matching algorithm. Things are set up in a way that should allow us to record the demos mentioned in the storyboard.

**Slide 16: Search by facility
Slide 18: Identify facility affiliations**

[...]

PR #315, opened 2019-03-14
Create more realistic fixtures and use real matching

jwalgran:
## Overview

Replaces random data in fixtures with more reasonable looking data and removes the use of facility and facilitymatch fixtures in favor of using the real match process.

- Adds CSVs with geocoded results
- Processor loads geocoded points from the CSV rather than making geocode requests
- Makes contributor names based on type
- Makes all fixtures lists active
- Make `cN@example.com` the username so that we can easily log in as any of the fixture users.
- Makes more realistic randomized file names rather then using person names.
- Adds a command to run all the processing steps on all the fixture data.

Connects #301 

## Demo

Screen Shot 2019-03-14 at 3 35 33 PM

Screen Shot 2019-03-14 at 3 39 41 PM

## Testing Instructions

* Verify `./scripts/manage resetdb` runs without error
* Verify `./scripts/manage processfixtures` runs without error 
* Log on as `c8@example.com`:`password`. Browse http://localhost:6543/lists/8 and verify that there are potential matches.


jwalgran:
Fixing up the lint now

kellyi:
Looks great!

Screen Shot 2019-03-14 at 6 57 35 PM

PR #320, opened 2019-03-15
Restore initial map bbox on reset button click

kellyi:
## Overview

Restore the initial map bbox position when the search tab's reset button
is clicked.

Disable the reset button altogether when there are no active filters and
there is no facilities data in the app state.

## Testing Instructions

- serve this branch, then search for some facilities
- verify that resetting the filter state repositions the map to its initial bbox

### DIFF
--- a/src/app/src/actions/ui.js
+++ b/src/app/src/actions/ui.js
@@ -8,3 +8,6 @@ export const updateSidebarFacilitiesTabTextFilter =
     createAction('UPDATE_SIDEBAR_FACILITIES_TAB_TEXT_FILTER');
 export const resetSidebarFacilitiesTabTextFilter =
     createAction('RESET_SIDEBAR_FACILITIES_TAB_TEXT_FILTER');
+
+export const recordSearchTabResetButtonClick =
+    createAction('RECORD_SEARCH_TAB_RESET_BUTTON_CLICK');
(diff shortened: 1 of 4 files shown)


kellyi:
I noticed one quirk with this which is that the map itself doesn't distinguish cases when the reset button has been clicked and cases when the filters have been emptied manually by the user removing all prior options. In both cases it will reset the initial zoom.

If that's acceptable behavior then this shouldn't need any additional changes; if it's not I can make adjustments to get rid of that behavior, too.

jwalgran:
>I noticed one quirk with this which is that the map itself doesn't distinguish cases when the reset button has been clicked and cases when the filters have been emptied manually

Good catch. I think that is counter intuitive. If you are looking at a particular region of the world and text search for "shoes" then you clear out the text box to search for "coats" you loose your zoom and pan.

kellyi:
> I think that is counter intuitive. 

Makes sense -- I'll figure out how to circumvent it and update this PR.

kellyi:
@jwalgran I rebased and pushed an updated commit which now only resets the map on reset button clicks -- and not when manually updating the filters. I also reverted the change to disable the reset button when the filters were empty, as it now makes sense to have it active when the map has changed but no filters are there.

Issue #281, opened 2019-03-08
Translation only works when logged in and cannot be disabled on the map page


jwalgran:
## Overview

The app supports translation via a Google Translate script and supporting code loaded on the index page.

https://github.com/open-apparel-registry/open-apparel-registry/blob/0d1003ce1e1168838f865b53d4fa8a28abe47807/src/app/public/index.html#L71-L76

There are two broken behaviors:
- Switching language has no effect unless you are logged in.
- If you log in, browse the map page and choose an alternate language, you cannot revert back to English. The `googtrans` cookie must be deleted to undo the translation.

### Expected Behavior 1

If I browse the map page and I am not logged in, when I choose a non-English language the page is translated into that language.

### Actual Behavior 1

Choosing a non-english language has no effect.

### Steps to Reproduce 1

* Browse http://localhost:6543 when not logged in.
* Choose a non-English language from the "LANGUAGE" menu.

---

### Expected Behavior 2 

When logged in and browsing the map page when I select a non-English language, the page is translated. When I switch back to English, the page is restored.

### Actual Behavior 2

When logged in and browsing the map page when I select a non-English language, the page is translated. When I switch back to English, the page briefly appears in English then reverts back to the previously selected language.

### Steps to Reproduce 2

* Browse http://localhost:6543 
* Log in
* Choose a non-English language from the "LANGUAGE" menu.
* Choose English from the the "LANGUAGE" menu.

PR #322, opened 2019-03-15
Replace Translate component with native Google Translate UI

kellyi:
## Overview

Recent changes to the app broke the previous implementation of Google
Translate using the Translate.jsx component. Further investigation
revealed that the way the Translate component had been set up was not
really in harmony with how Google Translate is supposed to be set up --
which increased its surface area for bugs and made it incompatible with
existing documentation like

https://www.w3schools.com/howto/howto_google_translate.asp

To remedy this, we drop the Translate component and use the standard
method for setting up Google Translate: giving Google Translate a div on
which to create its own UI that's managed outside of React. This
circumvents having to manually set the hash, call window.reload, and
sync the cookies. It also puts the Google Translate usage back into
harmony with the documentation that is available online.

Note that it turns out that Google Translate for websites seems to be
deprecated, so although it is currently working here we should probably
be prepared for it to break at any point in the future.

Connects #281 
Connects #325 

## Demo

icon-anchors

## Notes

Just to reiterate: it appears that Google translate for Websites is not accepting new signups. The landing page here https://translate.google.com/intl/en/about/website/ suggests that it will continue to work for websites that have already signed up, but it also suggests directing users to make use of browsers with built-in translation extensions.

Screen Shot 2019-03-18 at 1 16 49 PM

One additional thing: as depicted in the demo gif above, there's an interval between when the app starts and when translation is available -- and in this moment, the navbar does not have a "Select Language" or selected language tab. This is a consequence of waiting for the function call to return, and although it's visually different than what was there before I believe this is the correct way for it to behave.

## Testing

- serve this branch then use the translation element and verify that the bugs described in #281 no longer occur


src/app/src/components/UserProfile.jsx
```diff
@@ -250,7 +255,7 @@ function mapStateToProps({
 }) {
     return {
         user,
-        fetching: fetching || sessionFetching,
+        fetching: fetching || sessionFetching || authFetching,
```
kellyi: The changes here are to fix a bug which was surfaced by updating how translation works: namely, there's a moment when `profile` doesn't exist and the form fields shouldn't try to render at all, but they were and it threw an error.

I can move this out into another commit if it would make more sense to have it separated.

kellyi:
> What do you think about incorporating these changes?

I wasn't really able to find any documentation about terms of service for using Google Translate, and I assume it's mostly been removed since the service stopped accepting new signups. However, I was wondering whether there's some terms of service thing that required displaying the icon too?

Not having documentation around means we'd probably just be speculating one way or another. If you think it's ok to go without the icon, I'm happy to add the patch above. Alternatively, we could keep

```
b.childNodes[0].innerHTML = 'Language';
```

from the patch but drop

```
+.goog-te-gadget-icon {
+    display: none;
+}
```

... which would just shorten the string back to "Language" but keep the icon.

Happy to do whatever we deem best, though! As mentioned above, it seems like the translation thing is no longer supported to it's hard to guess accurately about what its terms were.

jwalgran:
>I wasn't really able to find any documentation about terms of service for using Google Translate

I would like to hide the icon since there is no indication that we can't do so.

kellyi:
> I would like to hide the icon since there is no indication that we can't do so.

:+1:

kellyi:
Sounds good. This was accomplished in 
09ba486 

Screen Shot 2019-03-18 at 3 48 12 PM

kellyi:
Hm, that timeout doesn't seem to be universally working. Going to see if there's a onload handler.

kellyi:
I force pushed 
c9a4764 which adjusts the style on load by using a MutationObserver: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver that gets removed once it has been executed.

I also adjusted how the code manipulated the navbar nodes: using `.innerHTML` on the "Select Language" node broke a feature whereby the language was propagated into the Navbar, so I kept "Select Language" around but added some logic for removing the dropdown arrow icon.

Screen Shot 2019-03-18 at 4 20 49 PM

Screen Shot 2019-03-18 at 4 23 21 PM

If we just want to have it always say "Language" then I can revert back to using = innerHTML

src/app/public/index.html
```diff
@@ -71,9 +71,45 @@
...
+                layout: google.translate.TranslateElement.InlineLayout.SIMPLE,
+                autoDisplay: true,
+            }, translateElementID);
+
+            var callback = function () {
+                var el = document.getElementsByClassName('goog-te-menu-value');
+
+                if (el.length) {
+                    var navbarElement = el[0];
+                    var childNodesLength = navbarElement.childNodes.length;
```
kellyi: Assigned out of the loop because otherwise the .length property would decrement as nodes were removed.

kellyi:
I added 83ac59c  which is an unrelated commit that removes the `LandingAlert` component which we decided to drop in #325

kellyi:
c6f9d2f replaces the for loop which removed some elements from the tab with a while loop. This was done because the for loop would run past the end of the `childNodes` list since removing an element from it would change its size.

kellyi:
Great! Thanks for your help with this. I'm going to squash then merge once the tests pass.

Issue #305, opened 2019-03-13
Attempting to delete a user account through Django admin raises an exception


jwalgran:
## Overview

Attempting to delete a user account through Django admin raises an exception

### Expected Behavior

A newly registered account can be deleted through the Django admin

### Actual Behavior

An exception is raised

### Steps to Reproduce

* Register a new user account
* Log into http://localhost:8081/admin/ as a superuser
* Try and delete the newly added user

### Demo

<img width="1155" alt="Screen Shot 2019-03-13 at 3 55 20 PM" src="https://user-images.githubusercontent.com/17363/54319956-ae251980-45a8-11e9-87f4-45e7d9a43619.png">

### Additional context

It looks like we need to add an installed app in order to run migrations for allauth social

Issue #269, opened 2019-03-06
Add feature to support user deleting an account


dboyer:
## Overview

Add option on user profile for users to delete their account

### Is your feature request related to a problem? Please describe.

GDPR compliance requires that users have the option to delete personal information stored in the system

### Describe the solution you'd like

Two options:

1. We add a note to the FAQ that users can be deleted by emailing OAR and requesting the account be deleted. OAR admins then forward the email to the development team.

2. We add a button that lets users delete their account and we are not involved at all

jwalgran:
Our situation is complicated by the fact that we have separated the `User` from the `Contributor` in our data model.

- User
   - email
   - should_receive_newsletter
   - has_agreed_to_terms_of_service
   - is_staff
   - is_active
- Contributor
   - admin (User)
   - name (currently labeled as Account Name)
   - description (currently labeled as Account Description)
   - website
   - contrib_type
   - other_contrib_type

All uploaded data is related to the `Contributor` rather than the `User`. This would make it easier to delete a given `User` account, but there are some questions:

- Do any of the fields on the `Contributor` model fall under "personal" information?
- `Contributor.admin` is currently required. Do we make this nullable? If not, to what `User` do we relate the `Contributor` after the previous `admin` is deleted?

dboyer:
After some reading, my thought is that Name could fall under personal information since there is no requirement that someone enter a brand name rather than their personal name.

PR #329, opened 2019-03-18
Enable deleting users

kellyi:
## Overview

- enable users to be deleted through the Django admin by adding a
missing `allauth.socialaccount` app to installed apps
- update migration to make Contributor a nullable field on FacilityLists
and to set the value to null when the Contributor is deleted
- adjust Facility get_contributors method to skip if
match.facility_list.contributor is None
- add `delete_contributor` Django management command that take an email
address as its argument and deletes the User, Contributor, and
EmailAddress objects associated with the email address

Connects #305 
Connects #269

## Demo

Screen Shot 2019-03-18 at 5 36 31 PM

->

Screen Shot 2019-03-18 at 5 37 31 PM

->

Screen Shot 2019-03-18 at 5 37 12 PM

-> (later, when added from another list)

Screen Shot 2019-03-18 at 5 49 19 PM

## Testing Instructions

- get this branch, then `./scripts/resetdb`
- register for a new user through the UI, then upload list 11 and batch_process all steps to create some facities
- search for some facilities and verify that they appear and that your user/list is displayed among the contributors on its details page
- run `./scripts/manage delete_contributor --email <YOUR_EMAIL>` and verify that it succeeds
- visit the map page and search for facilities. Verify that that works and then visit the details page for a few facilities to observe that the now have no contributors
- register for a new user with a different email address, then upload list 11 again, process it, then look at facilities details pages and verify that the new user is listed among the contributors
- verify that you can delete a user through the Django admin


jwalgran:
We test users registering accounts on staging and over 1000 new facilities have appeared on the staging map. It would be a little disruptive to reset the database at this time, so I would like to consider creating actual migrations for these changes, and then possibly squashing them back down into single migration later.

src/django/api/management/commands/delete_contributor.py
```diff
@@ -0,0 +1,42 @@
...
+
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from allauth.account.models import EmailAddress
+
+from api.models import User, Contributor
+
+
+class Command(BaseCommand):
```
jwalgran: I would like to consider giving this the more detailed name `delete_user_and_contributor` as a way of reinforcing that we are deleting multiple things.

kellyi:
> so I would like to consider creating actual migrations for these changes, and then possibly squashing them back down into single migration later.

:+1: Sounds good to me. I'll update this to create a new migration rather than replacing the prior one.

src/django/api/management/commands/delete_contributor.py
```diff
@@ -0,0 +1,42 @@
...
+        group.add_argument('-e', '--email',
+                           required=True,
+                           help='The user\'s email address')
+
+    def handle(self, *args, **options):
+        email = options['email']
+
+        try:
+            user = User.objects.get(email=email)
+            contributor = Contributor.objects.get(admin=user)
```
jwalgran: We should consider moving the some exception handling up to this contributor fetch. It is possible to have a user account without a related contributor. If that is the case, we still want to delete the user account.

kellyi:
dab1c32 adjusts the PR to create a new migration for the null Contributor work.

src/django/api/management/commands/delete_contributor.py
kellyi: Sounds good. I renamed it in 
26f33df

src/django/api/management/commands/delete_contributor.py
kellyi: Good point. I updated this in 
26f33df by adding separate nested try blocks for the Contributor and EmailAddress objects.

kellyi:
Renamed the command in ea654bf261f137328c28b4e99f8ecb94b0a9cf0b
to be `delete_user_and_contributor` which more accurately describes what it's doing!

Issue #290, opened 2019-03-12
Clicking the Submit button on /contribute multiple times submits multiple copies of the file


jwalgran:
## Overview

Clicking the `Submit` button on `/contribute` multiple times submits multiple copies of the file.

### Expected Behavior

After clicking `Submit` the form state is cleared.

### Actual Behavior

The selected facility list remains after clicking `Submit` allowing the form to be reposted.

### Steps to Reproduce

* Log in
* Browse `/contribute`
* Select a file and click `Submit`
* Click `Submit` a second time

PR #330, opened 2019-03-18
Clear file input on successful upload

kellyi:
## Overview

Clear file input on successful upload from contributor page to prevent
a bug whereby the file would still be attached to the input & would
upload again on clicking submit another time.

Connects #290 

## Demo

Screen Shot 2019-03-18 at 6 03 24 PM

## Testing Instructions

- ./scripts/server
- register for an account, then visit the contributor page and upload a file
- when the file has succeeded, click the "Submit" button again and verify it does not resubmit the prior file but instead displays an error message


rajadain:
Taking a look now.

PR #334, opened 2019-03-19
Update translate tab CSS rules

kellyi:
## Overview

Remedy a styling bug on staging by updating the translate tab's CSS rules to
ensure the custom rules are applied.

Connects #322 

## Demo

Screen Shot 2019-03-19 at 12 28 31 PM

## Notes

Don't know exactly why this isn't working on staging, but I was guessing it may have something to do with the loading order of the CSS rules.

## Testing Instructions

- check this branch locally to verify that the translate nav tab appears correctly
- visit staging in the browser and manually apply the set of rules declared here to the CSS override in the inspector and verify that it corrects the style bug

### DIFF
--- a/src/app/src/styles/css/Map.css
+++ b/src/app/src/styles/css/Map.css
@@ -50,17 +50,16 @@
 }
 
 .goog-te-gadget-simple {
-    font-size: 12px;
+    font-size: 12px !important;
     color: #0427a4;
-    cursor: pointer;
     text-decoration: none;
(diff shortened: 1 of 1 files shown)


kellyi:
I was able to recreate this bug on the local cibuild version, so instead of hacking staging instead you can:

- ./scripts/cibuild
- visit `localhost:8081` and verify that the tab is styled as expected:

Screen Shot 2019-03-19 at 3 16 10 PM

Issue #224, opened 2019-02-22
Setup production stack


jwalgran:
TODO: add a checklist of items that must be completed in order to provision a production stack and deploy a release.

hectcastro:
I created https://github.com/azavea/operations/issues/298 to track this because we are having trouble connecting the OAR org to Codetree. It was definitely working before, but it is not now. Working with Codetree support to get it resolved.

hectcastro:
Instantiate all of the resources currently supporting the staging environment in AWS, but for production.

- Updated production variables in remote `.tfvars` file
- DNS
- Certificates
- Email
- Rollbar
- Mapbox
- Google Geocoding
- Database w/ migrations

dboyer:
Note that #263 requires production be in an EU-based data center.

rbreslow:
- [ ] Make sure `develop`, `test/`, `release/`, and `hotfix/` all go to staging
- [ ] Add `Jenkinsfile.release` which takes the SHA of an existing container image and runs the `infra` process against the production settings bucket
- [ ] Add `Open Apparel Registry Release Pipeline` scripted pipeline job to Jenkins
- [ ] Create a GitHub issue template for releases, similar to what was done in https://github.com/raster-foundry/raster-foundry-python-client/pull/75, which walks through git flow steps, as well as running the pipeline

PR #338, opened 2019-03-19
Add release pipeline

rbreslow:
## Overview

* Add `Jenkinsfile.release` which takes the SHA of an existing container image and runs the `infra` process against the production settings bucket
* Add [`Open Apparel Registry Release Pipeline`](http://civicci01.internal.azavea.com/view/oar/) scripted pipeline job to Jenkins
* Make sure `develop`, `test/`, `release/`, and `hotfix/` all go to staging
* Set `PYTHONUNBUFFERED=1` in AppCLI task def

Connects #224 

## Testing Instructions

See: http://civicci01.internal.azavea.com/view/oar/job/Open%20Apparel%20Registry%20Release%20Pipeline/17/

You can also run your own build based off of the latest `develop` SHA.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -59,7 +59,7 @@ node {
 					withCredentials([[$class: 'StringBinding',
 									credentialsId: 'OAR_ROLLBAR_ACCESS_TOKEN',
 									variable: 'OAR_ROLLBAR_ACCESS_TOKEN']]) {
-						sh 'docker-compose -f docker-compose.yml -f docker-compose.ci.yml run --rm terraform ./scripts/infra apply'
+						sh 'docker-compose -f docker-compose.ci.yml run --rm terraform ./scripts/infra apply'
 					}
 				}
 			}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,9 +12,6 @@ node {
 
 		env.OAR_SETTINGS_BUCKET = 'openapparelregistry-testing-config-eu-west-1'
 
-	    // Execute `setup` wrapped within a plugin that translates
-	    // ANSI color codes to something that renders inside the Jenkins
-	    // console.
 		stage('setup') {
 			wrap([$class: 'AnsiColorBuildWrapper']) {
 				sh './scripts/bootstrap'
@@ -29,7 +26,7 @@ node {
 
 		env.OAR_SETTINGS_BUCKET = 'openapparelregistry-staging-config-eu-west-1'
 
-		if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('test/')) {
+		if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/')) {
 			// Publish container images built and tested during `cibuild`
 			// to the private Amazon Container Registry tagged with the
 			// first seven characters of the revision SHA.
@@ -44,10 +41,10 @@ node {
 				}
 			}
 
-			// Plan and apply the current state of the infrastructure as
-			// outlined by whatever branch of the `open-apparel-registry`
+			// Plan and apply the current state of the staging infrastructure
+			// as outlined by whatever branch of the `open-apparel-registry`
 			// repository passes the conditional above (`develop`, 
-			// `release/*`, `test/*`).
+			// `test/*`, `release/*`, `hotfix/*`).
 			stage('infra') {
 				// Use `git` to get the primary repository's current commmit SHA and
 		        // set it as the value of the `GIT_COMMIT` environment variable.
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -3,12 +3,14 @@
 node {
     properties([
         parameters([
-            string(
+            [
+                $class: 'ValidatingStringParameterDefinition',
                 defaultValue: '',
-                description: 'SHA of an existing container image. Usually what\'s currently on staging.', 
+                description: 'First 7 characters of the SHA for the commit you wish to deploy.',
+                failedValidationMessage: 'Invalid SHA.',
                 name: 'GIT_COMMIT',
-                trim: false
-            )
+                regex: /^[a-z0-9]{7}$/
+            ]
         ])
     ])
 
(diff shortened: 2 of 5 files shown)


Jenkinsfile
```diff
@@ -12,9 +12,6 @@ node {
 
 		env.OAR_SETTINGS_BUCKET = 'openapparelregistry-testing-config-eu-west-1'
 
-	    // Execute `setup` wrapped within a plugin that translates
```
rbreslow: I didn't think this comment was super valuable.

Jenkinsfile.release
```diff
@@ -0,0 +1,79 @@
+#!groovy
+
+node {
+    properties([
+        parameters([
+            string(
+                defaultValue: '',
```
rbreslow: We could also go the route of using `ValidatingStringParameterDefinition`. 

I could see this preventing a scenario where someone leaves this field blank, presses build, doesn't cancel the build, and then the infrastructure is compromised. At the same time, you can still enter the wrong or a non-existent SHA, and do just as much damage.

Another possibility is we require a release tag to be threaded through here, instead of a SHA.

I also feel fine leaving things the way they are.

Jenkinsfile.release
```diff
@@ -0,0 +1,79 @@
...
+                sh './scripts/manage ecsmanage -e production showmigrations'
+            }
+        }
+    } catch (err) {
+        // Some exception was raised in the `try` block above. Assemble
+        // an appropirate error message for Slack.
+        def slackMessage = ":jenkins-angry: *Open Apparel Registry (Release Pipeline) (${params.GIT_COMMIT}) #${env.BUILD_NUMBER}*"
+
+        slackMessage += "\n<${env.BUILD_URL}|View Build>"
+        //slackSend channel: '#oar', color: 'danger', message: slackMessage
```
rbreslow: Reminder to uncomment this before merging.

Jenkinsfile.release
```diff
@@ -0,0 +1,79 @@
...
+                                variable: 'OAR_ROLLBAR_ACCESS_TOKEN']]) {
+                    sh 'docker-compose -f docker-compose.ci.yml run --rm terraform ./scripts/infra apply'
+                }
+            }
+        }
+
+        // Kick off an ECS task to display any outstanding migrations.
+        stage('showmigrations') {
+            wrap([$class: 'AnsiColorBuildWrapper']) {
+                sh './scripts/manage ecsmanage -e production showmigrations'
```
rbreslow: Frustratingly, some of the logs for this task came back empty again.

Jenkinsfile.release
hectcastro: I think it makes sense to add support `ValidatingStringParameterDefinition` such that a valid value needs to be a 7 character alphanumeric value. I tested this regular expression in isolation to achieve that: `^[a-z0-9]{7}$`.

Would it also be valuable to enable the `trim` flag?

Jenkinsfile.release
```diff
@@ -0,0 +1,79 @@
+#!groovy
+
+node {
+    properties([
+        parameters([
+            string(
+                defaultValue: '',
+                description: 'SHA of an existing container image. Usually what\'s currently on staging.',
```
hectcastro: I'd also consider revising this text so that it is specific about requiring the first 7 characters of the relevant Git SHA.

Jenkinsfile.release
hectcastro: It might be valuable to include another message earlier in the `Jenkinsfile` that signals a production release is underway.

Jenkinsfile.release
hectcastro: I wonder if this is happening because the management command output is being buffered. I think it would be valuable to try setting the `PYTHONUNBUFFERED` environment variable for the CLI task definition to minimize I/O buffering.

See: https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUNBUFFERED

Jenkinsfile.release
rbreslow: Sounds good.

`$` matches the end of line, so I don't think this would be necessary.

Jenkinsfile.release
rbreslow: This discussion supports using this on AWS:

https://github.com/awslabs/amazon-sagemaker-examples/issues/319

Jenkinsfile.release
rbreslow: Because Rollbar lets you report the status of a deploy (`started`, `succeeded`, `failed`, or `timed_out`), I wanted to see if we could lean on them entirely for Slack notifications.

I added a function called `reportDeploy(String status);`, and got rid of the existing Rollbar stuff in `infra`.

Then, you could do stuff like:

```groovy
try { 
    stage('infra') {
        ...
    }
} catch(err) {
    reportDeploy("failure");
}
```

The problem is that Rollbar only seems to report when a deploy succeeds. Reporting another status seems to only populate metadata, which can be seen here: https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/deploys/.

See: https://docs.rollbar.com/reference#post-deploy

Jenkinsfile.release
rbreslow: Logs still don't appear to show up:

https://eu-west-1.console.aws.amazon.com/ecs/home?region=eu-west-1#/clusters/ecsProductionCluster/tasks/feefe694-ec8f-449c-83ec-b66c04e65b5f/details

Maybe I need to inject that environment variable in a different way?

hectcastro: Hm. Now wondering if this is an issue with our Django logging configuration. I noticed in `settings.py` that we don't have `LOGGING` configured. We have `LOGLEVEL`, but I'm not seeing that setting used any where, nor referenced in the Django docs.

It might be worth configuring Django logging to see if that works. If not, opening an issue so that we can address it later. Something like the snippet below adds support for `stdout` logging with a configurable `DJANGO_LOG_LEVEL`:

```python
import os

LOGGING = {
    'version': 1,
    'disable_existing_loggers': False,
    'handlers': {
        'console': {
            'class': 'logging.StreamHandler',
        },
    },
    'loggers': {
        'django': {
            'handlers': ['console'],
            'level': os.getenv('DJANGO_LOG_LEVEL', 'INFO'),
        },
    },
}
```

rbreslow: Tried to configure that as well, https://eu-west-1.console.aws.amazon.com/ecs/home?region=eu-west-1#/clusters/ecsProductionCluster/tasks/30d11c21-a9e8-44b5-8006-0d663db7e9b9/details, still doesn't work. 

I opened #358.

Issue #335, opened 2019-03-19
Error when retrieving list due to a gateway timeout


nataliegrillon:
## Overview

I had uploaded a list and when I went back to my lists , I can no longer access it via clicking on "my lists" and then clicking on the list entry. I get the response, "unable to retrieve list - an error prevented fetching facility list items for 25" - please see screen recording

### Expected Behavior

I should have bene able to click on the list and have the entries load.

### Actual Behavior

"Unable to retrieve list - an error prevented fetching facility list items for 25"

### Steps to Reproduce

Mac, Chrome in incognito

- After a csv has been uploaded, leave it for several hours to let it parse
- refresh page
- click on my lists 
- Select list from table of my lists
- Error message appears

### Demo

![list_access_issue_3 19 2019_trimmed](https://user-images.githubusercontent.com/17363/54635116-bb2a8880-4a40-11e9-97f2-68d771eb8b11.gif)

### Additional Context

[Esprit_02_2019_converted.March.13.2019.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/2985222/Esprit_02_2019_converted.March.13.2019.xlsx)

jwalgran:
To help with troubleshooting can you please attach the list that you uploaded? If you rename the file and add a `.txt` extension you will be able to attach it to this issue. Thanks.

nataliegrillon:
Yes of course - sorry I didn't think of that. Added it as an excel file - hope that is OK

[Esprit_02_2019_converted March 13 2019.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/2985059/Esprit_02_2019_converted.March.13.2019.xlsx)

jwalgran:
Cool. I forgot that Github accepts xlsx. That works. Thanks.

jwalgran:
The request to https://staging.openapparel.org/api/facility-lists/25/ is returning 504 Gateway Timeout. It is taking too long load the list. The source CSV has 1385 rows. 

Options:
- Implement server-side paging
- Delay loading the details for rows until they are clicked (with the exception of pending matches, which should be expanded by default.

PR #341, opened 2019-03-20
Implement server side paging for facility list items

jwalgran:
## Overview

We were running into gateway timeout issues when attempting to load all items for a large list (thousands of rows)

To solve this we have added a separate endpoint for fetching list items and implemented paging based on the built-in Django REST Framework paging support.

We were able to reuse the bulk of the client side paging logic but we now fetch new data when the page or page size is changed rather than just getting a slice of the full data in memory.

Connects #335

## Notes

This breaks the client side CSV downloading. This is logged as a separate issue (#344) and will be fixed in a follow up PR

## Testing Instructions

* Log in as c8@example.com
* Use the React tools in the developer console to verify that only 20 rows have been loaded into the state
* Use the page navigation and verify that the data is correctly paged.

### DIFF
--- a/src/app/src/actions/facilityListDetails.js
+++ b/src/app/src/actions/facilityListDetails.js
@@ -1,10 +1,12 @@
 import { createAction } from 'redux-act';
+import querystring from 'querystring';
 
 import csrfRequest from '../util/csrfRequest';
 
 import {
     logErrorAndDispatchFailure,
     makeSingleFacilityListURL,
+    makeSingleFacilityListItemsURL,
     createConfirmOrRejectMatchData,
     createConfirmFacilityListItemMatchURL,
     createRejectFacilityListItemMatchURL,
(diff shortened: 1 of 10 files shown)


jwalgran:
Posted a draft for some early feedback. The pagination is working but I need to fix the CSV download feature so it requests all the data one page at a time and joins it together.

jwalgran:
Assigned to @kellyi because he built the original paging.

kellyi:
Setting this up to try it out!

kellyi:
I've been testing this out and it's working great so far. I've tried out a 1700 item list and am currently setting up a 2900 item list and the paging works for all the cases I've tried:

- changing the page size or page number behaves correctly
- refreshing with the querystring params load the correct page
- the table renders correctly

This is probably known but just in case it's not: it looks like there's a propTypes error and a stray console.log statement:

Screen Shot 2019-03-20 at 11 38 49 AM

We should fix the proptypes error by just dropping the `.isRequired` suffix.

If we'd like, we could potentially spin off the CSV downloading to another card in the interest of expediting getting this onto staging for more user testing.

jwalgran:
>If we'd like, we could potentially spin off the CSV downloading to another card in the interest of expediting getting this onto staging for more user testing.

I do think it is wise to merge this before fixing downloads since it is a bit of a showstopper

kellyi:
I found a bug when trying to confirm/reject a facility match:

Screen Shot 2019-03-20 at 11 58 11 AM

The request goes through correctly to the API but it appears to be crashing on updating the app state afterward.

jwalgran:
facepalm
Posted a bit too early. Thanks. Will fix up these bugs.

jwalgran:
>We should fix the proptypes error by just dropping the .isRequired suffix.

Addressed in fixup aa30fc0

>I found a bug when trying to confirm/reject a facility match:

Addressed in fixup ac49218

>it looks like there's [...] a stray console.log statement

Addressed in fixup 7716228

jwalgran:
Fixed up and ready for review

kellyi:
Cool, about to test this again.

jwalgran:
Thanks for the review and the well-designed paging UI that was easily repurposed.

I squashed the fixups and will merge when the update passes CI.

Issue #326, opened 2019-03-18
Vietnam country filter not appearing correctly [SHARES POINT WITH 324]


katieashaw:
## Overview

Vietnam appears in the dropdown country filter as Viet Nam. When you type Vietnam into the country search box, it's not recognised.

### Expected Behavior

Users should be able to search the site using Vietnam (all one word).

### Actual Behavior

Vietnam is listed in the country drop down menu as Viet Nam. 

### Steps to Reproduce

* Visit staging.openapparel.org
* Search Vietnam in the country dropdown menu

### Demo

![Screenshot 2019-03-18 at 17 52 05](https://user-images.githubusercontent.com/46344729/54551768-a26d9480-49a6-11e9-8f87-6ca1658e1a45.png)
![Screenshot 2019-03-18 at 17 52 14](https://user-images.githubusercontent.com/46344729/54551781-a6011b80-49a6-11e9-986f-efc71f656169.png)

### Additional context

Optional. Add any other context about the problem here.

jwalgran:
Good catch. Thank you. The wikipedia list of 2-character country codes listed used the 2-word form, which is how it got into our list.
<img width="951" alt="Screen Shot 2019-03-18 at 10 57 07 AM" src="https://user-images.githubusercontent.com/17363/54552174-b98f9080-496c-11e9-9d13-a6689c02cc56.png">

katieashaw:
Gosh, how intriguing!

Issue #324, opened 2019-03-18
Cookie notification does not need to pop up every time you visit the site [SHARES POINT WITH 326]


katieashaw:
## Overview

The legally required pop-up re. tracking cookies that has been added to the site appears every single time you visit the site, rather than just the first time you access it. My understanding, and experience through using websites in Europe, is that the pop-up just needs to appear once, be accepted, and then that acceptance applies for all future visits to the site. 

### Expected Behavior

The cookies tracking pop-up should just appear the first time a site is visited. Having accepted that cookies will be tracked, the pop up should not appear again on future visits. 

### Actual Behavior

The pop up appears and has to be accepted every time you visit the OAR site.

### Steps to Reproduce

* Load the home page
* Accept the cookie popup
* Load the home page in a new tab or browser window

jwalgran:
The accept/reject status is currently tracked in `sessionStorage` which only is discarded when the tab/window is closed. We should move the status persistence to `localStorage`.

PR #345, opened 2019-03-20
Remember Cookie Acceptance, Use "Vietnam"

rajadain:
## Overview

Switches to `localStorage` so we can remember a user's cookie acceptance preference between sessions.

Also switches to using "Vietnam" instead of "Viet Nam".

Connects #324 
Connects #326 

## Demo

Cookie Acceptance variable in `localStorage`:

image

"Vietnam" in country dropdown:

image

And in parsed results:

image

## Testing Instructions

* Check out this branch and `server` and `manage migrate`
* Go to [:6543/](http://localhost:6543/)
  - [x] Ensure you see a Cookie Acceptance box
* Accept cookies into your computer
* Once the box is removed, close that tab, open a new tab and go to [:6543/](http://localhost:6543/)
  - [x] Ensure you don't see the Cookie Acceptance box
* Search for "Viet" in the country filter.
  - [x] Ensure you see "Vietnam" as a suggestion
  - [x] Ensure you don't see "Viet Nam" as a suggestion
* Sign up / Log in.
* Upload [15.csv](https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/django/api/management/commands/facility_lists/15.csv)
* Run `manage batch_process --action parse --list-id 16` with the correct list id of the newly uploaded list
* Go to that list's detail view
  - [x] Ensure the parsed Vietnamese address says "Vietnam" as the country


src/django/api/countries.py
```diff
@@ -241,7 +241,7 @@
     ('UZ', 'Uzbekistan'),
     ('VU', 'Vanuatu'),
     ('VE', 'Venezuela, Bolivarian Republic of'),
-    ('VN', 'Viet Nam'),
+    ('VN', 'Vietnam'),
```
kellyi: Is this going to prevent parsing facility lists that have `Viet Nam` listed as the country name? That seems to be the ISO standard for writing the country name.

src/django/api/migrations/0001_initial.py
```diff
@@ -58,7 +58,7 @@ class Migration(migrations.Migration):
...
...), ('ZA', 'South Africa'), ('GS', 'South Georgia and the South Sandwich Islands'), ('SS', 'South Sudan'), ('ES', 'Spain'), ('LK', 'Sri Lanka'), ('SD', 'Sudan'), ('SR', 'Suriname'), ('SJ', 'Svalbard and Jan Mayen'), ('SZ', 'Swaziland'), ('SE', 'Sweden'), ('CH', 'Switzerland'), ('SY', 'Syrian Arab Republic'), ('TW', 'Taiwan, Province of China'), ('TJ', 'Tajikistan'), ('TZ', 'Tanzania, United Republic of'), ('TH', 'Thailand'), ('TL', 'Timor-Leste'), ('TG', 'Togo'), ('TK', 'Tokelau'), ('TO', 'Tonga'), ('TT', 'Trinidad and Tobago'), ('TN', 'Tunisia'), ('TR', 'Turkey'), ('TM', 'Turkmenistan'), ('TC', 'Turks and Caicos Islands'), ('TV', 'Tuvalu'), ('UG', 'Uganda'), ('UA', 'Ukraine'), ('AE', 'United Arab Emirates'), ('GB', 'United Kingdom'), ('US', 'United States'), ('UM', 'United States Minor Outlying Islands'), ('UY', 'Uruguay'), ('UZ', 'Uzbekistan'), ('VU', 'Vanuatu'), ('VE', 'Venezuela, Bolivarian Republic of'), ('VN', 'Vietnam'), ('VG', 'Virgin Islands, British'), ('VI', 'Virgin Islands, U.S.'), ('WF', 'Wallis and Futuna'), ('EH', 'Western Sahara'), ('YE', 'Yemen'), ('ZM', 'Zambia'), ('ZW', 'Zimbabwe')], help_text='The ISO 3166-1 alpha-2 country code of the facility.', max_length=2)),
```
kellyi: We should make this into a new migration because we no longer want to drop staging data when we change the models.

src/django/api/countries.py
rajadain: No, parsing should still work correctly (and for both variants) because they are covered further down in the file:

https://github.com/open-apparel-registry/open-apparel-registry/blob/e7e8177aafebde721fdcbd6b8ac9b1e99e488acf/src/django/api/countries.py#L561-L562

src/django/api/migrations/0001_initial.py
rajadain: Ah, alright I'll make a new one.

rajadain:
Separated into a new migration in 465b2a0.

src/django/api/migrations/0001_initial.py
rajadain: Separated into a new migration in 465b2a0.

rajadain:
Thanks for taking a look! Squashed the fixups, going to merge now.

Issue #342, opened 2019-03-20
Geocoded addresses can sometimes be too long for the `geocoded_address` field


kellyi:
## Overview

Saw this error when geocoding a 2900 item list:

```
database_1  | 2019-03-20 15:40:01.209 UTC [11302] ERROR:  value too long for type character varying(200)
[...]

Issue #292, opened 2019-03-12
Processing demo list 14 causes the geocoding process to crash


kellyi:
## Overview

Processing demo list 14 causes the geocoding process to crash. There may be bad data in the list but whatever's in it breaks the exception handling.

### Expected Behavior

geocoding list 14 succeeds

### Actual Behavior

geocoding list 14 fails with the following message:

```
Exception ignored in: <function CPointerBase.__del__ at 0x7fbe04b19620>
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/django/contrib/gis/ptr.py", line 36, in __del__
  File "/usr/local/lib/python3.7/site-packages/django/contrib/gis/geos/libgeos.py", line 155, in __call__
  File "/usr/local/lib/python3.7/site-packages/django/utils/functional.py", line 36, in __get__
  File "/usr/local/lib/python3.7/site-packages/django/contrib/gis/geos/libgeos.py", line 159, in func
ImportError: sys.meta_path is None, Python is likely shutting down
```

### Steps to Reproduce

- upload demo list 14
- parse demo list 14
- geocode demo list 14

jwalgran:
We ran into a problem before with that exception masking an actual problem with inconsistent fixture data:

https://github.com/open-apparel-registry/open-apparel-registry/pull/195#issuecomment-462849518

jwalgran:
Likely the same as 292

PR #350, opened 2019-03-20
Increase the max length of FacilityListItem.geocoded_address

jwalgran:
## Overview

We were running into crashes in the geocoding process because we were trying to insert more than 200 characters into the `geocoded_address` field. The value of this field comes from the `formatted_address` field in the Google geocoder response. We could not find any documentation mentioning how long this field might be so we took a guess and made the field 5x larger that it was.

Connects #342 
Connects #292

## Testing Instructions

I was not able to reproduce the issue with 14.csv as reported in #292. I was able to reproduce it with 
[handm-long-geocode-test.csv.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/2990307/handm-long-geocode-test.csv.txt)

* On `develop`, upload [handm-long-geocode-test.csv.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/2990307/handm-long-geocode-test.csv.txt) and run it through `parse` and `geocode` Verify that an exception is raised.
* Checkout this branch and `./scripts/manage migrate`
* Run the `geocode` process again and verify that it completes without error (some items may "fail" because they were previously geocoded successfully)


jwalgran:
I updated the test instructions and attached a file that can be used to reproduce the error.

rajadain:
Taking a look now.

rajadain:
Unfortunately I'm unable to reproduce the issue on `develop`. Tried both `14.csv` and `handm-long-geocode-test.csv`.

jwalgran:
Interesting. I assume that the responses from the Google are changing out from under us.

I used this diff to force the error on develop.


```diff
diff --git a/src/django/api/processing.py b/src/django/api/processing.py
index eec5036..e5d6745 100644
--- a/src/django/api/processing.py
+++ b/src/django/api/processing.py
@@ -92,7 +92,7 @@ def geocode_facility_list_item(item):
                     data["geocoded_point"]["lng"],
                     data["geocoded_point"]["lat"]
                 )
-                item.geocoded_address = data["geocoded_address"]
+                item.geocoded_address = ('A' * 150) + data["geocoded_address"]
             else:
                 item.status = FacilityListItem.GEOCODED_NO_RESULTS
             item.processing_results.append({
```

rajadain:
Great, I'll try that.

jwalgran:
Thanks for the review.

Issue #346, opened 2019-03-20
Limit the number of results returned from a search on the map page


jwalgran:
## Overview

Searches that return thousands of facilities are slow and take a long time to clear. Limit the number of facilities returned from a search from the map page

### Describe the solution you'd like

* Implement paging on the /facilities API with a maximum page size of 500. Use #341 as a reference for paging querysets. rest-framework-gis provides [GeoJsonPagination](https://github.com/djangonauts/django-rest-framework-gis#pagination)
* Do not attempt to implement paging on the map. The map will always just load the first page of results.
* Run a `.count()` query and return the total number of matching results. Show a "showing 500 of 3423 results" message

PR #351, opened 2019-03-20
Paginate facilities geojson

kellyi:
## Overview

Use django-rest-framework-gis's built-in geojson pagination to limit the
number of facilities returned from any given search to 500 for regular
requests.

Adjust the facilities tab header to display "n facilities of m results"
when m is larger than n.

Connects #346 

## Demo

All potential results have returned from the API:

Screen Shot 2019-03-20 at 6 09 42 PM

There are more results available on the next page in the API:

Screen Shot 2019-03-20 at 6 05 03 PM

Narrowing the facilities even further with a search string:

Screen Shot 2019-03-20 at 6 05 10 PM

## Notes

I set the default page size to 500 for now

## Testing Instructions

- serve this branch then resetdb and processfixtures
- from the main map page, click search and verify that you see 500 facilities returned and a count of 914
- run another search for "shoe" and verify that that returns 35 facilities, with no additional results

### DIFF
--- a/src/app/src/components/FilterSidebarFacilitiesTab.jsx
+++ b/src/app/src/components/FilterSidebarFacilitiesTab.jsx
@@ -178,6 +178,12 @@ function FilterSidebarFacilitiesTab({
     const orderedFacilities =
         sortFacilitiesAlphabeticallyByName(filteredFacilities);
 
+    const facilitiesCount = get(data, 'count', null);
+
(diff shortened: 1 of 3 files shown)


rajadain:
Taking a quick look.

rajadain:
Oh I see, if I Search for "shoe" I get 35 results:

image

but if I Filter for "shoe" with no explicit search keyword I get 32. Makes sense, because of the pagination.

kellyi:
Thanks!

Issue #347, opened 2019-03-20
Use a different marker for the currently selected facility


jwalgran:
## Overview

Design a marker that is thematically similar to the existing markers and show it on the main map and the mini map when a feature is selected.

### Is your feature request related to a problem? Please describe.

>I can't tell which facility I've selected on the map after I click on it. The color remans purple on the google maps marker, making if difficult to remember which marker I clicked on and to be able to see where the facility is on the map

### Describe the solution you'd like

* Create a new marker
* Use the new marker for the selected facility and the mini map.

jwalgran:
CC @designmatty the development team will wire it in but we would appreciate help in designing a "selected facility" marker.

jwalgran:
Thanks, Matt.

![markerSelected@2x](https://user-images.githubusercontent.com/17363/54713803-bc28ec00-4b0c-11e9-982f-631ba11507bd.png)

PR #353, opened 2019-03-21
Add selected facility marker

kellyi:
## Overview

This PR changes the marker for a facility when it is selected, primarily by detecting the selected facility from the react-router details path.

One additional change here is that routes like http://localhost:6543/facilities/CN2019084R04PD8?name=shoe no longer load both the facility for the details page and the list of facilities for the search automatically on first arriving at the app from a URL entered in the browser. The details page still loads, but I updated the code to prevent the other facilities from loading, too, in order to simplify the number of cases that had to be handled for the imperative code set up around adjusting the marker styling.

Third commit is unrelated but fixes a linter complaint which I think issues from how react-router parses memoized components in the version we're using.

Connects #347

## Demo

regular-searches
visit-facility-directly
multiple-faciltiies

## Testing

- serve this branch then search for some facilities
- verify that selecting a facility changes its marker to selected and also unselects other facilities
- verify that you can navigate directly to a details page and see the selected facility marker
- verify that multiple facilities at the same point show the selected maker when they are selected


kellyi:
TODO:

- [ ] make sure marker shows on route like localhost:6543/facilities/TR2019080MJHCA7?name=shoe
- [ ] figure out what to do with markers with multiple facilities at the same point

jwalgran:
Looking at this now

src/app/src/components/FacilitiesMap.jsx
```diff
@@ -251,11 +290,119 @@ class FacilitiesMap extends Component {
...
+        }
+
+        if (this.markersLayer && data) {
+            const markerForOARID = this.markersLayer
+                .find(marker => marker.get('oarID') === oarID);
+
+            if (markerForOARID) {
+                markerForOARID.setIcon({
+                    url: selectedMarkerURL,
+                    scaledSize: new this.mapMethods.Size(30, 40),
```
jwalgran: We may want to consider extracting this literal `Size` into a constant since we repeat it several times.

kellyi: Good point, will try this out.

kellyi:
> I think we should update static-marker.png to be the selected marker color. It will help conceptually connect the point on the main map with the point on the mini map.

Good suggestion. I updated this in bf4ea56c4ef364cb99ab865d5c9daff6df1f1269

Screen Shot 2019-03-25 at 6 05 03 PM

Testing it prior to staging will require:

- serving /src/app/public dir from a static server
- ngrok http <whatever-port-of-the-static-server>
- updating the `window.location.origin` text in the FacilityDetailsStaticMap component to be `"https://whatever.ngrok.io"`

src/app/src/components/FacilitiesMap.jsx
kellyi: I tried this out in ac98a485884393861ebe112a9fc49644544fe101 however I can't tell if it's better or not. The `Size(30, 40)` value gets set from a call to `this.mapMethods` initially, which means it gets set after the map mounts.

I used that to set a `this.scaledMarkerSize` class variable, but once I added guards to re-initialize it if necessary in places where it wasn't already initialized it ended up adding more code rather than reducing code.

jwalgran: 👍

Issue #323, opened 2019-03-18
Show the total number of facilities on the search tab


nataliegrillon:
## Overview

~Add an "ALL contributors" Option: It would be nice to be able to see all the factories at once rather than only a selection.  By selecting all contributors in the drop down of contributors you could see all of the data and all of the facilities in the database.~ Perhaps we could also have a total facilities counter down in the lower right hand corner of the search bar so that a user knew how many facilities were in the database

### Is your feature request related to a problem? Please describe.

I can't see all the data in the database or know how  big the data set is.

### Describe the solution you'd like

~In the contributor drop down, have an "all contributors" option and~ have a counter in the bottom right hand corner of the search to see the total number of facilities in the database at any one point. 

@dboyer @katieashaw

jwalgran:
It would be possible, and relatively easy, to show the total number of facilities on the "Search" tab.

There are over 10,000 unique facilities in the original set of 42 public supplier lists. With the current map design we are going to struggle to load and display that many points simultaneously. Like the previous version of the application, we are probably going to want to put a cap on the maximum number of results we will return from the API and display at one time.

nataliegrillon:
There are over 10k in the database currently! That's great.

Understood on the results for the API- that makes sense and is no problem
from my perspective. The goal here is more to indicate to users what the
total database has just for their reference.

Best,

Natalie
----
Natalie Grillon
Project Director, Open Apparel Registry
+ 1 413 884 3355
*natalie@openapparel.org <natalie@openapparel.org> *
*Open Apparel Registry <http://openapparel.org>*

jwalgran:
We only have about 900 in the demo data, but when I was working on training the matching algorithm using the public lists you provided the total number of rows was around 18 thousand and there were several thousand duplicates. Unless we load a fewer number of list or the revised versions of those lists have shrunk dramatically I would expect that we will have around 10k.

jwalgran:
I am changing the title of this issue to reflect the change in scope.

nataliegrillon:
ok awesome!

Best,

Natalie
----
Natalie Grillon
Project Director, Open Apparel Registry
+ 1 413 884 3355
*natalie@openapparel.org <natalie@openapparel.org> *
*Open Apparel Registry <http://openapparel.org>*

PR #357, opened 2019-03-22
Display total facilities on sidebar search tab

kellyi:
## Overview

- add an endpoint to get the total facilities count
- add a component + Redux actions for retrieving and displaying total facilities count

Connects #323

## Demo

Screen Shot 2019-03-22 at 10 52 41 AM
Screen Shot 2019-03-22 at 10 52 21 AM

## Notes

Making a whole new reducer for this was probably a bit of overkill, but I wanted to limit the scope of the fetching/error state for this data since it doesn't impact anything else in the application.

## Testing Instructions

- serve this branch then visit the search tab
- verify that you see the total facilities count displayed in the bottom right when the facilities_count request returns

### DIFF
--- a/src/django/api/views.py
+++ b/src/django/api/views.py
@@ -481,6 +481,18 @@ class FacilitiesViewSet(ReadOnlyModelViewSet):
         except Facility.DoesNotExist:
             raise NotFound()
 
+    @action(detail=False, methods=['get'])
+    def count(self, request):
+        """
+        Returns a count of total Facilities available in the Open Apparel
+        Registry.
+
+        ### Sample Response
+            { "count": 100000 }
+        """
+        count = Facility.objects.count()
+        return Response({"count": count})
+
 
 class FacilityListViewSet(viewsets.ModelViewSet):
     """
--- a/src/app/src/actions/facilityCount.js
+++ b/src/app/src/actions/facilityCount.js
@@ -1,0 +1,38 @@
+import { createAction } from 'redux-act';
+
+import csrfRequest from '../util/csrfRequest';
+
+import {
+    logErrorAndDispatchFailure,
+    makeGetFacilitiesCountURL,
+} from '../util/util';
+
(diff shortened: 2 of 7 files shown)


src/django/api/views.py
```diff
@@ -307,6 +307,20 @@ def all_countries(request):
     return Response(COUNTRY_CHOICES)
 
 
+@api_view(['GET'])
+@permission_classes((AllowAny,))
+def facilities_count(request):
```
jwalgran: Consider moving this under the `FacilitiesViewSet` to so we can `GET /api/facilities/count/` without adding an explicit URL.

```python
@action(detail=False, methods=['get'])
def count(self, request):
    """
    Returns a count of total Facilities available in the Open Apparel
    Registry.

    ### Sample Response
       { "count": 100000 }

    """
    count = Facility.objects.count()
    return Response({"count": count})
```

src/django/oar/urls.py
```diff
@@ -39,6 +39,7 @@
     url(r'^api/contributor-types/', views.all_contributor_types,
         name='all_contributor_types'),
     url(r'^api/countries/', views.all_countries, name='all_countries'),
+    url(r'^api/facilities_count', views.facilities_count, name='facilities_count'),
```
jwalgran: Consider moving this under the FacilitiesViewSet, but if we need to keep an explicit URL, we should consider `api/facilities-count`. Hyphens are generally preferred in URLs

src/django/api/views.py
kellyi: Cool, that is definitely better:

Screen Shot 2019-03-22 at 2 21 13 PM

Screen Shot 2019-03-22 at 2 21 13 PM

I will swash the fixup together with the two commits here.

src/django/oar/urls.py
kellyi: Ended up moving this under the FacilitiesViewSet.

Issue #240, opened 2019-03-01
Script account creation for each brand with lists added by OAR


dboyer:
## Overview

OAR admins will add lists for several brands. Rather than have all the lists associated with a single OAR user account, we need to create a user account for each brand (ex. OAR - Brand Name X, OAR - Brand Name Y).

### Is your feature request related to a problem? Please describe.

If we upload all the lists under a single admin account, it limits search options.

### Describe the solution you'd like

Script the creation of user accounts so that OAR admins can add lists more easily and the accounts can be transferred to the brands in the future.

- Create a management command that takes a few parameters
  - A comma separated list of contributor names
  - A contributor type matching one of the types defined in models.py
  - A password
- The command creates new user/contributor accounts 
  - The email for the account should be info+{cleaned-name}@openapparel.org where cleaned-name is the name of the contributor lowercased with whitespace and punctuation changed to `-`
  - The description is set to "Public facility lists for {name} managed by the OAR Team"
  - Contributor type and password set to the specified values

rajadain:
Is this the kind of interface we're imagining:

```bash
$ manage create_accounts --contributors "abc", "D ef", "g.h.i" --type "Service Provider" --password "hunter2"
```

That creates the following rows:

| email                    | name | description                                           | type             | password |
|--------------------------|------|-------------------------------------------------------|------------------|----------|
| info+abc@openapparel.org | abc  | Public facility lists for abc managed by the OAR team | Service Provider | hunter2  |
| info+d-ef@openapparel.org | D ef  | Public facility lists for D ef managed by the OAR team | Service Provider | hunter2  |
| info+g-h-i@openapparel.org | g.h.i  | Public facility lists for g.h.i managed by the OAR team | Service Provider | hunter2  |

jwalgran:
You got it 👍

PR #363, opened 2019-03-22
Include public list suffix when generating supplier names

jwalgran:
## Overview

We want to differentiate these accounts from real accounts that a contributor
may create in the future. This requirement was mistakenly omitted from the
initial feature request.

This suffix was approved by @katieashaw and @nataliegrillon 

Connects #240

## Demo

Screen Shot 2019-03-22 at 12 04 46 PM

## Testing Instructions

* Run `manage create_accounts --contributors abc "D ef" "g.h.i" --type "Brand/Retailer" --password hunter2048` and verify that the accounts created appear in the contributor list with the proper suffix.

### DIFF
--- a/src/django/api/management/commands/create_accounts.py
+++ b/src/django/api/management/commands/create_accounts.py
@@ -71,7 +71,7 @@ class Command(BaseCommand):
                 )
 
                 contributor = Contributor(
-                    name=name,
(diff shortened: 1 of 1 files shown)


jwalgran:
Thanks.

Issue #366, opened 2019-03-22
The parse task crashes when a CSV has fields that are too long for the database columns


jwalgran:
## Overview

When a name, address of a CSV row exceeds the length of the corresponding `FacilityListItem` field the call to `item.save()` causes a crash at the ORM level which kills the entire parse task.

### Expected Behavior

Lines in the CSV that violate the database constraints do not raise unhandled exceptions and a detailed message explaining the failure appears when viewing the list. The list item should be in the `ERROR_PARSING` status.

### Actual Behavior

The entire parse process fails, leaving items in the "UPLOADED" status.

### Steps to Reproduce

* Log in
* Upload [row-2-address-201-characters.csv.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/2998380/row-2-address-201-characters.csv.txt)
* Run the `batch_process` `parse` action. An error will be raised.

PR #367, opened 2019-03-22
Run Django model validation when parsing CSV lines

jwalgran:
## Overview

Prior to this PR a CSV row with a value that exceeded the length of one of the `FacilityListItem` fields was causing the parse task to fail with an unhandled
exception at the ORM level. By running a `full_clean` on the model we can catch
and report the validation errors on that specific line and safely save them to
the database where the contributor can see them.

Connects #366

## Demo

Screen Shot 2019-03-22 at 2 49 33 PM

## Testing Instructions

This assumes you have run `resetdb`. Adjust the list IDs accordingly if you have not.

* Log in
* Upload [row-2-address-201-characters.csv.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/2998380/row-2-address-201-characters.csv.txt)
* Parse the uploaded list and verify that parsing does not crash
* View the list and verify that the problematic line is in the `ERROR_PARSING` status and has an error message like the demo screenshot above.

### DIFF
--- a/src/django/api/processing.py
+++ b/src/django/api/processing.py
@@ -11,6 +11,7 @@ from unidecode import unidecode
 
 from django.conf import settings
 from django.contrib.gis.geos import Point
+from django.core.exceptions import ValidationError
 from django.db.models import Q
 
 from api.constants import CsvHeaderField, ProcessingAction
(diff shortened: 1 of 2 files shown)


jwalgran:
Thanks for the speedy review.

Issue #344, opened 2019-03-20
Faclility list download incomplete after switching to server-side paging


jwalgran:
## Overview

After switching to server side paging (#341) only a portion of the list is downloadable

### Expected Behavior

Clicking the Download CSV button downloads the entire list

### Actual Behavior

The downloaded CSV only contains the current page of results

### Steps to Reproduce

* Log in as c8@example.com
* Browse http://localhost:6543/lists/8
* Click Download CSV

PR #369, opened 2019-03-22
Restore facility list items CSV download

kellyi:
## Overview

In #341 we temporarily broke downloading facility list items CSVs by switching to server side paging. This PR restores the download functionality by adding a Redux action that:

- creates a list of all the URLs for the paginated CSV list items, with 100 items per page
- maps over that list with a Promise.all that resolves everything into an array of the `results` values from each facilities list item page
- concats that into an array structured like the CSV download functionality expected before
- downloads it as it did before

The pagination and map operation preserves the order of the list items, so the CSV will end up running from 0 - 1385 or whatever sequentially.

I also added some unit tests to check for large-ish and small-ish list item counts. In particular:

- for 1385 items we want to request 14 pages
- for < 100 items, we want to request 1 page

I also added a loading indicator to the download button since it now takes longer than it did, especially for longer lists.

Connects #344

## Testing Instructions

- serve this branch then download a few CSVs of varying sizes and ensure that the resulting downloads has a complete set of the list items and that they are ordered correctly

### DIFF
--- a/src/app/src/__tests__/utils.tests.js
+++ b/src/app/src/__tests__/utils.tests.js
@@ -53,6 +53,8 @@ const {
     sortFacilitiesAlphabeticallyByName,
     updateListWithLabels,
     makeSubmitFormOnEnterKeyPressFunction,
+    makeFacilityListItemsRetrieveCSVItemsURL,
+    makeFacilityListDataURLs,
 } = require('../util/util');
 
 const {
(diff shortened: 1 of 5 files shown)


kellyi:
> Is the lack of message intentional?

Yeah this was intentional -- I wasn't sure whether or where to display it I'll add it somewhere and push a fixup.

kellyi:
@jwalgran c4c7681a620133941b852af6d3b30423212d93b4
 adds a generic error message display above the buttons:

Screen Shot 2019-03-22 at 8 04 30 PM

Issue #361, opened 2019-03-22
Links to the map page that set query string arguments do not work


jwalgran:
## Overview

The "My facililes" links in the profile menu and the facility list page don't work.

### Expected Behavior

The "My facilities" links in the profile menu and on the facility list page run a search for all the facilities from the current contributor.

### Actual Behavior

The current search state overrides the new state.

### Steps to Reproduce

* Log in as c8@example.com
* Filter by Contributor Type = Auditor. 
* Click the "My facilities" link in the profile menu. The search does not change
* Click the "My lists" link in the profile menu
* From the lists page click the "View my facilities" link. The map page shows the wrong search results

### Demo

![2019-03-22 10 50 31](https://user-images.githubusercontent.com/17363/54843173-03d88080-4c91-11e9-9803-bf169f1cacb8.gif)

### Additional context

The `withQueryStringSync` component is overwriting the query string with the values in the app state.

kellyi:
I believe we can fix this by clearing the set filters before pushing the new link onto the history stack.

kellyi:
I think something like this will work:

```
commit 261e4a4f0ec86846d0093379b0c1c55f5ead12f4
Author: Kelly Innes <kinnes@azavea.com>
Date:   Fri Mar 22 14:33:30 2019 -0400

    WIP

diff --git a/src/app/src/components/FacilityLists.jsx b/src/app/src/components/FacilityLists.jsx
index 4c36b70..6237d1c 100644
--- a/src/app/src/components/FacilityLists.jsx
+++ b/src/app/src/components/FacilityLists.jsx
@@ -15,6 +15,8 @@ import {
     resetUserFacilityLists,
 } from '../actions/facilityLists';
 
+import { resetAllFilters } from '../actions/filters';
+
 import { authLoginFormRoute } from '../util/constants';
 import { makeMyFacilitiesRoute } from '../util/util';
 import { facilityListPropType } from '../util/propTypes';
@@ -36,6 +38,7 @@ class FacilityLists extends Component {
             userHasSignedIn,
             fetchingSessionSignIn,
             myFacilitiesRoute,
+            navigateToMyFacilities,
         } = this.props;
 
         if (fetching || fetchingSessionSignIn || error || !userHasSignedIn) {
@@ -90,13 +93,12 @@ class FacilityLists extends Component {
             <AppOverflow>
                 <AppGrid title="My Lists">
[...]

PR #370, opened 2019-03-23
Fix "My Facilities" navigation bug

kellyi:
## Overview

Previously clicking a "My Facilities" link would not work as expected
because the `?contributors=id` querystring would get vaporized by the
`withQueryStringSync` component before it became a filter in the the
Redux store. Since there were two links -- one from the list page, the
other from the navbar dropdown -- there are two fixes here:

First, for the lists page link I removed a condition from
`withQueryStringSync` that would skip hydrating from the querystring in
`componentDidMount` if there were already some filters in the app.

Secondly, for the navbar dropdown I replaced the `Link` element with a
call to `setFiltersFromQueryString`, giving it a custom querystring
parameter, and an ensuring call to `fetchFacilities`.

This second thing feels a little hacky, but the alternative I considered was
writing a brand new function that was exactly like `setFiltersFromQueryString`
except that it took a parameter of a different form and was meant to be
used only for internal links. The existing function seemed to work well
enough for now.

Connects #361 

## Demo

dropdown-link

list-page-link

regular-searches

## Testing Instructions

- serve this branch, then verify that the bugs described in #361 have been fixed
- use the search inputs on the search tab and verify that they work as before: the querystring should remain in sync with the selected filters
- use the "Share this Search" button to get a search querystring and open it in a new tab and verify that that works

### DIFF
--- a/src/app/src/components/Navbar.jsx
+++ b/src/app/src/components/Navbar.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import AppBar from '@material-ui/core/AppBar';
 import Toolbar from '@material-ui/core/Toolbar';
-import { Link } from 'react-router-dom';
+import { Link, Route } from 'react-router-dom';
 
 import logo from '../styles/images/OpenApparelRegistry_logo.png';
 import NavbarDropdown from './NavbarDropdown';
@@ -84,7 +84,7 @@ export default function Navbar() {
                     </Link>
                 </div>
                 <div id="google_translate_element" />
-                <NavbarLoginButtonGroup />
+                <Route component={NavbarLoginButtonGroup} />
             </Toolbar>
         </AppBar>
     );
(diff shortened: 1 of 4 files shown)


src/app/src/components/Navbar.jsx
```diff
@@ -84,7 +84,7 @@ export default function Navbar() {
                     </Link>
                 </div>
                 <div id="google_translate_element" />
-                <NavbarLoginButtonGroup />
+                <Route component={NavbarLoginButtonGroup} />
```
jwalgran: Confirming that we are using <Route> here so that we get access to the history object in the component lifecycle methods.

kellyi: Yeah, that's right. I also tried out using the `withRouter` decorator but I needed `history.push` to be available in the `mapStateToDispatch` function, which didn't seem to happen with `withRouter`.

Issue #368, opened 2019-03-22
Large files raise 504 timeout errors when trying to upload them to staging but do not show a helpful error


jwalgran:
## Overview

A clear description of the bug.

### Expected Behavior

If there is a 504 timeout, the error message shown says that it took to long to upload the file.

### Actual Behavior

When there is a 504 timeout the error message only says "An error prevented uploading that file"

### Steps to Reproduce

* Log into staging
* Attempt to upload 
[large-test-file-4316.csv.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/2998430/large-test-file-4316.csv.txt)

### Additional context

Ultimately we want to change or optimize the upload process in some way so that we are not susceptible to these timeouts. For the time being, we need a message that explains more clearly the type of failure we encountered.

jwalgran:
On staging there were several copies of the large file queued up to be processed, which indicates that the Django code may have kept running and completed the insert and job scheduling after the timeout. This makes things trick to deal with.

I did some testing and found that the problematic file took 3 seconds to import in development with the `develop` branch, and 1 second after I used `bulk_create` rather than individual model saves. We should be using `bulk create` in this situation, but I think the full solution is temporarily dropping the file somewhere and having a Batch job slice it up without any time pressure.

Issue #336, opened 2019-03-19
Do not try an create a Batch array job if there is only one facility in a list


jwalgran:
The AWS Batch client raises an error if we attempt to submit an array job with only 1 item.

View details in Rollbar: [https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/35/](https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/35/)

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 126, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/viewsets.py", line 116, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 495, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 455, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 492, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/contextlib.py", line 74, in inner
    return func(*args, **kwds)
  File "/usr/local/src/api/views.py", line 422, in create
    submit_jobs(ENVIRONMENT, new_list)
  File "/usr/local/src/api/aws_batch.py", line 108, in submit_jobs
    is_array=True)
  File "/usr/local/src/api/aws_batch.py", line 76, in submit_job
    'action': action,
  File "/usr/local/lib/python3.7/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/python3.7/site-packages/botocore/client.py", line 661, in _make_api_call
    raise error_class(parsed_response, operation_name)
ClientException: An error occurred (ClientException) when calling the SubmitJob operation: Error executing request, Exception : Array Job size must be greater than 1, RequestId: 6de31865-4a71-11e9-aeed-5321bfd6e706
```

PR #371, opened 2019-03-23
Bulk create list items and processing results / Process single row CSVs

jwalgran:
## Overview

Large list uploads to staging were timing out. For a list with 4000 items, we were running 4,000 insert statements to create rows and 12,000 individual update queries
just to append the AWS Batch job IDs to the `processing_results`. Switching to `bulk_create` and raw SQL we can drop from 16,000 queries down to 4.

We were also failing to process single-row CSVs because AWS Batch does not allow creating array jobs with a single row. We now use array jobs conditionally.

Connects #336 
Connects #368

## Demo

Screen Shot 2019-03-22 at 10 47 29 PM

Screen Shot 2019-03-22 at 10 40 40 PM

Screen Shot 2019-03-22 at 10 56 56 PM

## Notes

The tradeoffs involved with using raw SQL are breaking out of the ORM, which we mitigate by using `_meta` to dynamically fetch the table name, and no longer writing the job ID with a row index suffix the way it appears in the AWS Batch console, which is not really required because we have access to the `row_id` and we are adding an additional `is_array` flag to the processing result.

## Testing Instructions

All testing needs to be done on staging because the timeouts only appeared on staging and the fixes involve AWS Batch processing.

I pushed this branch as `test/bulk-create-list-items` to deploy it to staging and create the screenshots

I used these two files for testing 

[OAR Amfori  3.22 JW double length.csv.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/2999063/OAR.Amfori.3.22.JW.double.length.csv.txt)

[amfori-one-row.csv.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/2999062/amfori-one-row.csv.txt)

After successfully uploading the 8600+ row CSV I went into the AWS Batch console and terminated the related jobs to prevent day long geocoding.

### DIFF
--- a/src/django/api/aws_batch.py
+++ b/src/django/api/aws_batch.py
@@ -121,16 +121,18 @@ def submit_jobs(environment, facility_list):
 
     # GEOCODE
     started = str(datetime.utcnow())
+    row_count = facility_list.facilitylistitem_set.count()
+    is_array = row_count > 1
     geocode_job_id = submit_job('geocode',
                                 depends_on=[{'jobId': parse_job_id}],
-                                is_array=True)
+                                is_array=is_array)
     finished = str(datetime.utcnow())
     append_processing_result({
         'action': ProcessingAction.SUBMIT_JOB,
         'type': 'geocode',
         'job_id': geocode_job_id,
         'error': False,
-        'is_array': True,
+        'is_array': is_array,
         'started_at': started,
         'finished_at': finished,
     })
--- a/src/django/api/aws_batch.py
+++ b/src/django/api/aws_batch.py
@@ -1,9 +1,11 @@
 import boto3
+import json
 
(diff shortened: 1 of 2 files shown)


src/django/api/aws_batch.py
```diff
@@ -82,58 +88,66 @@ def submit_job(action, is_array=False, depends_on=None):
             raise RuntimeError(
                 'Failed to submit job {0}. Response {1}'.format(job_name, job))
 
+    def append_processing_result(result_dict):
+        query = ("UPDATE {item_table} "
```
jwalgran: We are not building our query with any user submitted data, so we are not opening ourselves up to SQL injection.

jwalgran:
Thanks for the review.

Issue #354, opened 2019-03-21
Swagger doc requests not working on staging due to http/https mixed content problem


kellyi:
## Overview

Swagger doc requests not working on staging due to http/https mixed content problem

### Expected Behavior

Interactive equests work

### Actual Behavior

Interactive requests don't work

PR #372, opened 2019-03-23
Help Django determine inbound request security

hectcastro:
## Overview

Set `SECURE_PROXY_SSL_HEADER` to use the AWS ALB `HTTP_X_FORWARDED_PROTO` header to determine whether the inbound HTTP request occurred via HTTPS.

Connects https://github.com/open-apparel-registry/open-apparel-registry/issues/354

## Demo

Screen Shot 2019-03-23 at 11 08 02

## Notes

See: https://docs.aws.amazon.com/elasticloadbalancing/latest/userguide/how-elastic-load-balancing-works.html#http-headers

## Testing Instructions

* Ensure this change set is on staging
* Navigate to https://staging.openapparel.org/api/docs/#!/facilities/facilities_count
* Click **Try it out!**
* Ensure the API call returns a legitimate response

### DIFF
--- a/src/django/oar/settings.py
+++ b/src/django/oar/settings.py
@@ -66,11 +66,15 @@ if ENVIRONMENT in ['Production', 'Staging'] and BATCH_MODE == '':
         response = response.json()
 
         for container in response['Containers']:
-          for network in container['Networks']:
-            for addr in network['IPv4Addresses']:
-              ALLOWED_HOSTS.append(addr)
+            for network in container['Networks']:
+                for addr in network['IPv4Addresses']:
+                    ALLOWED_HOSTS.append(addr)
     else:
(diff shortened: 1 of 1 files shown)


src/django/oar/settings.py
```diff
@@ -66,12 +66,16 @@
         response = response.json()
 
         for container in response['Containers']:
-          for network in container['Networks']:
```
hectcastro: These are just indentation fixes.

hectcastro:
Yep, we're safe on both of the last two bullets. Both CloudFront and ALB will remove the `X-Forwarded-Proto` header (`HTTP_X_FORWARDED_PROTO` is what Django `request.META` converts it to).

Issue #373, opened 2019-03-23
create_accounts command fails with longer contributor names


jwalgran:
## Overview

The `create_accounts` command fails with longer contributor names due to the fact that it is setting `username` (which we do not use in the application) which has a max length of 20.

### Expected Behavior

`create_accounts` succeeds when passed a long name like `"MEC (Material Manufacturers)"`

### Actual Behavior

`create_accounts` raises an DataError when passed a long name like `"MEC (Material Manufacturers)"`

### Steps to Reproduce

* `./scripts/manage create_accounts --type "Brand/Retailer" --password 1234pass --contributors "MEC (Material Manufacturers)"`

PR #374, opened 2019-03-23
Fix create_accounts management command

jwalgran:
## Overview

Our real data includes account names longer than 20 characters. We are not using the username field and it is left on the model only for compatibility so it is safe to not populate it.

Connects #373

## Testing Instructions

* Run `./scripts/manage create_accounts --type "Brand/Retailer" --password 1234pass --contributors "MEC (Material Manufacturers)"` and verify that it does not raise an exception.


jwalgran:
Thanks.

Issue #327, opened 2019-03-18
Show a summary status for an entire facility list


jwalgran:
## Overview

We need to clearly tell the contributor if they need to wait for the app to process their CSV, or if the app is waiting on the contributor to confirm or reject matches. 

### Describe the solution you'd like

We can use the area on the left side of the pagination toolbar to show status messages.

<img width="1160" alt="Screen Shot 2019-03-18 at 11 06 50 AM" src="https://user-images.githubusercontent.com/17363/54552911-5e5e9d80-496e-11e9-8a0d-6b163c6daa6c.png">

* If any rows are in a terminal error state, include that in the summary message.
   * "Processing failed on one or more rows"
* If all the rows are in either the `UPLOADED`, `PARSED`, or `GEOCODED` states, show a "wait" message.
* If any rows have matches in the `PENDING` status prompt the contributor to take action.
* If all the rows are in a terminal state, show a "finished processing" message. The terminal states are:
  * Automatic match
  * All proposed matches confirmed or rejected
  * New facility
  * Error

rajadain:
Since the front-end only ever gets one page worth of items, I believe the status for the entire list should be sent from the server. Doing so from the `/lists/8` endpoint seems most reasonable. My approach will be to add it as a computed field on `FacilityListSerializer`, perhaps `status`, with a set of declared states `ERROR`, `PROCESSING`, `PENDING` (or would `AWAITING` be better?), and `COMPLETE`. The front-end will switch the displayed message on these states.

Calculating the status for a list on every API call may be expensive in the long run, and we can consider making this a field on the `FacilityList` model. Would that be a premature optimization?

PR #378, opened 2019-03-25
Add Summary Status for Facility Lists

rajadain:
## Overview

Adds a field to the facility list detail endpoint that is a DISTINCT list of all item statuses in the facility list. This list is inspected in the front-end and is used to show a status message in the pagination row.

Connects #327 

### Demo

Processing:

image

Error:

image

## Testing

* Go to [:6543](http://localhost:6543/) and log in
* Upload a list
  - [ ] Ensure the list details page shows a summary status above the list items in the table
* `manage batch_process` the list
  - [ ] Ensure the list details page shows the correct summary status at each step

### DIFF
--- a/src/app/src/__tests__/utils.tests.js
+++ b/src/app/src/__tests__/utils.tests.js
@@ -55,6 +55,7 @@ const {
     makeSubmitFormOnEnterKeyPressFunction,
     makeFacilityListItemsRetrieveCSVItemsURL,
     makeFacilityListDataURLs,
+    makeFacilityListSummaryStatus,
 } = require('../util/util');
 
 const {
@@ -66,6 +67,8 @@ const {
     DEFAULT_PAGE,
     DEFAULT_ROWS_PER_PAGE,
     ENTER_KEY,
+    facilityListItemStatusChoicesEnum,
+    facilityListSummaryStatusMessages,
 } = require('../util/constants');
 
 it('creates a route for checking facility list items', () => {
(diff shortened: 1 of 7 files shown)


src/django/api/serializers.py
```diff
@@ -171,6 +172,42 @@ def get_items_url(self, facility_list):
...
+            FacilityListItem.CONFIRMED_MATCH,
+        ])
+
+        item_statuses = set(
+            facility_list.facilitylistitem_set
+            .values_list('status', flat=True)
+            .distinct())
+
+        if (item_statuses.intersection(ERROR_STATUSES)):
+            status = 'ERROR'
```
jwalgran: A list can have both errors and pending items (many lists we have seen so far have some parse errors on individual lines. On average, the data is pretty dirty). We should consider returning  multiple additional "computed flags" from the serializer, one for the presence of an error and one for "waiting on you or waiting on us" processing status.

rajadain: Switched to send a distinct list of all item statuses from the back-end, and have the front-end choose messages accordingly.

src/app/src/util/util.js
```diff
@@ -453,3 +456,29 @@ export const makeFacilityListDataURLs = (id, count) => {
...
+        facilityListItemStatusChoicesEnum.PARSED,
+        facilityListItemStatusChoicesEnum.GEOCODED,
+        facilityListItemStatusChoicesEnum.GEOCODED_NO_RESULTS,
+    ].includes(s)) ?
+        facilityListSummaryStatusMessages.PROCESSING : '';
+    const completeMessage = statuses.every(s => [
+        facilityListItemStatusChoicesEnum.MATCHED,
+        facilityListItemStatusChoicesEnum.CONFIRMED_MATCH,
+    ].includes(s)) ?
+        facilityListSummaryStatusMessages.COMPLETED : '';
```
kellyi: Not sure how to suggest cleaning it up, but this series of chained ternaries is a little hard to follow. I think the test does enough to verify that it's working, but in theory we could revise it with something like https://lodash.com/docs/4.17.11#cond that would do a form of pattern matching, which is I think what we want here.

rajadain: `cond` does a single match, but here we want multiple matches.

kellyi: Now that I'm looking at it again, this could also be a reduce operation that just rolls through the list and figures out what to do on each pass.

This is probably fine to leave as is assuming we're confident that the test is working, which it seems to be!

rajadain:
How does this look:

Black and bold:

image

vs

Black only:

image

vs

Bold only:

image

rajadain:
Added "Bold only" styling in 6762406.

kellyi:
Cool. Bold only looks great!

jwalgran:
Looking at this now

src/app/src/util/util.js
jwalgran: 👍

rajadain:
> One way of dealing with this would be to re-fetch the list details using the existing API and update the state.

Alright I'll look into this. May be tricky to keep the pagination state in place when refreshing the larger list.

src/app/src/__tests__/utils.tests.js
```diff
@@ -1022,3 +1025,47 @@ it('creates a list of data URLs for retrieving facility list items data', () =>
         smallerExpectedURLs,
     )).toBe(true);
 });
+
+it('creates a summary status message given a list of facility list item statuses', () => {
```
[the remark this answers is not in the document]
rajadain: Added in 1f683e3.

rajadain:
Added ff57c65 which updates the summary status as the user is interacting with the items in the list.

The manner it is implemented in makes the shape of the data from the confirm and reject endpoints different from the other list item details endpoints, but perhaps that is okay for now.

This is ready for another look.

jwalgran:
Looks like a reasonable solution. Testing now

rajadain:
Squashed fixups. Merging now. Thanks for all the feedback!

Issue #383, opened 2019-03-25
Text in the Guide panel needs replacing


katieashaw:
## Overview

The copy in the Guide panel on the site is legacy copy, which needs replacing. 

### Expected Behavior

New copy needs to appear in the Guide panel on the preview site. Please replace with copy in this doc: https://docs.google.com/document/d/1uz1q6ORlNjknoMx7g1WYNzTBhFM2hLR_TSgqt_TTezU/edit

### Actual Behavior

Legacy copy is appearing in the Guide panel on the preview site.

### Steps to Reproduce

* Visit preview.openapparel.org
* Click the guide panel

### Demo

Optional. Screenshots, gifs, `curl` examples, console errors, log files, etc.

### Additional context

Optional. Add any other context about the problem here.

kellyi:
Shares a point with #376

dboyer:
Use the text:

The Open Apparel Registry (OAR) is an open source tool which maps garment facilities worldwide and assigns a unique ID number to each. Its goal is to become the go-to source for identifying apparel facilities and their affiliations by collating disparate supplier lists from industry stakeholders into one central, open source map and database.

The collated database of facility names, addresses and affiliated parties is powered by an advanced name and address-matching algorithm and will be available for use by any organization, for free.

To contribute to the database, users must create a free account. Anyone can sign up and contribute. Users interested in browsing the OAR are able to access it without creating an account.

Issue #382, opened 2019-03-25
Text formatting on Facility List page is odd


katieashaw:
## Overview

The formatting of new text on the Facility List page is appearing justified right.

### Expected Behavior

The additional text should be justified left, under the other text on the page.

### Actual Behavior

The additional text is justified right. 

### Steps to Reproduce

* Log into a user account
* Navigate to https://preview.openapparel.org/lists/2

### Demo

![Screenshot 2019-03-25 at 16 48 22](https://user-images.githubusercontent.com/46344729/54939559-c853e580-4f20-11e9-89f0-f950538de23c.png)

### Additional context

Optional. Add any other context about the problem here.

kellyi:
Shares a point with #376

Issue #381, opened 2019-03-25
There's a typo in the troubleshooting text on the Contribute page


katieashaw:
## Overview

A word is mis-spelt in the troubleshooting tips at the bottom of the Contribute page

### Expected Behavior

Separate should have an 'a" after the "p"

### Actual Behavior

The "a" after "p" in separate is missing. 

The font in this section also looks odd, so would be worth checking. 

### Steps to Reproduce

* Visit https://preview.openapparel.org/contribute and scroll to the bottom of the page to the Troubleshooting text. 

### Demo

![Screenshot 2019-03-25 at 16 48 02](https://user-images.githubusercontent.com/46344729/54939142-ba519500-4f1f-11e9-9a0b-50b59486fafc.png)

### Additional context

Optional. Add any other context about the problem here.

kellyi:
Shares a point with #376

Issue #376, opened 2019-03-25
Redirect to the home page after logging in


jwalgran:
## Overview

Redirect to the home page after logging in

### Is your feature request related to a problem? Please describe.

Comment from a tester

>I got confused too when i logged in, was sent to profile page, and I saw an empty password field at the bottom with a required red star, I too thought i had to log in, again.

### Describe the solution you'd like

Redirect to the home page after logging in

PR #385, opened 2019-03-25
Miscellaneous UI fixes

kellyi:
## Overview

- Fix contributor form troubleshooting typo / Connects #381
- Left-align facility list items guide text snippet / Connects #382 
- Show country name in facility details address / Connects #375 
- Reroute to facilities on login / Connects #376 
- Conditionally redirect to facilities map on logout / Connects #343 
- Restrict map bounds to tile extent / Connects #365 
- Autofocus first user input on forms / Connects #273 
- Update Guide tab panel text / Connects #383 

## Testing

- for each of the issues above, verify that the UI issue has been remedied by the changes here.

### DIFF
--- a/src/app/src/components/ContributeTroubleshooting.jsx
+++ b/src/app/src/components/ContributeTroubleshooting.jsx
@@ -10,7 +10,7 @@ const ContributeTroubleshooting = memo(() => (
                 </p>
                 <ul className="helper-list">
                     <li className="helper-list__item">
-                        Check that your CSV file uses commas to seprate the
+                        Check that your CSV file uses commas to separate the
                         individual fields, not semcolons, tabs, or other
                         characters.
                     </li>
--- a/src/app/src/components/ControlledTextInput.jsx
+++ b/src/app/src/components/ControlledTextInput.jsx
@@ -13,6 +13,7 @@ export default function ControlledTextInput({
     placeholder,
     disabled,
     submitFormOnEnterKeyPress,
+    autoFocus,
 }) {
     return (
         <Fragment>
@@ -20,6 +21,7 @@ export default function ControlledTextInput({
                 {hint}
             </p>
             <input
+                autoFocus={autoFocus} // eslint-disable-line jsx-a11y/no-autofocus
                 type={type}
                 id={id}
                 className="noFocus form__text-input"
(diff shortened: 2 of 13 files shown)


kellyi:
@designmatty for https://github.com/open-apparel-registry/open-apparel-registry/pull/385/commits/57a4836d1e3d2e7a91c5a8854354e634a2b064a4 I had to disable an ESLint rule about not using the `autoFocus` prop.

In this case, I am using it to focus the first field of form inputs -- which is, I think, the correct behavior -- but can you concur that it's okay from an accessibility perspective for me to disable the linting rule for that specific line in this specific circumstance?

rajadain:
Seeing this error in the console:

image

kellyi:
That error's from a bug or quirk of how react-router sees the NoRouteMatch component. I fixed it in a commit on this PR:

https://github.com/open-apparel-registry/open-apparel-registry/pull/353

kellyi:
Specifically: https://github.com/open-apparel-registry/open-apparel-registry/pull/353/commits/669d756b9720b70eebcf07facfb422f7c95eef86

Unfortunately that error is also currently on develop, but it's non-breaking and it's coming from a library, not our code.

rajadain:
This may be worth its own card, but here's an odd behavior I noticed:

* Be logged out
* Search for a facility
* See it in the search results
* Log in
* Be redirected to search results, but now they are empty
* Search again to see them populated



If instead of being redirected to the search results, I was redirected to an empty state, that would've been better. Or, the search results should've stuck around.

kellyi:
> This may be worth its own card, but here's an odd behavior I noticed:

Makes sense. Can you write this up as its own card? This is how things work on `develop`, too.

kellyi:
I think the fix might be to reset the selected tab in the UIReducer when necessary so it'll default back to the search tab.

rajadain:
I don't see a login link when going to a facility detail page if I'm logged out, as asked for in #343 

image

kellyi:
> I don't see a login link when going to a facility detail page if I'm logged out

Thanks for catching. I added this in 

c6b80cc8bf15fd18660d4fafc22c02782813d936

Screen Shot 2019-03-25 at 4 43 16 PM

kellyi:
Cool, thank you for testing!

kellyi:
I updated the commit message for the commit which includes the autoFocus prop and turns off the jsx-a11y linting for the relevant line. We checked into it and because we are using the prop to make the forms behave as normal forms, it appears to be a blessed use of the `autoFocus` prop there.

Issue #343, opened 2019-03-20
Logging out from the list detail page shows a "no list found for that ID" message / browsing list when logged out shows technical error


jwalgran:
## Overview

Logging out from the list detail page shows a "no list found for that ID" message

Browsing a list page when not logged in shows a technical error message.

### Expected Behavior

The when logging out from a list page the app redirects to the map page

When visiting a list page when not logged in a login link is shown

### Actual Behavior

A not found message is displayed after logging out on a list page

A technical error is shown when browsing a list page when logged out.

### Steps to Reproduce

* Log in as c3@example.com
* Browse http://localhost:6543/lists/3
* Log out
* Refresh

### Demo

<img width="947" alt="Screen Shot 2019-03-20 at 9 44 40 AM" src="https://user-images.githubusercontent.com/17363/54702835-e1116500-4af4-11e9-85de-1f8787361fa4.png">

<img width="710" alt="Screen Shot 2019-03-20 at 10 23 41 AM" src="https://user-images.githubusercontent.com/17363/54705590-6a776600-4afa-11e9-9b8b-fb5d1d37a42b.png">

kellyi:
Shares a point with #376

PR #387, opened 2019-03-25
Use && rather than || for logout redirect logic

kellyi:
## Overview

I noticed a bug in the logout redirect logic which is that a redirect
should happen if

- the current route is NOT the main route, '/'
AND
- the current route does NOT start with '/facilities`, as it would for
the facilities version of the main route and the facility details route

The previous OR logic would be true and cause a redirect if just
`pathname !== mainRoute` was true, which meant it would redirect from
the facilities and facility details routes.

Connects #343 

## Testing Instructions

- serve this branch and login
- on the `/` route, logout and verify that you are not redirected
- from `/facilities`, logout and verify that you are not redirected
- from a facilities details page, logout and verify that you are not redirected

From the following pages, logout and verify that you are redirected:

- contribute
- profile
- lists
- list details

### DIFF
--- a/src/app/src/components/NavbarLoginButtonGroup.jsx
+++ b/src/app/src/components/NavbarLoginButtonGroup.jsx
@@ -161,7 +161,7 @@ function mapDispatchToProps(dispatch, {
         logout: () => {
             dispatch(submitLogOut());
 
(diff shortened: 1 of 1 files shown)


jwalgran:
Looking at this now

kellyi:
Thanks!

Issue #394, opened 2019-03-26
Send back Google Analytics key with web/environment.js var


kellyi:
## Overview

Update the app to send the Google Analytics key back with the web/environment.js var, the same way we do with the Google Maps Key and the Rollbar key.

### Describe the solution you'd like

In the client, `window.ENVIRONMENT.REACT_APP_GOOGLE_ANALYTICS_KEY` is available as a working GA key to use for #267

PR #395, opened 2019-03-26
Google Analytics Environment Variable

rajadain:
## Overview

Wires up Google Analytics code as an environment variable. Also updated `terraform.tfvars` in the production bucket with the real value.

I did not add it to the staging bucket because we should not be tracking that.

The final commit is for demoing, and will be dropped before the merge.

Connects #394 

## Notes

When wiring up Google Tag Manager, we should do something like what's done for Rollbar, where we check if the key is available:

https://github.com/open-apparel-registry/open-apparel-registry/blob/d9978b79474a6b06e4fd18846a9e919199c5b475/src/app/public/index.html#L40-L42

This will ensure that analytics are not captured for dev / test environments.

## Testing Instructions

* Check out this branch
* Add a value to `.env.backend` like this:

      REACT_APP_GOOGLE_ANALYTICS_KEY=UA-123456789-10

* Run `cibuild` and `server`
  - [x] Go to [:6543/](http://localhost:6543/) and ensure you see it logged to the console.
  - [x] Go to [:8081/](http://localhost:8081/) and ensure you see it logged to the console.

### DIFF
--- a/deployment/terraform/batch.tf
+++ b/deployment/terraform/batch.tf
@@ -63,6 +63,8 @@ data "template_file" "default_job_definition" {
     django_secret_key          = "${var.django_secret_key}"
     google_server_side_api_key = "${var.google_server_side_api_key}"
 
+    rollbar_server_side_access_token = "${var.rollbar_server_side_access_token}"
+
     aws_region = "${var.aws_region}"
   }
 }
--- a/deployment/terraform/batch.tf
+++ b/deployment/terraform/batch.tf
@@ -64,8 +64,6 @@ data "template_file" "default_job_definition" {
     google_server_side_api_key = "${var.google_server_side_api_key}"
 
     aws_region = "${var.aws_region}"
-
-    rollbar_client_side_access_token = "${var.rollbar_client_side_access_token}"
   }
 }
 
--- a/deployment/terraform/task-definitions/app.json
+++ b/deployment/terraform/task-definitions/app.json
@@ -16,6 +16,7 @@
         { "name": "DEFAULT_FROM_EMAIL", "value": "${default_from_email}" },
         { "name": "GOOGLE_SERVER_SIDE_API_KEY", "value": "${google_server_side_api_key}" },
         { "name": "REACT_APP_GOOGLE_CLIENT_SIDE_API_KEY", "value": "${google_client_side_api_key}" },
+        { "name": "REACT_APP_GOOGLE_ANALYTICS_KEY", "value": "${google_analytics_key}" },
         { "name": "ROLLBAR_SERVER_SIDE_ACCESS_TOKEN", "value": "${rollbar_server_side_access_token}" },
         { "name": "REACT_APP_ROLLBAR_CLIENT_SIDE_ACCESS_TOKEN", "value": "${rollbar_client_side_access_token}" }
     ],
--- a/deployment/terraform/variables.tf
+++ b/deployment/terraform/variables.tf
@@ -198,6 +198,10 @@ variable "google_server_side_api_key" {}
 
 variable "google_client_side_api_key" {}
 
+variable "google_analytics_key" {
+  default = ""
+}
+
 variable "rollbar_server_side_access_token" {}
 
 variable "rollbar_client_side_access_token" {}
--- a/src/app/public/index.html
+++ b/src/app/public/index.html
@@ -32,6 +32,15 @@
     <title>Open Apparel Registry (beta)</title>
     <!-- Environment Variables -->
     <script src="%PUBLIC_URL%/web/environment.js"></script>
+    <!-- TODO: Replace with actual Google Tag Manager snippet -->
+    <script>
+      if (window.ENVIRONMENT &&
+          window.ENVIRONMENT.REACT_APP_GOOGLE_ANALYTICS_KEY &&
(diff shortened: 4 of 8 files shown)


deployment/terraform/variables.tf
```diff
@@ -198,6 +198,8 @@ variable "google_server_side_api_key" {}
 
 variable "google_client_side_api_key" {}
 
+variable "google_analytics_key" {}
```
rbreslow: We need a default value here otherwise the `$ terraform plan` operation will be stuck at an interactive prompt. 

```suggestion
variable "google_analytics_key" {
  default = ""
}
```

src/app/public/index.html
```diff
@@ -32,6 +32,10 @@
     <title>Open Apparel Registry (beta)</title>
     <!-- Environment Variables -->
     <script src="%PUBLIC_URL%/web/environment.js"></script>
+    <!-- TODO: Replace with actual Google Tag Manager snippet -->
```
rbreslow: Going on my first comment, maybe we should have some check to see if `REACT_APP_GOOGLE_ANALYTICS_KEY` is `null`, and if it is, we shouldn't inject anything into the page. 

Is something like this possible?

deployment/terraform/variables.tf
rajadain: Ah, so this is how defaults are specified, good to know! 👌

src/app/public/index.html
rajadain: Yes. I added a demo of that in 0ef53f2.

src/app/public/index.html
```diff
@@ -32,6 +32,15 @@
...
+    <script>
+      if (window.ENVIRONMENT &&
+          window.ENVIRONMENT.REACT_APP_GOOGLE_ANALYTICS_KEY &&
+          window.ENVIRONMENT.REACT_APP_GOOGLE_ANALYTICS_KEY !== "" &&
+          window.ENVIRONMENT.ENVIRONMENT !== 'development') {
```
rajadain: This is here to demo how to implement this in production, but this line should be commented out to see the logging in development.

deployment/terraform/batch.tf
```diff
@@ -62,6 +62,7 @@ data "template_file" "default_job_definition" {
     environment                = "${var.environment}"
     django_secret_key          = "${var.django_secret_key}"
     google_server_side_api_key = "${var.google_server_side_api_key}"
+    google_analytics_key       = "${var.google_analytics_key}"
```
rbreslow: This isn't being used as it isn't within the job definition JSON file.

I think we can remove this altogether though, since we're not passing any other client side environment variables to the job definition.

deployment/terraform/task-definitions/app.json
```diff
@@ -16,6 +16,7 @@
         { "name": "DEFAULT_FROM_EMAIL", "value": "${default_from_email}" },
         { "name": "GOOGLE_SERVER_SIDE_API_KEY", "value": "${google_server_side_api_key}" },
         { "name": "REACT_APP_GOOGLE_CLIENT_SIDE_API_KEY", "value": "${google_client_side_api_key}" },
+        { "name": "REACT_APP_GOOGLE_ANALYTICS_KEY", "value": "${google_analytics_key}" }
```
rbreslow:
```suggestion
        { "name": "REACT_APP_GOOGLE_ANALYTICS_KEY", "value": "${google_analytics_key}" },
```

kellyi:
> I'm at a loss for how the React development server is reading environment variables from .env.backend though. 

These are coming in from a web/environment.js file injected by Django.

kellyi:
I tested this out and it's working as described! I think you can feel free to include the demo commit -- since I'll use it -- but change the commit name message to indicate that it is in place for subsequent work.

rbreslow:
>These are coming in from a web/environment.js file injected by Django.

🤦‍♂️. I keep getting confused when I see `.env.frontend`, but I realize that's a remnant from the legacy API.

kellyi:
@rbreslow this is the further-background context for how this is working for the front-end variables:

https://github.com/open-apparel-registry/open-apparel-registry/pull/212

deployment/terraform/batch.tf
rajadain: Done in the latest version. Also removed `rollbar_client_side_access_token` from this file.

rajadain:
Just force pushed all recommendations. Ready for another look.

deployment/terraform/batch.tf
```diff
@@ -64,8 +64,6 @@ data "template_file" "default_job_definition" {
     google_server_side_api_key = "${var.google_server_side_api_key}"
 
     aws_region = "${var.aws_region}"
-
-    rollbar_client_side_access_token = "${var.rollbar_client_side_access_token}"
```
rbreslow: Just realized, we should be threading through the `rollbar_server_side_access_token` here, as well as in the job definition so that exceptions in Batch jobs are surfaced to Rollbar.  

Also happy with splitting this out into a separate issue.

rajadain: Added that in af2c914.

rajadain:
Thanks for taking a look! Merging now.

Issue #267, opened 2019-03-06
Clicking "Accept" on compliance notification should opt users in to Google Analytics tracking


dboyer:
## Overview

If users click reject in the cookie notification implemented as part of #218, we need to not track any info in Google Analytics

### Is your feature request related to a problem? Please describe.

GDPR requires an option to opt out of analytics tracking

### Describe the solution you'd like

If a user clicks reject, we need to not send any of their information to Google Analytics or store tracking cookies

### Additional Context

These posts may have helpful info:

https://www.blastam.com/blog/gdpr-need-consent-for-google-analytics-tracking
https://www.blastam.com/blog/5-actionable-steps-gdpr-compliance-google-analytics

dboyer:
We should use Google Tag Manager - https://www.portent.com/blog/analytics/gtm-gdpr-consent.htm

kellyi:
It does not appear that Google Analytics is set up in the application at all yet, so we may need to do that as part of this work too.

kellyi:
After discussing this we determined that the proper behavior is to only start GA tracking after the user has clicked "Accept" on the notification bar rather than "always unless the user has clicked 'Reject'". Going to update the title to reflect that change.

kellyi:
@dboyer reminder that we also need some updated text for the notification dialog.

PR #397, opened 2019-03-26
Set accept/reject cookies value in localStorage via utility functions

kellyi:
## Overview

This is a preliminary PR for #267 which updates the GDPRNotification component to set specific values for `GA_TRACKING` in localstorage via some testable utility functions. 

Clicking "Accept" will set `GA_TRACKING` to `HAS_ACCEPTED_GA_TRACKING`.

Clicking "Reject" will set `GA_TRACKING` to `HAS_REJECTED_GA_TRACKING`.

I also added some tests which check the value of localStorage, using a mock version of localStorage described here: https://github.com/facebook/jest/issues/2098#issuecomment-395233511

Prior to writing the custom mock for localStorage, I tried out a Jest plugin. It turns out that Create React App does not let you use Jest plugins without ejecting, ergo the custom mock.

Connects #267

## Testing Instructions

- get this branch then run `./scripts/test --app` to verify the new tests pass
- `./scripts/server`, visit localhost:6543 and run `localStorage.clear()` in the browser console when the app loads
- refresh the page and click "Reject" in the notification Snackbar
- refresh the page again
- type `localStorage` into the browser console and verify that the object looks like:

```
Storage { GA_TRACKING: "HAS_REJECTED_GA_TRACKING", length: 1 }
```

- run `localStorage.clear()` again, then refresh the page
- this time choose "Accept" from the Snackbar, then type `localStorage` and verify that the state is now:

```
Storage { GA_TRACKING: "HAS_ACCEPTED_GA_TRACKING", length: 1 }
```

- refresh the page, then type `localStorage` again and verify that the state is still

```
Storage { GA_TRACKING: "HAS_ACCEPTED_GA_TRACKING", length: 1 }
```

### DIFF
--- a/src/app/src/util/util.ga.js
+++ b/src/app/src/util/util.ga.js
@@ -1,0 +1,27 @@
+export const GA_TRACKING = 'GA_TRACKING';
+export const HAS_ACCEPTED_GA_TRACKING = 'HAS_ACCEPTED_GA_TRACKING';
+export const HAS_REJECTED_GA_TRACKING = 'HAS_REJECTED_GA_TRACKING';
+
+export const userHasAcceptedOrRejectedGATracking = () =>
+    Boolean(localStorage.getItem(GA_TRACKING));
+
+export const userHasAcceptedGATracking = () =>
+    localStorage.getItem(GA_TRACKING) === HAS_ACCEPTED_GA_TRACKING;
+
+export const userHasRejectedGATracking = () =>
+    localStorage.getItem(GA_TRACKING) === HAS_REJECTED_GA_TRACKING;
+
+export const rejectGATracking = () => {
+    localStorage.setItem(GA_TRACKING, HAS_REJECTED_GA_TRACKING);
+};
+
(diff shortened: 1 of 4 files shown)


src/app/src/util/util.ga.js
```diff
@@ -0,0 +1,27 @@
...
+};
+
+export const acceptGATracking = () => {
+    // If user has already rejected GA tracking, this is a noop
+    if (!userHasRejectedGATracking()) {
+        localStorage.setItem(GA_TRACKING, HAS_ACCEPTED_GA_TRACKING);
+    }
+};
+
+export const clearGATracking = () => {
```
jwalgran: Consider naming this `clearGATrackingDecision` so we are not implying that this actually clears tracking data.

kellyi: Sounds good -- will do!

kellyi: Added in force-pushed amended commit a186a04ddf5e72987abed8678fc4bb8bdeeab13b

Issue #389, opened 2019-03-25
Overlapping content in the facility list


jwalgran:
[This is a draft issue that will be updated with more details on how to reproduce the problem in development]

<img width="1026" alt="Screen Shot 2019-03-25 at 3 59 55 PM" src="https://user-images.githubusercontent.com/17363/54959713-33002380-4f17-11e9-8810-7e8e9a4a1790.png">

designmatty:
I've been looking into this issue a bunch this morning. This happens because we are setting a `height: 55px` on the divs within the table cells. We define this height to force alignment across each cell in a given row.

When the string bleeds beyond this 55px height we see the issue above.

If we remove the 55px height, then the cells no longer line up and it becomes very difficult to know which actions belong to which row. I have a potential solution in the form of truncating/tooltips. I'll be pushing a PR soon.

katieashaw:
This bug is visible on the Puma T1 list dated 23rd June 2020:

![Screenshot 2020-06-23 at 16 33 08](https://user-images.githubusercontent.com/46344729/85424909-6b563d80-b570-11ea-8203-59f77cfd73bc.png)

PR #398, opened 2019-03-26
WIP: Tooltip truncate component

designmatty:
## Overview

This is an initial stab at fixing the problems stated in the connected issue. This component will check the length of the provided string, and if it is over 40 characters, the string truncates with an ellipsis and renders a `<Tooltip>` on hover of the full string.

Connects #389 

## Demo

Screen Shot 2019-03-26 at 12 51 16 PM

## Testing Instructions

* Open a list
* Find an item with a long address and hover on it.
* Does a tooltip show up?

### DIFF
--- a/src/app/src/App.css
+++ b/src/app/src/App.css
@@ -572,7 +572,12 @@ hr {
   overflow: auto;
 }
 
-
+.text-overflow {
+  text-overflow: ellipsis;
+  overflow: hidden;
+  white-space: nowrap;
+  display: block;
+}
 
 /* @keyframes App-logo-spin {
   from { transform: rotate(0deg); }
--- a/src/app/src/components/CellElement.jsx
+++ b/src/app/src/components/CellElement.jsx
@@ -6,11 +6,12 @@ import Dialog from '@material-ui/core/Dialog';
 import DialogActions from '@material-ui/core/DialogActions';
 import DialogContent from '@material-ui/core/DialogContent';
 import DialogTitle from '@material-ui/core/DialogTitle';
+import Tooltip from '@material-ui/core/Tooltip';
 import isObject from 'lodash/isObject';
 
 import { facilityMatchStatusChoicesEnum } from '../util/constants';
-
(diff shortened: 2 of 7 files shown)


jwalgran:
Thank you for trying to make this work. The problem with tooltips is that for this particular operation, comparing two records to see if they are a match or not, it is vital that the contributor is able to see both full names and addresses simultaneously.

kellyi:
Not sure if we should do it here or in a separate changeset but it seems like we could drop or drastically narrow the CSV ROW INDEX column since it occupies a disproportionate amount of space:

Screen Shot 2019-03-29 at 1 36 17 PM

kellyi:
Seems like that gets partway to a fix:

Screen Shot 2019-03-29 at 1 43 36 PM

jwalgran:
The row index is a helpful reference when paging through a long list. I don't want to drop it, but our longest lists are in the thousands, so we only need enough width to show 4 characters. We could shorten the header to just "Row"

designmatty:
The only way I can imagine this getting fixed without the use of truncation/tooltips would require a complete rewrite of how we handle these tables. It would be a substantial amount of work and I don't feel we have the time to do it currently.

kellyi:
> The only way I can imagine this getting fixed without the use of truncation/tooltips would require a complete rewrite of how we handle these tables.
 
I'm going to take a small amount of time to see what changes can be made to the existing table to make this mostly workable on Macbook-sized screens. I think there is currently quite a bit of unused space in on the table screen and so I think it's possible we can get it most of the way there for now without undertaking a complete rewrite, which we can decide whether we want to do later.

designmatty:
@kellyi You might be able to do something where you get the string length, and if it is cover a specified amount, change the max-height of the containing div. That hard defined div height is what ultimately causes the overflow issues. If you'd like to pair on this at all, let me know.

jwalgran:
Thanks for taking the time to dig into this issue. I agree with the assessment that we need to restructure the table. We have decided to keep #389 in the backlog for now.

Issue #396, opened 2019-03-26
Facilities from an inactive list appear in search by contributor


jwalgran:
## Overview

Facilities from an inactive list appear in search by contributor

### Expected Behavior

When searching by contributor, facilities from inactive lists do not appear.

### Actual Behavior

When searching by contributor, facilities from inactive lists do appear.

### Steps to Reproduce

* Register a new account
* Upload 2-row csv [2019-03-26-2-row.csv.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/3009109/2019-03-26-2-row.csv.txt)
* Use `batch_process` to `parse`, `geocode`, and `match` the list.
* Upload 1-row csv and use the "replaces" option to deactivate the 2-row [2019-03-26-1-row.csv.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/3009110/2019-03-26-1-row.csv.txt)
* Use `batch_process` to `parse`, `geocode`, and `match` the list.
* On the map page, search by the new contributor account. 2 facilities appear instead of 1.

rajadain:
First attempt was to make sure that a Facility's `created_from` pointed to an active facility list, but that unfortunately doesn't work:

```python
>>> Facility.objects.filter(created_from__facility_list__is_active=True).filter(id__in=name_match_facility_ids)
<QuerySet []>
```

Because the facility's stay linked to the first list that they were created from, which is now `inactive`. Seeing next how that facility is related to the, `active` list.

jwalgran:
On the detail page for a facility we correctly suppress the relationship

This facility was created from the 2-item list. You can see the contributor is not shown
<img width="564" alt="Screen Shot 2019-03-26 at 10 11 09 AM" src="https://user-images.githubusercontent.com/17363/55018225-95a4fe00-4faf-11e9-97c2-aa0bc4e257f3.png">
https://staging.openapparel.org/facilities/CN20190857KAR17?contributors=135

We need to incorporate the same logic into the search filter.

PR #401, opened 2019-03-26
Do not search elements from inactive lists

rajadain:
## Overview

Do not include results from inactive lists in search.

Connects #396 

## Demo

image

## Testing Instructions

* Follow the instructions in #396 to try and recreate the bug
  - [x] Ensure you see only one facility, not two


rajadain:
Thanks for taking a look!

Issue #399, opened 2019-03-26
Redirect to facility list after upload


jwalgran:
## Overview

As of #378 we are showing the processing status on the list detail page. 

### Is your feature request related to a problem? Please describe.

It is not clear what the contributor should do next after they upload a file.

### Describe the solution you'd like

- After a file is successfully uploaded, redirect to the list page for that file.

PR #404, opened 2019-03-26
Redirect to list after upload

rajadain:
## Overview

Redirects to list detail view after upload completes successfully.

Connects #399 

## Testing Instructions

* Upload a list
  - [ ] Ensure you are redirected to the right list

### DIFF
--- a/src/app/src/actions/upload.js
+++ b/src/app/src/actions/upload.js
@@ -21,7 +21,7 @@ export const completeUploadFile = createAction('COMPLETE_UPLOAD_FILE');
 
 export const resetUploadState = createAction('RESET_UPLOAD_STATE');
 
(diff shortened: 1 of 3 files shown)


src/app/src/actions/upload.js
```diff
@@ -60,7 +60,10 @@ export function uploadFile(file = null) {
 
         return csrfRequest
             .post(makeFacilityListsURL(), formData)
-            .then(() => dispatch(completeUploadFile()))
+            .then(({ data: { id }}) => {
```
kellyi: Linter's complaining about `}}` after `id` here, suggesting there needs to be a space between like `id } })`

rajadain:
Thanks for taking a look!

Issue #265, opened 2019-03-06
Update text on GDPR notification


dboyer:
## Overview

Update text on GDPR compliance notification

### Is your feature request related to a problem? Please describe.

Current text is a placeholder and needs to be updated.

### Describe the solution you'd like

Text is still pending

dboyer:
Use this text:

The Open Apparel Registry uses cookies to collect and analyze site performance and usage. By clicking the Accept button, you agree to allow us to place cookies and share information with Google Analytics. For more information, please visit our Terms and Conditions of Use and Privacy Policy.

Link Terms and Conditions of Use and Privacy Policy to https://info.openapparel.org/tos/

jwalgran:
I removed the design label because this is just a copy update.

PR #409, opened 2019-03-27
Add Google Analytics tracking if and only if a user has clicked "Accept" on the notification

kellyi:
## Overview

This PR configures the application to add Google Analytics tracking if -- and *only* if -- a user has clicked "Accept" on the Notification. To make this easier to prove & verify, I opted to initialize the GA tracking conditionally in a utility function around which I was able to set up some testing apparatus.

If a user clicks "Reject", we set a value for `GA_TRACKING` in localStorage to `"HAS_REJECTED_GA_TRACKING"` and then that sticks until the user clears localStorage.

If a user clicks "Accept", we set a value for `GA_TRACKING` in localStorage to
`"HAS_ACCEPTED_GA_TRACKING"` and then set up tracking for that visit. On subsequent visits, the application will check for the `GA_TRACKING` value and if it sees that there is a value *AND* that that value is `"HAS_ACCEPTED_GA_TRACKING"` then tracking starts.

For the following conditions, the application does not start GA Tracking at all:

- `window.localStorage` is unavailable for whatever reason
- `GA_TRACKING` has not been set at all in `window.localStorage`
- `GA_TRACKING` is in `window.localStorage` but does not have a value of `"HAS_ACCEPTED_GA_TRACKING"`.

As an additional guard, I set an initial global opt out value per https://developers.google.com/analytics/devguides/collection/analyticsjs/user-opt-out which is only deleted when all the acceptance conditions have been met.

Connects #265 
Connects #267

## Demo

Screen Shot 2019-03-27 at 12 40 56 PM

## Notes

The second commit here is entirely for testing and I will drop it before merging this in. GA provides a helpful debug library which logs actions to the console so you can easily see when it is called.

## Testing Instructions

- serve this branch, then visit the application in Chrome
- clear your cookies and your localStorage
- refresh the page and verify that you see the new GDPR notification in the Snackbar
- open the network tab

#### Rejection testing

- click "Reject"
- verify that you do not see any requests made to google analytics in the network tab
- switch to the console and verify that you *DO NOT* see a message like the following logged amidst the Redux logger actions:

Screen Shot 2019-03-27 at 12 44 32 PM

- look at the application cookies and verify that no Google Analytics cookies have been set
- refresh the page and repeat these verification steps
- refresh a couple more times and repeat these verification steps

At no point prior to "Accepting" or after "Rejecting" should you see:

- requests made to Google Analytics in the Network tab
- cookies set by Google Anayltics

#### Acceptance testing

- clear cookies and localStorage
- refresh the app and click Accept
- verify that you do see a request made to GA in the network tab and that you also see something like this in the console with subequent info about settings:

Screen Shot 2019-03-27 at 12 44 32 PM

- verify that you see this in the console and that that instruction matches the guidance here: https://developers.google.com/analytics/devguides/collection/analyticsjs/ip-anonymization

Screen Shot 2019-03-27 at 12 47 20 PM

- refresh the page and verify that you see the Google Analytics stuff load again along with the `"anonymizeIp"` setting 
- refresh the page again and verify the same

##### No localStorage testing

- open Firefox, then open `about:config`
- Follow these instructions to disable localStorage:

```
In FireFox: Type “about:config” in your address bar and hit enter to view your internal browser settings. Scroll down to „dom.storage.enabled”, right click on it and hit „Toggle” to disable the DOM Storage.
```

- visit the application in Firefox and verify that it doesn't crash and that Google Analytics tracking does not start

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/__tests__/util.ga.tests.js
+++ b/src/app/src/__tests__/util.ga.tests.js
@@ -1,0 +1,91 @@
+/* eslint-env jest */
+/* eslint-disable object-shorthand */
+/* eslint-disable func-names */
+/* eslint-disable space-before-function-paren */
+
+import {
+    userHasAcceptedOrRejectedGATracking,
+    userHasAcceptedGATracking,
+    userHasRejectedGATracking,
+    rejectGATracking,
+    acceptGATrackingAndStartTracking,
+    clearGATrackingDecision,
+    createGADisableKey,
+} from '../util/util.ga.js';
+
+beforeEach(() => {
+    const window = {};
+
+    window.localStorage = {
+        store: {},
+        getItem: function(key) {
+            return this.store[key] || null;
+        },
+        setItem: function(key, value) {
+            this.store[key] = value.toString();
+        },
+        removeItem: function(key) {
+            delete this.store[key];
+        },
+        clear: function() {
+            this.store = {};
+        },
+    };
+
+    window.ENVIRONMENT = {
+        REACT_APP_GOOGLE_ANALYTICS_KEY: 'GA_KEY',
+    };
+
+    global.window = window;
+    jest.spyOn(global.window.localStorage, 'setItem');
+});
+
+afterEach(() => {
+    window.localStorage = null;
+});
+
+it('creates a `ga-disable-ID` key for opting out of tracking', () => {
+    const expectedGADisableKey = 'ga-disable-test-key';
+    const testGAKey = 'test-key';
+
+    expect(createGADisableKey(testGAKey)).toBe(expectedGADisableKey);
+});
+
+it('correctly sets a `HAS_REJECTED_GA_TRACKING` value in localStorage', () => {
+    rejectGATracking();
+
+    expect(window.localStorage.setItem).toHaveBeenCalledTimes(2);
+    expect(window.localStorage.setItem).toHaveBeenCalledWith('GA_TRACKING', 'HAS_REJECTED_GA_TRACKING');
+    expect(window.localStorage.getItem('GA_TRACKING')).toBe('HAS_REJECTED_GA_TRACKING');
+
+    expect(userHasAcceptedOrRejectedGATracking()).toBe(true);
+    expect(userHasRejectedGATracking()).toBe(true);
+    expect(userHasAcceptedGATracking()).toBe(false);
+});
+
+it('correctly sets a `HAS_ACCEPTED_GA_TRACKING` value in localStorage', () => {
+    acceptGATrackingAndStartTracking();
+
+    expect(window.localStorage.setItem).toHaveBeenCalledTimes(2);
+    expect(window.localStorage.setItem).toHaveBeenCalledWith('GA_TRACKING', 'HAS_ACCEPTED_GA_TRACKING');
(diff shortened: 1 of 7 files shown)


src/app/src/__tests__/util.ga.tests.js
```diff
@@ -0,0 +1,91 @@
...
+
+    window.ENVIRONMENT = {
+        REACT_APP_GOOGLE_ANALYTICS_KEY: 'GA_KEY',
+    };
+
+    global.window = window;
+    jest.spyOn(global.window.localStorage, 'setItem');
+});
+
+afterEach(() => {
```
rajadain: Should we be restoring `global.window` once we're finished?

src/app/src/components/GDPRNotification.jsx
```diff
@@ -1,23 +1,43 @@
-import React, { PureComponent } from 'react';
+import React, { Component } from 'react';
 import Snackbar from '@material-ui/core/Snackbar';
 import Button from './Button';
 import ShowOnly from './ShowOnly';
 import COLOURS from '../util/COLOURS';
 
-class GDPRNotification extends PureComponent {
```
rajadain: Why change from `PureComponent` to `Component`? (I'm not familiar with the subtle differences)

src/app/src/__tests__/util.ga.tests.js
kellyi: I accidentally marked this as resolved but I don't think so. Given that window.localStorage is also re-assigned, I think setting it to `null` here kind of doesn't matter anymore for the tests.

src/app/src/components/GDPRNotification.jsx
kellyi: My understanding of PureComponent is that it has some characteristics which memoize it in certain circumstances, but I don't quite know whether that's true or what those circumstances are. In the interest of being able to reason about this a little more accurately I switched it out for a regular Component.

src/app/src/util/util.ga.js
```diff
@@ -0,0 +1,122 @@
...
+export const GA_TRACKING = 'GA_TRACKING';
+export const HAS_ACCEPTED_GA_TRACKING = 'HAS_ACCEPTED_GA_TRACKING';
+export const HAS_REJECTED_GA_TRACKING = 'HAS_REJECTED_GA_TRACKING';
+
+
+export const maybeGetReactAppGoogleAnalyticsKey = () => {
+    const gaKey = env(REACT_APP_GOOGLE_ANALYTICS_KEY);
+    const environment = env('ENVIRONMENT');
+
+    if (!environment || environment !== 'development') {
```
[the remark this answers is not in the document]
rajadain: This is part of the test commit that will be dropped before merging. It's supposed to be `if (environment === development) return null;`.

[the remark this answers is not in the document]
kellyi: Yeah, sorry for not explaining well. I'll drop c68fabfa56b3e2a55addc9443e9915eb461ad863 before merging this, but it is essentially just to be able to test this in local development. For production and staging it will be set to `=== 'development'` -> `return null`

kellyi:
Thanks for the feedback and help with this!

Issue #424, opened 2019-03-28
Return 400 from API endpoints when the datatype of one or more parameters is incorrect


jwalgran:
## Overview

Return a 400 response rather than an unhandled 500 when the datatype of one or more parameters is incorrect.

### Describe the solution you'd like

- Validate the API request parameters and raise a suitable error that is serialized and returned as a 400.

### Additional Context

View details in Rollbar: [https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/67/](https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/67/)

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 126, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/viewsets.py", line 116, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 495, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 455, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 492, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/src/api/views.py", line 445, in list
    .filter(facility_list_item__facility_list__contributor__id__in=contributors) # NOQA
  File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 839, in filter
    return self._filter_or_exclude(False, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 857, in _filter_or_exclude
    clone.query.add_q(Q(*args, **kwargs))
  File "/usr/local/lib/python3.7/site-packages/django/db/models/sql/query.py", line 1253, in add_q
    clause, _ = self._add_q(q_object, self.used_aliases)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/sql/query.py", line 1277, in _add_q
    split_subq=split_subq,
[...]

PR #433, opened 2019-03-28
Raise 400 error for invalid contributor param type

kellyi:
## Overview

Raise a 400 error if not all of the contributor querystring parameters
are integers, sending back a detailed error response about the type.

The other custom parameters already expect strings, so they don't fail
in the same way, and the pagination params have their own error
handling. Therefore we don't need to add similar handling to the other
errors.

Connects #424 

## Demo

Screen Shot 2019-03-28 at 4 25 35 PM

Screen Shot 2019-03-28 at 4 25 48 PM

## Testing Instructions

- resetdb then processfixtures
- serve this branch
- visit http://localhost:8081/api/docs/#!/facilities/facilities_list and try entering a string value in the contributors input, then search and verify that it returns 400
- change the string value to and int, like 2, and verify that it returns 200 and that you see a FeatureCollection
- visit the main app on 6543 then search for facilities in various ways and verify that that also still works

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Return 400/Bad Request error for /api/facilities request with invalid contributor parameter type. [#433](https://github.com/open-apparel-registry/open-apparel-registry/pull/433)
+
 ### Security
 
 ## [2.0.0] - 2019-03-27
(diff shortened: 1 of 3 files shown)


kellyi:
Good point. It appears that we can do this using a serializer on its own -- https://tinyendian.com/articles/parsing-query-parameters-in-rest-framework/ -- so I will update this to do that.

kellyi:
bbf02ed8e9b4a5bc5f6da1ac6ddbcbbcf06791c7 is an amended commit that users a custom serializer + the built-in ValidationError for handling errors related to the queryparams

kellyi:
Rebased on `develop`

kellyi:
Thanks!

kellyi:
> This lead me to an edge case:

Oof, I merged the PR before I saw this. Will make a new PR to fix.

Issue #443, opened 2019-04-01
Place ErrorBoundary around FacilitiesMap component & log crashes to Rollbar


kellyi:
## Overview

- enable the map component to crash without crashing the whole app by setting up an appropriately placed `componentDidCatch` method somewhere in the tree
- manually log errors to Rollbar when the map crashes, as we do with other kinds of errors

### Describe the solution you'd like

Enable the FacilitiesMap component to crash when necessary -- for example, when a browser with no access to the Google Maps API attempts to create a map -- by placing an ErrorBoundary around the component. This potentially could involve just adding a `componentDidCatch` method in the FacilitiesMap itself, or adding a `componentDidCatch` method to the MapAndSidebar component which can do some backup rendering.

When the map component crashes we should still send the error to Rollbar in order to keep having visibility into the errors.

PR #446, opened 2019-04-01
Enable FacilitiesMap to crash without crashing the rest of the app

kellyi:
## Overview

Place a new error boundary in the MapAndSidebar component which will enable to map to crash for whatever reason without crashing the entire app. Map errors are logged to Rollbar conditionally using the same mechanism we use in the ErrorBoundary component, and application crashes issuing from elsewhere still take down the app and get caught by the ErrorBoundary.

Second commit here is just for testing and demos what happens when the map crashes when trying to render.  You can also try out throwing an error in its `componentDidUpdate` method with similar effect.

Connects #443 

## Demo

Screen Shot 2019-04-01 at 1 06 16 PM

## Testing Instructions

- serve this branch, then load the application and verify that you see the error thrown in the console and the UI updated as depicted above to handle the map crash
- add an error in the FilterSidebar component's `render` method and verify that that error does take down the whole app and that you see the standard error message UI displayed

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/components/FacilitiesMapErrorMessage.jsx
+++ b/src/app/src/components/FacilitiesMapErrorMessage.jsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import Typography from '@material-ui/core/Typography';
+
+const mapErrorMessageStyles = Object.freeze({
+    errorMessageContainerStyles: Object.freeze({
+        height: '100%',
+        width: '100%',
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'center',
+    }),
+});
+
+export default function FacilitiesMapErrorMessage() {
+    return (
+        <div style={mapErrorMessageStyles.errorMessageContainerStyles}>
+            <Typography variant="title">
+                An error prevented loading the map.
+            </Typography>
+        </div>
+    );
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Add a new error boundary to enable the FacilitiesMap component to crash without crashing the rest of the app [#446](https://github.com/open-apparel-registry/open-apparel-registry/pull/446)
+
 ### Security
 
 ## [2.1.0] - 2019-04-01
--- a/src/app/src/components/MapAndSidebar.jsx
+++ b/src/app/src/components/MapAndSidebar.jsx
@@ -1,10 +1,11 @@
-import React, { Fragment } from 'react';
+import React, { Component, Fragment } from 'react';
 import { Route, Switch } from 'react-router-dom';
 import Grid from '@material-ui/core/Grid';
 
 import FilterSidebar from './FilterSidebar';
 import FacilityDetailsSidebar from './FacilityDetailSidebar';
 import FacilitiesMap from './FacilitiesMap';
+import FacilitiesMapErrorMessage from './FacilitiesMapErrorMessage';
 
 import '../styles/css/Map.css';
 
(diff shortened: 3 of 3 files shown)


src/app/src/components/FacilitiesMapErrorMessage.jsx
```diff
@@ -0,0 +1,22 @@
...
+    errorMessageContainerStyles: Object.freeze({
+        height: '100%',
+        width: '100%',
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'center',
+    }),
+});
+
+export default function FacilitiesMapErrorMessage() {
```
kellyi: I did not memoize this because doing so issues a linter error from the `Route component` call in which it is instantiated which I think comes from a quirk with how our version of react-router sees memoized components.

kellyi:
CI is failing because the demo commit is a linter error because everything after its change is now unreachable code.

kellyi:
Rebased on develop.

jwalgran:
Looking at this now. Sorry for the delay

kellyi:
No worries! Thanks for taking a look.

src/app/src/components/FacilitiesMapErrorMessage.jsx
```diff
@@ -0,0 +1,22 @@
...
+    return (
+        <div style={mapErrorMessageStyles.errorMessageContainerStyles}>
+            <Typography variant="title">
+                An error prevented loading the map!
```
jwalgran: Personal preference to not exclaim the error report.

kellyi:
> When I browse on :6543 I briefly see the error boundary message, then drop to the crash screen

This one's expected because it the overlay comes from the Create React App development setup.

> Is this expected behavior?

This double-logging is not-quite the expected behavior but it appears that there might be a logger somewhere in the production build which uses `console.error` rather than the manual `console.warn`. If it's worthwhile I can drop the manual warning from https://github.com/open-apparel-registry/open-apparel-registry/pull/446/files#diff-ae486d67e1af14cf34cbb2f194b78ea2R20

src/app/src/components/FacilitiesMapErrorMessage.jsx
kellyi: :+1: Changed in 9e27d85620b6f99f207d26bbd65e5523951b7ba9

jwalgran:
Please confirm my interpretation, but it looks like we are not actually catching the error, just changing the way we render. If that is the case, I misunderstood what error boundaries do and we should not manually console.warn or manually call rollbar, because our built-in exception handling will take care of it.

kellyi:
> Please confirm my interpretation, but it looks like we are not actually catching the error, just changing the way we render.

I believe this is correct. The existing ErrorBoundary component, for instance, responds to the error by adjusting the app display, except in that case it replaces the entire app with the text of the error message rendered using `btoa`.

kellyi:
https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html#introducing-error-boundaries

> Error boundaries are React components that catch JavaScript errors anywhere in their child component tree, log those errors, and display a fallback UI instead of the component tree that crashed. Error boundaries catch errors during rendering, in lifecycle methods, and in constructors of the whole tree below them.

kellyi:
83cb71fdab8e181952abab148d434f302d81791f drops the logging and removes the `componentDidCatch` method. `getDerivedStateFromError` is the preferred error boundary lifecycle method now: https://reactjs.org/docs/react-component.html#error-boundaries

jwalgran:
Sorry for the back-and-forth but I am now seeing no errors logged to the console or rollbar when browsing :8081

Screen Shot 2019-04-04 at 10 46 43 AM

It looks like we may need to restore manually logging to rollbar.

kellyi:
> It looks like we may need to restore manually logging to rollbar.

No worries! I dropped the previous last fixup commit which removed the Rollbar error and pushed 13e46bf01fcf05631a44cedccabdcd04a1fdca9f which just drops the console.warn.

kellyi:
Squashed and force pushed everything except the demo error in 
719444c

jwalgran:
Just need to clean up the build failure
```
 19:37  error  'error' is defined but never used  no-unused-vars
```

PR #451, opened 2019-04-02
Update release checklist to keep default commit messages

rbreslow:
We should keep the default commit messages instead of overriding them. 

image

vs 

image

There was a miscommunication at first, we just want to make sure that the body of the "merge tag" commit contains the tag, which it already does by default.


rbreslow:
Sorry, I missed the notification that this was approved!

Issue #321, opened 2019-03-15
Update Django geocoding tests to check for correct keys rather than identical content.


kellyi:
## Overview

Some of the geocoding tests are set up to expect specific content for the geocoding results. This causes them to fail intermittently when the actual values differ in some way from the expectations we've hard-coded into the app.

Adjust these tests so that they check on the keys of the geocoding responses rather than the values.

### Describe the solution you'd like

Adjust geocoding tests so that they check on the keys of the geocoding responses rather than the values.

Geocoding tests should pass if the responses have the correct keys/shapes.

PR #466, opened 2019-04-10
Test that geocode reponses contain expected keys

jwalgran:
## Overview

We have repeatedly run into geocoding tests breaking because the specifics of the response data change. To make the test less brittle we are now only verifying that the keys we expect are present in the response.

Connects #321

## Testing Instructions

* Verify that `./scripts/test` pass

## Checklist

- [x] CI passes after rebase


jwalgran:
Thanks for the quick review. I snuck in a commit that cleans up some mistakenly merged debug printing.

a623296746a599a6894ad28bdc67104f5b64257d

Issue #504, opened 2019-05-02
Allow facility list items to be filtered by status


jwalgran:
## Overview

Allow facility list items to be filtered by status.

### Describe the solution you'd like

* Add a `status` parameter to the `/api/facility-lists/{id}/items/` API endpoint
  * Filter the results based on the supplied `status` value. Be mindful of the `NEW_FACILITY` meta status.
* Add a dropdown that allows contains a default "All Statuses" as well as an entry for each possible status or meta-status. Selecting an item should submit the new `status` parameter to the API and update the results.
* Add a reset / show all button that resets the dropdown to "All Statuses" (Shared with #503)

PR #507, opened 2019-05-08
Add an API parameter and UI to filter facility list items by status

jwalgran:
## Overview

Allow facility list items to be filtered by selecting one or more statuses. It is an uncommon situation in a thousand item list to only have a dozen potential matches. Having a filter makes it dramatically easier to find and act on these items.

Connects #504 

## Demo
Screen Shot 2019-05-08 at 10 31 47 AM

Screen Shot 2019-05-08 at 10 31 59 AM

Screen Shot 2019-05-08 at 10 32 32 AM

## Testing Instructions

* Reset the database 
  * `./scripts/resetdb`
  * `./scripts/processfixtures`
* Log in as `c8@example.com`
* Filter by different statuses and verify that the list results match what is expected.
* Filter by `POTENTIAL_MATCH` and approve one of the matches. Verify that the item disappears from the table and the item count is correctly updated.
* Approve or reject the rest of the potential matches and verify that "No matching items" is displayed in the table.
* Verify that clicking the "SHOW ALL ITEMS" button works.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/components/FacilityListItemsTable.jsx
+++ b/src/app/src/components/FacilityListItemsTable.jsx
@@ -8,6 +8,8 @@ import Paper from '@material-ui/core/Paper';
 import Table from '@material-ui/core/Table';
 import TableHead from '@material-ui/core/TableHead';
 import TableBody from '@material-ui/core/TableBody';
+import TableRow from '@material-ui/core/TableRow';
+import TableCell from '@material-ui/core/TableCell';
 import TablePagination from '@material-ui/core/TablePagination';
 import ReactSelect from 'react-select';
 import ShowOnly from './ShowOnly';
@@ -221,6 +223,8 @@ class FacilityListItemsTable extends Component {
             </Grid>
         );
 
+        const listIsEmpty = !fetchingItems && items && items.length === 0;
+
         const tableRows = fetchingItems ? [] : items
             .map((item) => {
                 const handleSelectRow = makeSelectListItemTableRowFunction(item.row_index);
@@ -353,6 +357,13 @@ class FacilityListItemsTable extends Component {
                         </TableHead>
                         <TableBody>
                             {tableRows}
+                            <ShowOnly when={listIsEmpty}>
+                                <TableRow>
+                                    <TableCell colSpan={5} style={{ textAlign: 'center' }}>
+                                        No matching items.
+                                    </TableCell>
+                                </TableRow>
+                            </ShowOnly>
                         </TableBody>
                     </Table>
                 </div>
--- a/src/app/src/components/FacilityListItemsTable.jsx
+++ b/src/app/src/components/FacilityListItemsTable.jsx
@@ -24,6 +24,7 @@ import {
     makePaginatedFacilityListItemsDetailLinkWithRowCount,
     getValueFromEvent,
     createPaginationOptionsFromQueryString,
+    createParamsFromQueryString,
     makeFacilityListSummaryStatus,
 } from '../util/util';
 
@@ -53,6 +54,7 @@ const facilityListItemsTableStyles = Object.freeze({
 function FacilityListItemsTable({
     list,
     items,
+    filteredCount,
     fetchingItems,
     selectedFacilityListItemsRowIndex,
     makeSelectListItemTableRowFunction,
@@ -74,13 +76,16 @@ function FacilityListItemsTable({
         rowsPerPage,
     } = createPaginationOptionsFromQueryString(search);
 
+    const params = createParamsFromQueryString(search);
+
     const handleChangePage = (_, newPage) => {
         push(makePaginatedFacilityListItemsDetailLinkWithRowCount(
             listID,
             (newPage + 1),
             rowsPerPage,
+            params,
         ));
-        fetchListItems(listID, newPage + 1, rowsPerPage);
+        fetchListItems(listID, newPage + 1, rowsPerPage, params);
     };
 
     const handleChangeRowsPerPage = (e) => {
@@ -88,8 +93,9 @@ function FacilityListItemsTable({
             listID,
             page,
             getValueFromEvent(e),
+            params,
         ));
-        fetchListItems(listID, page, getValueFromEvent(e));
+        fetchListItems(listID, page, getValueFromEvent(e), params);
     };
 
     const paginationControlsRow = (
@@ -113,7 +119,7 @@ function FacilityListItemsTable({
                 md={6}
             >
                 <TablePagination
-                    count={list.item_count}
+                    count={filteredCount}
                     rowsPerPage={Number(rowsPerPage)}
                     rowsPerPageOptions={rowsPerPageOptions}
                     onChangeRowsPerPage={handleChangeRowsPerPage}
@@ -249,6 +255,7 @@ FacilityListItemsTable.defaultProps = {
 
 FacilityListItemsTable.propTypes = {
     items: arrayOf(facilityListItemPropType),
+    filteredCount: number.isRequired,
     match: shape({
         params: shape({
             listID: string.isRequired,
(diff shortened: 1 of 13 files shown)


jwalgran:
I'd like to discuss our options for dealing with the situation where you have filtered the list to see only `POTENTIAL_MATCH`es and then you approve a match. Our current implementation updates the item in place, changing the status to `MATCHED`, creating an inconsistency with the current state.

The solution I am thinking about implementing is to pass the `page`, `rowsPerPage` and `status` query string arguments to the approve/reject endpoints and modify the view to return an entire updated page of items instead of a single item if a `page` query string parameter was specified. That allows us to easily update the state and keep the data in the table in sync with the filter.

Does this sound like a good plan?

kellyi:
> The solution I am thinking about implementing is to pass the page, rowsPerPage and status query string arguments to the approve/reject endpoints and modify the view to return an entire updated page of items instead of a single item if a page query string parameter was specified. That allows us to easily update the state and keep the data in the table in sync with the filter.

This makes sense to me. I was thinking about the general problem outside of this particular solution and came up with a similar-ish idea that put the re-fetch-page logic on the frontend.

The client "knows" in some way which filter is applied, so if the `POTENTIAL_MATCH` filter was applied and an accept/reject request had been completed, one of the `componentDidUpdate` methods could check the new items list against the previous items list to see whether it still had the same number of `POTENTIAL_MATCH` status facilities. If not, then it could immediately re-fetch the entire page.

The potential upside of this is that it would not require adding conditional logic to the API endpoints; a potential downside is that it'd push the re-fetch logic into a React component's `componentDidUpdate` method. Not sure which is preferable.
 
If we were to put the conditional logic in the API accept/reject handlers, I think that it'd make sense to have the client have to pass an additional querystring parameter along with page/rowsPerPage/status: something like `returnEntireListPage=true` or similar, so that the querystring params would be a little more self-documenting.

jwalgran:
Detecting the change on the client side and making a follow up request is compelling. 

As I was thinking about implementing that I realized that if the contributor is scrolled to the bottom of a page of 100 results when they approve a pending match, triggering a page refresh will result in scrolling them back to the top of the page. We may be able to work around it by avoiding or updating `startFetchFacilityListItems` so that we don't set the existing list of items to null before loading the new ones.

https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/reducers/FacilityListDetailsReducer.js#L145-L148

I think I'll move ahead with the client-side approach.

jwalgran:
I implemented refreshing after approving/rejecting potential matches. Thanks for the design suggestion. It worked out well.

This is ready for review.

src/app/src/components/FacilityListItemsTable.jsx
```diff
@@ -73,251 +73,285 @@ const createSelectedStatusChoicesFromParams = (params) => {
...
-        push(makePaginatedFacilityListItemsDetailLinkWithRowCount(
-            listID,
-            (newPage + 1),
-            rowsPerPage,
-            params,
-        ));
-        fetchListItems(listID, newPage + 1, rowsPerPage, params);
-    };
+        const params = createParamsFromQueryString(search);
+        if (params.status && params.status.includes('POTENTIAL_MATCH') && items && prevProps.items) {
```
kellyi: Do we have `POTENTIAL_MATCH` set as a constant somewhere in the app? If so we could consider checking on the constant value here; if not, we could add one?

src/app/src/components/FacilityListItemsTable.jsx
```diff
@@ -73,251 +73,285 @@ const createSelectedStatusChoicesFromParams = (params) => {
...
+            push(makePaginatedFacilityListItemsDetailLinkWithRowCount(
+                listID,
+                1,
+                rowsPerPage,
+                newParams,
+            ));
+            fetchListItems(listID, 1, rowsPerPage, newParams);
+        };
+
+        const handleShowAllClicked = () => {
```
kellyi: Now that this is a class component I think it'd make sense to move these functions out of `render` and make them class methods. I think they could just be cut and paste, and then callers updated to call them with `this.whatever`.

src/app/src/components/FacilityListItemsTable.jsx
kellyi: Looks like it's set on the `facilityListItemStatusChoicesEnum`.

src/app/src/util/util.js
```diff
@@ -195,7 +195,9 @@ export const createParamsFromQueryString = (qs) => {
     } = querystring.parse(qsToParse);
 
     if (status) {
-        return Object.freeze({ status });
+        return Array.isArray(status)
+            ? Object.freeze({ status })
+            : Object.freeze({ status: [status] });
```
kellyi: Consider adding updating the test to check for this new outcome.

src/app/src/util/util.js
```diff
@@ -185,6 +185,22 @@ export const createPaginationOptionsFromQueryString = (qs) => {
     });
 };
 
+export const createParamsFromQueryString = (qs) => {
```
kellyi: Consider adding a test for this new function.

src/app/src/components/FacilityListItemsTable.jsx
```diff
@@ -48,199 +56,313 @@ const facilityListItemsTableStyles = Object.freeze({
...
+                                address="Address"
+                                status="Status"
+                                hover={false}
+                            />
+                        </TableHead>
+                        <TableBody>
+                            {tableRows}
+                            <ShowOnly when={listIsEmpty}>
+                                <TableRow>
+                                    <TableCell colSpan={5} style={{textAlign: 'center'}}>
```
kellyi: Looks like civicci's seeing a linter error here resolvable by adding a space before and after `textAlign: 'center'`.

kellyi:
Question! Is it correct that selecting the `MATCHED` option should display both `MATCHED` and `NEW_FACILITY` status items? I know the latter is a computed status, but didn't know whether we intended for `MATCHED` to now be inclusive of `NEW_FACILITY`:

Screen Shot 2019-05-13 at 11 01 42 AM

kellyi:
It looks like a similar thing is happening with `CONFIRMED_MATCH` and `NEW_FACILITY` status:

Screen Shot 2019-05-13 at 11 05 06 AM

kellyi:
This is working as expected otherwise!

src/app/src/components/FacilityListItemsTable.jsx
jwalgran: I did some reading and discovered that defining closures inside `render` like this is a performance drag because since redefining these functions every time  triggers unnessesary rendering. Thanks for suggestion.

src/app/src/components/FacilityListItemsTable.jsx
jwalgran: Addressed in fixup 0aeb695

src/app/src/util/util.js
jwalgran: Added tests in fixup b1e998a

src/app/src/util/util.js
jwalgran: Added tests in fixup b1e998a

src/app/src/components/FacilityListItemsTable.jsx
jwalgran: Addressed in fixup 02efe99

src/app/src/components/FacilityListItemsTable.jsx
jwalgran: Addressed in fixup ca65c8c

jwalgran:
Thanks for the great review. I addressed all of your comments and this is ready for another look.

jwalgran:
Thanks again for the great feedback.

Issue #457, opened 2019-04-03
Improve the visibility and resiliency of AWS Batch processing of CSVs


jwalgran:
## Overview

AWS Batch jobs can fail. We designed the processing to be idempotent, so it is safe to rerun failed jobs, but we don't have an automated system for doing so.

### Is your feature request related to a problem? Please describe.

You cannot see that a job has failed within the application, only in the AWS Batch console.

### Describe the solution you'd like

- Evaluate whether or not the built-in AWS Batch job retry feature is appropriate for our workflow.
- Re-run failed jobs a fixed number of times.
- Alert development/operations when jobs fail (Reference: https://docs.aws.amazon.com/batch/latest/userguide/batch_sns_tutorial.html)

hectcastro:
I would be interested in discussing this a bit with whoever pulls it. In RF we've made use of:

- Job retries (which require some additional application logic to track the retry count)
- Job timeouts (to prevent jobs that hang from hogging up resources)
- Send Rollbar notifications when jobs fail (A Ruby Goldberg device that uses part of the strategy in SNS article above, but to wake up a Lambda that invokes Rollbar)

PR #512, opened 2019-05-14
Add AWS Batch job retries and failure reporting to Rollbar via AWS Lambda

jwalgran:
## Overview

- Set all jobs to retry up to 3 times in an attempt to automatically recover from transient issues that may cause a job to fail.
- Set up job failure notifications by setting up a CloudWatch event to detect failures and connecting that event to a Lambda function that posts the details of the failure to Rollbar. 

Connects #457

### Tasks being retried
Screen Shot 2019-05-15 at 1 47 07 PM

## Job failure reported in Rollbar
Screen Shot 2019-05-15 at 1 47 31 PM

## Notes

The batch processing management command is designed to only process items in a
specific state so rerunning a partially completed job will not process any items
multiple times.

Posting job failures to Rollbar via a Lambda function has been use successfully in another project.

## Testing Instructions

This must be deployed to staging as batch processing only works in AWS.

There is a testing commit that makes jobs fail 80% of the time that will be dropped prior to merging.

This was tested on staging by manually deploying 71045fd. 

I verify that dependent jobs do not fail until all of the retries of the parent job have also failed.

* Register and confirm an account
* Upload a facility list
  * Verify in the AWS Batch console that failed jobs are retried.
  * Verify that job failures are reported to Rollbar.

## Checklist

- [ ] CI passes after rebase


src/django/api/management/commands/batch_process.py
```diff
@@ -40,6 +40,11 @@ def add_arguments(self, parser):
                            help='The id of the facility list to process.')
 
     def handle(self, *args, **options):
+        # TODO: REMOVE TESTING CODE #####
```
hectcastro: Reminder to remove before merge.

jwalgran:
Thanks for the review, and for running `terraform fmt`. I squashed that into the main terraform commit.

Issue #519, opened 2019-05-15
Add django-simple-history library for auditing


kellyi:
## Overview

Add django-simple-history library for auditing. We'll also need to determine which models should have the historical audit trail.

### Describe the solution you'd like

In #500 I looked into django-simple-history and suggested that we use that for OAR audit trails. Here we should add it.

I don't know that we need to check in any particular models now, but it may be worthwhile to add a history to the Facility model initially.

### Additional Context

Adding django-simple-history to a model will create a `Historical...` sibling model which stores some of the same data in order to track it and make it revert-able. We will need to figure out a solution for handling `Historical...` model instances that contain personally identifying info, as those should be deleteable for GDPR purposes. One option might be to split models with PII apart: a FacilityClaim with PII, for instance, might have a FacilityClaimVerification or similar that does not have PII?

PR #521, opened 2019-05-16
Add django-simple-history

kellyi:
## Overview

- add django-simple-history library
- create HistoricalFacility audit model
- display history log for Facility on Django admin

Connects #519 

## Demo

Screen Shot 2019-05-16 at 11 02 05 AM
Screen Shot 2019-05-16 at 11 02 54 AM
Screen Shot 2019-05-16 at 11 02 30 AM

## Testing Instructions

On this branch
- `./scripts/update`
- `./scripts/manage resetdb`
- `./scripts/manage processfixtures`
- `./scripts/manage createsuperuser`
- `./scripts/server`
- sign into the django admin
- visit a Facility object and verify that you can see its history
- update a facility, then verify that you can see and revert the change
- use `./scripts/manage shell_plus` to interact with a HistoricalFacility object

## Checklist

- [ ] CI passes after rebase


jwalgran:
Looking at this now

jwalgran:
The docs mention the included management command that will create the initial history for existing records.
https://django-simple-history.readthedocs.io/en/2.7.2/quick_start.html#existing-projects

We should consider creating a 2.4.0 release issue now and adding the running of this management command as a step after running the migrations. What do you think?

kellyi:
> We should consider creating a 2.4.0 release issue now and adding the running of this management command as a step after running the migrations. What do you think?

Sounds good to me -- it'll help make sure we don't lose track of the migration related work.

kellyi:
I made https://github.com/open-apparel-registry/open-apparel-registry/issues/522 for the release and noted the extra step in a comment there.

kellyi:
Cool, thanks! I'll run the migration via ecsmanage after merging this. We'll still need to run the command to populate it afterwards.

Issue #523, opened 2019-05-16
Using the back button after filtering a facility list does not refresh the list / Filters urls should be replaces rather than pushed


jwalgran:
## Overview

Using the back button after filtering a facility list does not refresh the list.

### Expected Behavior

Changes to the list filter replace the current URL rather than pushing onto the history stack.

### Actual Behavior

Each filter change adds to the page history and the displayed items get out of sync with the URL.

### Steps to Reproduce

* Browse a list
* Filter by status
* Use the back button.

PR #524, opened 2019-05-16
Replace facility list detail urls when changing pages or filters

jwalgran:
## Overview

Each change to the page number of filter was pushing an item onto the navigation
history. This had some drawbacks:

- It is redundant, because the list items page has its own page navigation.
- The using the back button would change the url but not the content of the
page.

Rather than fix the back button we simply treat paging and filtering more like
local state by using `replace` instead of `push`.

Connects #523 

## Testing Instructions

* Log in as a user with an uploaded and processed list. If you have freshly processed fixture data, `c8@example.com` is a good choice.
* Browse a list details page, move through 2 pages of results, then verify that the browser back button returns you to the previous application page rather than staying on the list items page.
* Repeat the test but apply different filters instead of changing pages.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/components/FacilityListItemsTable.jsx
+++ b/src/app/src/components/FacilityListItemsTable.jsx
@@ -124,7 +124,7 @@ class FacilityListItemsTable extends Component {
                 },
             },
             history: {
-                push,
+                replace,
                 location: {
                     search,
                 },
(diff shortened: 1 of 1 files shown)


jwalgran:
Thanks for the review.

It looks like we already do this on the facility search. If I search for "Aldi [Public List]" then go to the "Search" tab, add "China" as an search filter, then use the Back button in the browser, I go back to the previous page rather than the Aldi-only search.

Issue #510, opened 2019-05-14
The country Macedonia has been re-named


katieashaw:
## Overview

Macedonia has been renamed, so we need to update the primary name associated with the country code.

### Expected Behavior

Facilities listed in "North Macedonia" are currently being rejected from the system. 

### Actual Behavior

Facilities listed as "North Macedonia" or "Republic of North Macedonia" should be accepted into the tool. 

### Steps to Reproduce

* Attempt to upload a facility with the country listed as "North Macedonia"

### Additional context

https://en.wikipedia.org/wiki/North_Macedonia
https://en.wikipedia.org/wiki/ISO_3166-2:MK

PR #525, opened 2019-05-17
Change the official name of Macedonia

jwalgran:
## Overview

In March of 2019 the country changed its name from The Former Yugoslav Republic
of Macedonia to North Macedonia. This commit updates our country code parsing
and model choice fields to match.

https://en.wikipedia.org/wiki/ISO_3166-2:MK
Connects #510 

## Demo

Screen Shot 2019-05-17 at 8 08 45 AM

## Testing Instructions

* Run `./scripts/update` to apply the migration
* Log in and upload
[macedonia.csv.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/3192078/macedonia.csv.txt)
* Run `./scripts/manage batch_process --list-id {the-id} --action parse` and verify that all three rows are correctly parsed as North Macedonia
* Verify that the country drop down on the main page contains North Macedonia

## Checklist

- [ ] CI passes after rebase


kellyi:
Do we need to migrate instances with the old country name & code too?

jwalgran:
We only every store the country code, MK, so no migration is required.

jwalgran:
Thanks for the review.

Issue #526, opened 2019-05-17
Attempting to browse a facility in the Django admin causes a service failure


jwalgran:
## Overview

On staging, attempting to browse a facility in the Django admin causes the ECS task to fail.

<img width="1350" alt="Screen Shot 2019-05-17 at 10 33 38 AM" src="https://user-images.githubusercontent.com/17363/57946084-59479d00-7890-11e9-892e-afd1e96cf1c6.png">

### Expected Behavior

An administrator can browse a facility record via the Django admin on staging.

### Actual Behavior

The application process fails and staging is offline for a minute or two while a replacement task is started. 

### Steps to Reproduce

* Log in as an administrative user
* Attempt to load a facility detail page in the Django admin 

### Additional context

Some potentially relevant log activity around the time of a failure.

### In RDS 

```
2019-05-16 17:55:03
UTC:10.0.1.210(50846):openapparelregistry:[23506]:LOG:
duration: 2115.492 ms statement: DECLARE "_django_curs_139786784380672_1" NO
SCROLL CURSOR WITH HOLD FOR SELECT "api_facilitylistitem"."id",
"api_facilitylistitem"."facility_list_id", "api_facilitylistitem"."row_index",
"api_facilitylistitem"."raw_data", "api_facilitylistitem"."status",
"api_facilitylistitem"."processing_started_at",
"api_facilitylistitem"."processing_completed_at",
"api_facilitylistitem"."processing_results", "api_facilitylistitem"."name",
"api_facilitylistitem"."address", "api_facilitylistitem"."country_code",
"api_facilitylistitem"."geocoded_point"::bytea,
"api_facilitylistitem"."geocoded_address", "api_facilitylistitem"."facility_id",
"api_facilitylistitem"."created_at", "api_facilitylistitem"."updated_at" FROM
"api_facilitylistitem"
```

Several of these `FETCH` statements appear
```
2019-05-16 17:55:18
UTC:10.0.1.210(50846):openapparelregistry:[23506]:STATEMENT:
FETCH FORWARD 2000 FROM "_django_curs_139786784380672_1"
```

### Then In ECS

```
ECS
[2019-05-16 17:56:01 +0000] [1] [CRITICAL] WORKER TIMEOUT (pid:9)
```

### Then Back In RDS

```
2019-05-16 17:57:25 UTC:10.0.1.210(50898):openapparelregistry:[24431]:LOG: could not receive data from client: Connection reset by peer
```

PR #527, opened 2019-05-17
Use some read only fields in the Django admin

jwalgran:
## Overview

The Facility, FacilityListItem, and FacilityMatch admin were loading combo boxes with all possible related objects. With a database populated with 10s of thousands of records this would make admin page requests so slow that the ECS task would be terminated.

This commit fixes that by simply making the related fields read only. This works because we would not want to adjust these specific foreign key relationships anyway.

While making some fields read only we also took the opportunity to also make the JSON status fields read only and add a helper to render them in a more readable way.

Connects #526

### Facility

Screen Shot 2019-05-17 at 1 18 25 PM

### FacilityListItem

Screen Shot 2019-05-17 at 1 17 59 PM

### FacilityMatch

Screen Shot 2019-05-17 at 1 18 40 PM

## Notes

## Testing Instructions

This `test/` branch was tested on staging which has a stale copy of production data.

* Log in as an admin and verify that you can successfully load Facility, FacilityListItem, and FacilityMatch detail pages.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Made some fields read only in the Django admin to prevent slow page loads resulting in service interruptions. [#527](https://github.com/open-apparel-registry/open-apparel-registry/pull/527)
+
 ### Security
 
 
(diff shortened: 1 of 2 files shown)


jwalgran:
Thanks for the review.

Issue #514, opened 2019-05-15
Evaluate django-waffle as a library for doing feature flags


kellyi:
## Overview

We need a feature flag solution order to move toward doing continuous delivery and to enable building the claim a facility feature while keeping it behind a feature flag. For this issue, evaluate https://github.com/django-waffle/django-waffle as a possible library for adding feature flags -- and, potentially, add it to the project if it is going to be viable.

### Describe the solution you'd like

- evaluate https://github.com/django-waffle/django-waffle
- if it is viable, add it to the project

### Additional Context

In addition to feature flags on the backend, we also need to have feature flags work in the React app. We already have a mechanism for sending environment data from Django to the React app, so we'll need to make sure that django-waffle can be used with that to propagate the data to the frontend.

There is also https://github.com/garbles/flag specifically for React/Redux, but I don't know whether we need it or not.

kellyi:
This will also need to apply on the frontend -- we should send back the feature flag configuration back to the frontend via the environment endpoint, and then have then operative in the React app.

PR #531, opened 2019-05-21
Enable feature flags in Django & React using django-waffle 

kellyi:
## Overview

- add & configure django-waffle library
- create api-feature-flags API endpoint
- add feature flags Redux actions & reducers to store currently active feature flags in state
- create FeatureFlag component to accept a flag and check whether it is active in the feature flags state, rendering only if so
- add temporary "Claim a facility" text on FacilityDetailsSidebar component to activate only when the claim_a_facility feature is active
- set some default props in the NavbarDropdown component to fix two prop types warnings

Connects #514

## Demo

With `claim_a_facility` switch active:

Screen Shot 2019-05-21 at 9 47 25 PM

->

Screen Shot 2019-05-21 at 9 44 16 PM

With `claim_a_facility` switch inactive:

Screen Shot 2019-05-21 at 9 44 47 PM

->

Screen Shot 2019-05-21 at 9 45 04 PM

With no `claim_a_facility` switch:

Screen Shot 2019-05-21 at 9 45 21 PM

->

Screen Shot 2019-05-21 at 9 45 32 PM

## Notes

We discussed potentially using this library https://github.com/garbles/flag but after investigating a bit further it seems like the functionality we needed from it was fairly straightforward to implement on our own. Since the library's still pre-official-release it seemed preferable to use our own implementation.

## Testing Instructions

- get this branch then `./scripts/update`
- `./scripts/manage resetdb`, `./scripts/manage processfixtures` and `./scripts/manage createsuperuser`
- sign in to the Django admin and verify that you now see entries for `Django-Waffle`
- without setting a switch, visit the React app and then look at the details page for a facility. Verify that you do not see the "Claim this facility." text at the bottom of the sidebar
- back in the Django admin, add a `claim_a_facility` switch but do not yet activate it, then save the new object
- return to the React facility details page and refresh, verifying that you still don't see the "Claim this facility." text
- back again in the Django admin, activate the `claim_a_facility` switch and then save the update.
- return again to the React facility details page and refresh, verifying that when the page fully loads you do see the "Claim this facility." text

## Checklist

- [x] CI passes after rebase


src/django/web/views.py
```diff
@@ -11,7 +11,6 @@
 # Add Environment variable, lowered for easier comparison
 # One of `development`, `testing`, `staging`, `production`
 environment['ENVIRONMENT'] = settings.ENVIRONMENT.lower()
-
```
kellyi: committed this change by mistake

jwalgran:
Looking at this now

src/django/api/views.py
```diff
@@ -1219,3 +1220,13 @@ def reject_match(self, request, pk=None):
             raise NotFound()
         except FacilityMatch.DoesNotExist:
             raise NotFound()
+
+
+@api_view(['GET'])
+@permission_classes((AllowAny,))
+def api_feature_flags(request):
+    response_data = {
+        'claim_a_facility': switch_is_active('claim_a_facility'),
```
jwalgran: We should consider codifying the deployment-time task of creating this flag in the database by adding a data migration that creates the inactive flag record.

src/app/src/components/NavbarDropdown.jsx
```diff
@@ -75,7 +75,7 @@ class NavbarDropdown extends Component {
                     aria-haspopup="true"
                     onClick={this.handleToggle}
                 >
-                    {title}
+                    {title || 'User'}
```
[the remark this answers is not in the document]
kellyi: Good point. I split this into 
5c8ea98

src/app/src/components/FeatureFlag.jsx
```diff
@@ -0,0 +1,40 @@
...
+        <React.Fragment>
+            {children}
+        </React.Fragment>
+    );
+}
+
+FeatureFlag.propTypes = {
+    flag: featureFlagPropType.isRequired,
+    children: node.isRequired,
+    featureFlags: arrayOf(featureFlagPropType).isRequired,
```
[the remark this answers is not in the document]
kellyi: That makes sense -- I am going to rename the prop to `activeFeatureFlags`

src/django/api/views.py
kellyi: Sounds good -- I created the migration here:

8303bd8a93bea2f2161c569134e46104fb566e83

kellyi:
Thanks for your help with this! Going to merge then will run the migrations on staging.

Issue #515, opened 2019-05-15
Add mimimal, feature-flagged claim a facility workflow


kellyi:
## Overview

Add a minimal claim-a-facility workflow which will sit behind a feature flag.

### Describe the solution you'd like

- add a "Claim this facility" `Link` to the facility details page which is visible only to users who are signed in and have contributor types of facility owners
- clicking the link should move to a new route -- `/facilities/OARID/claim` or similar -- which would display a form with a currently minimal set of fields to submit
- add a new `/api/facilities/OARID/claim` endpoint or similar which accepts a posted form and creates a new FacilityClaim. We'll also need to create a FacilityClaim model as part of this work
- when a facility claim is pending for a specific user, that user should not be able to make another claim on the facility until it has been accepted of rejected
- the FacilityClaim model and the notes models should be opted into the the audit library (likely django-simple-history)

### Additional Context

We should chat with @designmatty about what the wireframes for the non-MVP version of this form will look like. For now the UI will just be a placeholder.

PR #540, opened 2019-05-28
Enable submitting claim a facility form

kellyi:
## Overview

This PR enables users to complete and submit a facility claim form via a new route in the UI. That route currently lives behind a feature flag -- and we render the map as a fallback if the correct flag is not active.

The workflow is:

- user clicks on a link or a badge in the facility details page
- user goes to a `facilities/id/claim` route
- if the user is not signed in, we show a link to the login page
- if the user is signed in, we render the wizard from from the wireframes
- the form validates the email address and company name as required fields. Everything else is optional
- submitting the form creates a new `FacilityClaim` object in the API and sends an email to the user to confirm that the claim has been submitted.

There is some remaining work here but I wanted to get some feedback on it now. The remaining work I see here is:

- prevent users who have already submitted a claim on a facility from re-submitting the form for that facility (this is now #555)
- ~render error messages from the API in the form wizard~
- ~add some kind of "end state" to the form wizard to indicate that the user has successfully submitted the form~
- ~clean up the form style~

Connects #515
Connects #544 

## Demo

Screen Shot 2019-06-03 at 3 53 13 PM

Screen Shot 2019-06-03 at 3 53 30 PM

Screen Shot 2019-06-03 at 3 53 36 PM

Screen Shot 2019-06-03 at 3 53 43 PM

Screen Shot 2019-06-03 at 3 53 58 PM

Screen Shot 2019-06-03 at 3 54 03 PM

## Testing Instructions

- get this branch then `./scripts/update`, `./scripts/manage resetdb`, `./scripts/manage createsuperuser` and `./scripts/server`
- visit localhost:6543 and search for a facility. from the details page, add the `/claim` path after the OAR ID and verify that it does not render the claim form (since the feature flag is still off)
- sign in to the django-admin and turn the feature flag on
- refresh the claim path on localhost:6543 and verify that you now see the claim form, since you are signed in
- sign out, then sign in as an account with a contributor attached (c8@example) then return to the claim form
- complete the form and submit it, verifying that the submission completes, that you reach the confirmation step of the wizard, and that you see the confirmation email sent in the terminal

## Checklist

- [x] CI passes after rebase


src/app/src/components/ClaimFacilityStepper.jsx
```diff
@@ -0,0 +1,97 @@
...
+    Object.freeze({
+        name: 'Finish Up',
+    }),
+]);
+
+export default function ClaimFacilityStepper() {
+    const [activeStep, setActiveStep] = useState(0);
+
+    const incrementActiveStep = () => setActiveStep(activeStep + 1);
+    const decrementActiveStep = () => setActiveStep(activeStep - 1);
```
kellyi: Reminder to me to clamp these values between 0 & `step.length`: https://lodash.com/docs/4.17.11#clamp

kellyi:
@jwalgran I believe this is 95% done but I wanted to request a review now to get some feedback on what's currently here.

src/django/api/views.py
```diff
@@ -1211,3 +1214,52 @@ def api_feature_flags(request):
     }
 
     return Response(response_data)
+
+
+@transaction.atomic
+@api_view(['POST'])
+@permission_classes((IsRegisteredAndConfirmed,))
+def claim_facility(request, oar_id=None):
```
kellyi: I made an api_view method for this because I wanted to use the permission classes decorator, which wasn't possible with the class method. You should also try to CURL at this endpoint and verify that it returns an authorization error.

src/app/src/App.jsx
```diff
@@ -78,6 +82,18 @@ class App extends Component {
...
                                 />
+                                <Route
+                                    exact
+                                    path={claimFacilityRoute}
+                                    render={() => (
+                                        <FeatureFlag
+                                            flag={CLAIM_A_FACILITY}
+                                            alternative={<Route component={MapAndSidebar} />}
+                                        >
+                                            <Route component={ClaimFacility} />
```
kellyi: The wonky `Route` nesting here is because I wanted to continue getting the OAR ID from the path.

src/app/src/components/ClaimFacilityStepper.jsx
```diff
@@ -0,0 +1,226 @@
...
+    email,
+    companyName,
+    contributorEmail,
+    setEmailFromContributorEmail,
+}) {
+    const [activeStep, setActiveStep] = useState(0);
+
+    useEffect(() => {
+        setEmailFromContributorEmail(contributorEmail);
+    }, [contributorEmail, setEmailFromContributorEmail]);
```
kellyi: This will only fire on mount and when `contributorEmail` and `setEmailFromContributorEmail` props change -- which will not happen while the component is mounted, so it will only fire once.

There's a hooks linting rule to require that one declare all dependencies of `useEffect` in the array, which is why I've added them here.

src/django/api/views.py
jwalgran: On a hunch I tried passing `permission_classes` to the `@action` decorator on the facilities `count` method and it appears to work correctly.

```diff
diff --git a/src/django/api/views.py b/src/django/api/views.py
index ac3afde..a058ba9 100644
--- a/src/django/api/views.py
+++ b/src/django/api/views.py
@@ -533,7 +533,8 @@ class FacilitiesViewSet(ReadOnlyModelViewSet):
         except Facility.DoesNotExist:
             raise NotFound()
 
-    @action(detail=False, methods=['get'])
+    @action(detail=False, methods=['get'],
+            permission_classes=(IsRegisteredAndConfirmed,))
     def count(self, request):
         """
         Returns a count of total Facilities available in the Open Apparel

```

I think we can use this to nest an authenticated `claim` action under the otherwise open `FacilitiesViewSet`.

src/django/api/views.py
```diff
@@ -1211,3 +1214,52 @@ def api_feature_flags(request):
...
+    if not switch_is_active('claim_a_facility'):
+        return Response("Bad request", status.HTTP_400_BAD_REQUEST)
```
jwalgran: Consider returning 404 here. It doesn't really matter, but I imagine a feature flag more like making it appear that a feature does not exist.

src/django/api/views.py
kellyi: Great! I'll update this and drop the newly added URL pattern.

src/django/api/models.py
```diff
@@ -342,6 +342,92 @@ def __str__(self):
...
+        help_text='The contributor who submitted this facility claim')
+    facility = models.ForeignKey(
+        'Facility',
+        null=False,
+        on_delete=models.PROTECT,
+        help_text='The facility for which this claim has been submitted')
+    contact_person = models.CharField(
+        max_length=200,
+        null=False,
+        blank=True,
```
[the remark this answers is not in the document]
kellyi: Good point -- I was mostly just going by what fields had been required or not in the wireframes but I think this should be required. I'll update this and re-assess whether other fields should also be required.

src/django/api/models.py
```diff
@@ -342,6 +342,92 @@ def __str__(self):
...
+        help_text='A description of the facility')
+    verification_method = models.TextField(
+        null=False,
+        blank=True,
+        help_text='An explanation of how the facility can be verified')
+    preferred_contact_method = models.CharField(
+        max_length=200,
+        null=True,
```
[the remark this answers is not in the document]
kellyi: Good catch -- this was originally due to not having the preferred contact select be required in the client which means that it was possible to send `null` in as the value here. However, I think it should be an empty string on the API side so I'll default it to that & also re-assess whether the option should be required.

kellyi:
@jwalgran this is ready for testing. I made some modifications to the form wizard to depart from the API: the end state is now a confirmation page, and I split the verification text list into a separate step from adding verification info.

kellyi:
I also spun off the "prevent users from re-submitting facility claims" to #555 because it requires some thought about how to implement it.

jwalgran:
Looking at this now.

jwalgran:
The current implementation allows you to progress even if you have left a step incomplete. Is this by design? Since we have client-side validation, we could check the validity of each page individually.

kellyi:
> The current implementation allows you to progress even if you have left a step incomplete. Is this by design? Since we have client-side validation, we could check the validity of each page individually.

This was by design but I can adjust it to prevent moving forward past invalid fields if we would like.

jwalgran:
We can keep the validation workflow as is but we may want to consider mentioning the using the back button to look at previous pages in the error message. There may be some confusion if the last page is completely filled out but we tell them that required fields are missing.

kellyi:
> We can keep the validation workflow as is but we may want to consider mentioning the using the back button to look at previous pages in the error message.

I think I'll just disable the "next" button if a required field is missing. That'd limit the error messages on the penultimate step to those issued from the server-side.

src/app/src/components/ClaimFacilityConfirmationStep.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import Typography from '@material-ui/core/Typography';
+
+import { claimAFacilityFormStyles } from '../util/styles';
+
+export default function ClaimFacilityConfirmationStep() {
+    return (
+        <div style={claimAFacilityFormStyles.inputGroupStyles}>
```
jwalgran: Small style issue. Text sometimes runs up against the right side of the wizard panel.

Screen Shot 2019-06-05 at 11 14 11 AM

jwalgran:
I'm getting a 404 when I submit my new claim, but the wizard is moving to a success page.

Stopped server, ran `./scripts/update` and restarted the server just to be sure. I can repeat this.

Screen Shot 2019-06-05 at 11 27 24 AM

jwalgran:
Ah. This was my fault. I was logged in as a superuser without an associated contributor.

I think it is still a bug, however, that non 2XX responses still move on to a success page.

kellyi:
> I think it is still a bug, however, that non 2XX responses still move on to a success page.

Yeah -- I'll update the handler for that to check that there is not also currently an error when the fetching stops.

kellyi:
2a3c473f7e5b4acd94604473f0d289d85c191a7a fixes the issue whereby the stepper would advance to the confirmation step regardless of the outcome of the API request. I had set up the `useEffect` code incorrectly, so now it's working as expected.

src/app/src/components/ClaimFacilityConfirmationStep.jsx
kellyi: Fixed this in ee5499c9b4c74eb7c761a672acdab065c021485c
by giving the message's element a width of 95%.

kellyi: Screen Shot 2019-06-05 at 6 29 41 PM

Screen Shot 2019-06-05 at 6 30 03 PM

kellyi:
a76d1117475977d8b29dd27cf88707bc9c2ddaa6 handles validating the form at each step to prevent users from advancing past invalid fields.

jwalgran:
Taking a look now.

kellyi:
Pushed 46f3e85ff40dbfa3edbf5ce9b6162a9abd343daf following a discussion with @designmatty and checking out the new wireframes:

- flip the boxes on the contact method/submit step per the wireframes:

Screen Shot 2019-06-06 at 11 26 36 AM

- move the verification list to the confirmation step, per discussion:

Screen Shot 2019-06-06 at 11 26 06 AM

kellyi:
Thanks for your help with this! I updated the text on the confirmation page here:

https://github.com/open-apparel-registry/open-apparel-registry/pull/540/commits/beae14f5903bd56c5f4a31b2c4c1a4ef73d9f931

```
                Your request to claim the facility was submitted successfully!
                To manage your facility, we&#39;ll need to verify your
                connection with the facility. Once verified you will be able
                to:
```

Issue #538, opened 2019-05-28
Add client code to perform IP geolocation and write the country to the app state


jwalgran:
## Overview

Add a component that makes a request to a web service which returns client information based on IP address. Write the detected country to the app state.

### Is your feature request related to a problem? Please describe.

We need to detect when users are browsing from China and update our client-side map resource loading accordingly.

### Describe the solution you'd like

- From the client, make a request to https://api.userinfo.io/userinfos
- If available, write the value of `country.code` in the response to the Redux store. 
- This should only happen once, on page load

PR #541, opened 2019-05-28
Fetch client country code and write to app state

jwalgran:
## Overview

We need to detect if someone is browsing the app from China so that we can
change the way in which we load our map components. To facilitate this we make a
request to the free userinfo.io service when the `App` component loads and save
the returned country to the app state.

Connects #538 	

### The userinfo.io service returning the correct country with the China VPN connection.

Screen Shot 2019-05-28 at 2 51 26 PM

### Writing country to app state

Screen Shot 2019-05-28 at 2 54 21 PM

### Setting fetched flag on failure

Screen Shot 2019-05-28 at 2 56 54 PM

## Testing Instructions

* Open developer tools and browse http://localhost:6543/. Verify that `clientInfo.countryCode` is set in the app state.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/reducers/ClientInfoReducer.js
+++ b/src/app/src/reducers/ClientInfoReducer.js
@@ -1,0 +1,21 @@
+import { createReducer } from 'redux-act';
+import update from 'immutability-helper';
+
+import {
+    startFetchClientInfo,
+    failFetchClientInfo,
+    completeFetchClientInfo,
+} from '../actions/clientInfo';
+
+const initialState = Object.freeze({ fetched: false, countryCode: null });
+
+export default createReducer({
+    [startFetchClientInfo]: () => initialState,
+    [failFetchClientInfo]: state => update(state, {
+        fetched: { $set: true },
+    }),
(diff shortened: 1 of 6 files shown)


src/app/src/reducers/ClientInfoReducer.js
```diff
@@ -0,0 +1,21 @@
...
+
+export default createReducer({
+    [startFetchClientInfo]: () => initialState,
+    [failFetchClientInfo]: (state, payload) => update(state, {
+        fetched: { $set: true }
+    }),
+    [completeFetchClientInfo]: (state, payload) => update(state, {
+        fetched: { $set: true },
+        countryCode: { $set: payload.country.code }
+    }),
```
kellyi: Would there ever be a need to clear this during the running of the app? It seems like the answer is no but I thought I'd check.

jwalgran: I don't think we would ever need to clear the fetched client info.

jwalgran:
Thanks for the review

Issue #503, opened 2019-05-02
Allow free text search of facility list items


jwalgran:
## Overview

On the facility list details page, allow free text searching of names and addresses.

### Describe the solution you'd like

* Add a `search` parameter to the `/api/facility-lists/{id}/items/` API endpoint
  * Filter the results based on a basic case-insensitive substring match against the name and address
* Add a search text box to the facility list page. Pass its value to the API as the `search` parameter 
* Add a reset / show all button that clears the search.

rajadain:
Where should I add the search box? Beside the filter box (and the related controls) (red) or below it (blue)?

![image](https://user-images.githubusercontent.com/1430060/58499061-5892ee80-814d-11e9-99f5-f35125664011.png)

jwalgran:
My first instinct is to put it before the status filter dropdown (orange arrow)

![search](https://user-images.githubusercontent.com/17363/58501061-5bcfaf80-8138-11e9-8ca8-ca86e42a4ec1.png)

The label and button to the right of the status filter dropdown disappear. The left edge is a less volatile location

Putting it on a new line is also less volatile but it may look a little unbalanced and it pushes the start of the actual content even further down the page.

PR #542, opened 2019-05-29
Free Text Search for Facility List Items

rajadain:
## Overview

Adds a free text search control to the facility list items table that allows filtering values with arbitrary text. Text search works in conjunction with the status filter, operating as an "AND" to the status filter. The search text itself is taken as a single phrase, and we look for an exact case-insensitive occurrence in the name or address for matches.

The control supports both mouse and keyboard events. Users can search by typing in the value and pressing "Enter" to search, or "Escape" to reset it. When entering a search term, there is a 🔍  button that can be pressed to start the search. Afterwards, there is a ✖️ button which can be used to reset the box.

Also upgrades Material UI to 3.1.0 which adds support for `outlined` and `filled` variants of controls such as `TextField`.

Connects #503 

## Notes

During the search there should be a spinner, but often times the search is so fast that the spinner doesn't render. There may also be an issue with the parent table re-rendering resetting the spinner state of the search control.

The initial implementation is using the `useState` hook. There is an alternative implementation in the final commit that uses `useReducer` instead. We can choose between them. I have a slight preference for `useState`.

## Testing Instructions

* Check out this branch, `update` and `server`
* Go to a list
  - [x] Ensure you see a search control to the left of the status filter
* Type in a search term. Press <kbd>Enter</kbd>
  - [x] Ensure the list is updated to reflect the search
* ~~Focus on the search box. Press <kbd>Escape</kbd>~~
  - ~~Ensure the search is reset~~
* Try the above with the mouse
  - [x] Ensure using the 🔍 button triggers a search
  - [x] Ensure using the ✖️ button resets a search
* Try in combination with status filters
  - [x] Ensure the search and status filters work together
* Try refreshing the page with an active search
  - [x] Ensure the new page loads the correctly filtered items

### DIFF
--- a/src/app/src/components/FacilityListItemsFilterSearch.jsx
+++ b/src/app/src/components/FacilityListItemsFilterSearch.jsx
@@ -1,6 +1,8 @@
 import React, { useEffect, useState } from 'react';
 import { func, string } from 'prop-types';
 
+import trim from 'lodash/trim';
+
 import IconButton from '@material-ui/core/IconButton';
 import InputAdornment from '@material-ui/core/InputAdornment';
 import TextField from '@material-ui/core/TextField';
@@ -24,7 +26,9 @@ export default function FacilityListItemsTableSearch({
             return;
         }
 
-        onSearch(searchText);
+        const term = trim(searchText);
+        onSearch(term);
+        setSearchText(term);
     };
 
     const handleTextChange = (e) => {
--- a/src/app/src/components/FacilityListItemsFilterSearch.jsx
+++ b/src/app/src/components/FacilityListItemsFilterSearch.jsx
@@ -67,7 +67,6 @@ export default function FacilityListItemsTableSearch({
 
     return (
         <TextField
-            margin="dense"
             variant="outlined"
             type="text"
             label="Search"
--- a/src/app/src/components/FacilityListItemsFilterSearch.jsx
+++ b/src/app/src/components/FacilityListItemsFilterSearch.jsx
@@ -1,0 +1,95 @@
+import React, { useEffect, useState } from 'react';
+import { func, string } from 'prop-types';
+
+import IconButton from '@material-ui/core/IconButton';
+import InputAdornment from '@material-ui/core/InputAdornment';
+import TextField from '@material-ui/core/TextField';
+import Search from '@material-ui/icons/Search';
+import Close from '@material-ui/icons/Close';
+
+import { ENTER_KEY } from '../util/constants';
+
+export default function FacilityListItemsTableSearch({
+    currentValue,
+    onSearch,
+}) {
+    const [searchText, setSearchText] = useState(currentValue);
+    const isCurrent = searchText === currentValue;
+
+    // Update internal state if the prop is updated
+    useEffect(() => setSearchText(currentValue), [currentValue]);
+
+    const handleSearch = () => {
+        if (isCurrent) {
+            return;
+        }
+
+        onSearch(searchText);
+    };
+
+    const handleTextChange = (e) => {
+        setSearchText(e.target.value);
+    };
+
+    const handleReset = () => {
+        setSearchText('');
+        onSearch('');
+    };
+
+    const handleTextKeyDown = (e) => {
+        if (e.key === ENTER_KEY) {
+            handleSearch();
+        }
+    };
+
+    const searchButton = (
+        <IconButton
+            edge="end"
+            aria-label="Search"
+            onClick={handleSearch}
+        >
+            <Search />
+        </IconButton>
+    );
+    const resetButton = (
+        <IconButton
+            edge="end"
+            aria-label="Reset"
+            onClick={handleReset}
+        >
+            <Close />
+        </IconButton>
+    );
+
+    const adornment = (searchText !== '' && isCurrent)
+        ? resetButton
+        : searchButton;
+
+    return (
+        <TextField
+            margin="dense"
+            variant="outlined"
+            type="text"
+            label="Search"
+            onChange={handleTextChange}
+            onKeyDown={handleTextKeyDown}
+            value={searchText}
+            InputProps={{
+                endAdornment: (
+                    <InputAdornment position="end">
+                        {adornment}
+                    </InputAdornment>
+                ),
+            }}
+        />
+    );
+}
+
+FacilityListItemsTableSearch.propTypes = {
+    currentValue: string,
+    onSearch: func.isRequired,
+};
+
+FacilityListItemsTableSearch.defaultProps = {
+    currentValue: '',
+};
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add django-waffle and configure Django & React apps to enable feature flags [#531](https://github.com/open-apparel-registry/open-apparel-registry/pull/531)
 - Support uploading Excel files [#532](https://github.com/open-apparel-registry/open-apparel-registry/pull/532)
 - Fetch client country code based on IP [#541](https://github.com/open-apparel-registry/open-apparel-registry/pull/541)
+- Free text search filter for facility list items [#542](https://github.com/open-apparel-registry/open-apparel-registry/pull/542)
 - Add admin-authorized dashboard route: [#553](https://github.com/open-apparel-registry/open-apparel-registry/pull/553)
 
 ### Changed
--- a/src/app/package.json
+++ b/src/app/package.json
@@ -5,7 +5,7 @@
     "private": true,
     "dependencies": {
         "@google/markerclusterer": "1.0.3",
-        "@material-ui/core": "3.0.1",
+        "@material-ui/core": "3.1.0",
         "@material-ui/icons": "3.0.1",
         "@turf/turf": "5.1.6",
         "@turf/distance": "6.0.1",
--- a/src/app/src/components/FacilityListItemsTable.jsx
+++ b/src/app/src/components/FacilityListItemsTable.jsx
@@ -58,15 +58,15 @@ const facilityListItemsTableStyles = Object.freeze({
         lineHeight: '1.2',
     }),
     searchFilterStyles: Object.freeze({
-        display: 'inline-block',
         marginRight: '5px',
     }),
     statusFilterStyles: Object.freeze({
-        padding: '20px 0 0 20px',
+        padding: '10px 20px 0 20px',
+        display: 'flex',
+        alignItems: 'center',
     }),
     statusFilterSelectStyles: Object.freeze({
-        display: 'inline-block',
-        width: '40%',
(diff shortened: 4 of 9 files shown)


src/app/src/components/FacilityListItemsTableSearch.jsx
```diff
@@ -0,0 +1,153 @@
...
+import TextField from '@material-ui/core/TextField';
+import Search from '@material-ui/icons/Search';
+import Close from '@material-ui/icons/Close';
+
+const KEYCODE_ENTER = 13;
+const KEYCODE_ESCAPE = 27;
+
+function reducer(state, action) {
+    switch (action.type) {
+        case 'update':
```
kellyi: Consider setting these cases as constants.

kellyi:
Reassigning this to @jwalgran because I may not have time to look at it before RRP.

jwalgran:
Looking at this now

jwalgran:
Searching whitespace triggers goes to an "Unable to retrieve that list" view. We should consider simply trimming the contents of the search box.

Screen Shot 2019-05-30 at 9 59 16 AM

jwalgran:
The behavior of the "Show All Items" button when filtering by both text and status is a bit intuitive. The word "All" leads me to believe that it would clear both the status filter and the text search.

src/app/src/components/FacilityListItemsTableSearch.jsx
```diff
@@ -0,0 +1,123 @@
...
+        if (searchText !== '' && isCurrent) {
+            return resetButton;
+        }
+
+        return searchButton;
+    })();
+
+    return (
+        <TextField
+            id="outlined-adornment-password"
```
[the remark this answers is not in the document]
rajadain: Sorry, copypasta from Material UI docs 🤦‍♂

src/app/src/components/FacilityListItemsTableSearch.jsx
rajadain: Switched back to `useState`. Thought this decoupled the logic from the component too much in this case. I'm sure there's a better use case for it coming in the future.

src/app/src/components/FacilityListItemsTableSearch.jsx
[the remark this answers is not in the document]
rajadain: Removed in 4e5432f

src/app/src/components/FacilityListItemsTableSearch.jsx
```diff
@@ -0,0 +1,122 @@
...
+const KEYCODE_ENTER = 13;
+const KEYCODE_ESCAPE = 27;
+
+export default function FacilityListItemsTableSearch({
+    currentSearchTerm,
+    onSearch,
+}) {
+    const [isSearching, setSearching] = useState(false);
+    const [isCurrent, setCurrent] = useState(true);
+    const [searchText, setSearchText] = useState(currentSearchTerm);
```
kellyi: Question: since this last `useState` is set from a prop, what happens if the prop's value changes? I assume the initial value is captured on mounting and doesn't automatically update thereafter but I'm not entirely sure.

I think this may be a good case for the useEffect hook: essentially if `currentSearchTerm` ever changes from above I think we want to ensure that `setSearchText(currentSearchTerm)` is called. I think it'd look like:

```
const [searchText, setSearchText] = useState('');

useEffect(() => {
    setSearchTerm(currentSearchTerm);
}, [currentSearchTerm, setSearchText]);
```

The quirk here generally seems to be that we're storing `currentSearchTerm` somewhere else (either Redux or the querystring param?) but then scoping the changes to the actual text input only here for some part of the lifecycle.

rajadain: My understanding was that if the prop value changes then the whole thing is re-rendered.

kellyi: This also may circumvent needing to have `isCurrent` around? It's not clear from the var name what `isCurrent` represents -- although I think it may be a bool indicating whether the search input state is the same as `currentSearchTerm`.

jwalgran:
Related to my comment about the "Show All Items" button, I think it would make sense to show the same "Showing N of M items" label and the "Show All Items" button when filtering by text. Would that complicate the implementation?

src/app/src/components/FacilityListItemsTableSearch.jsx
```diff
@@ -0,0 +1,122 @@
...
+    const handleReset = () => {
+        setSearchText('');
+        setCurrent(false);
+        doSearch('');
+    };
+
+    const handleTextKeyDown = (e) => {
+        if (e.keyCode === KEYCODE_ENTER) {
+            handleSearch();
+        } else if (e.keyCode === KEYCODE_ESCAPE) {
```
kellyi: We can probably keep these constants here but I noticed that the `constants` file already has a constant for the enter key to compare against `e.key`.

We also have this `makeSubmitFormOnEnterKeyPress` function -- https://github.com/open-apparel-registry/open-apparel-registry/blob/490d2866ffc7a72666299535dbc8611df2643715/src/app/src/util/util.js#L468 -- although it was designed to be used like this:

https://github.com/open-apparel-registry/open-apparel-registry/blob/312ccdbdb2659bafd952a180ee79d39abddac2e4/src/app/src/components/SendResetPasswordEmailForm.jsx#L146

rajadain:
Hmm. The interaction with the "Show All" button is a good point. I'm going to take a step back and see if the two filters can be combined under an umbrella component.

src/app/src/components/FacilityListItemsTableSearch.jsx
rajadain: Ah thanks for the tip! I'll try to reuse that.

src/app/src/components/FacilityListItemsTableSearch.jsx
kellyi: > My understanding was that if the prop value changes then the whole thing is re-rendered.

We should check this to make sure. I was thinking that `useState` would only be called once, on initially mounting the component -- otherwise the state wouldn't be durable across renders.

kellyi: Found this: https://stackoverflow.com/a/54866051

> The argument passed to useState is the initial state much like setting state in constructor for a class component and isn't used to update the state on re-render

rajadain: Yes I'm just finding out the same thing. Alright, will switch to another hook.

rajadain:
Just pushed up a rewrite of the status filter component in Material UI, since React Select was equally difficult to style (and difficult to figure out what styles were needed for the height / text baseline correspondence):



I think the decision to move away from React Select is potentially contentious, and we can talk about it. Figuring out why the styles were different, and which styles to update on each to make them correspond, and then how to get those styles wired in through the various custom components, was more work than rewriting in Material UI. We do lose out on the ability to easily ✖️ out a selected option, and lose a bit of the visibility of many options in the select textbox. But the dropdown has easily visible checkboxes, and checking / unchecking them updates the filters in real time, so it's not a total loss.

The final two things I'd like to do now are:

1. Make the search box clear when the "Show All" button is pressed
2. Maybe have a spinner somewhere when the table is loading

kellyi:
> We do lose out on the ability to easily ✖️ out a selected option, and lose a bit of the visibility of many options in the select textbox. But the dropdown has easily visible checkboxes, and checking / unchecking them updates the filters in real time, so it's not a total loss.

I think these sound like potentially big losses. It seems like it's difficult to know what options have been selected in that box now, since all the statuses are so long that they will overflow. Does the new textbox include auto-suggest completion?

src/app/src/components/FacilityListItemsFilterSearch.jsx
```diff
@@ -0,0 +1,98 @@
...
+    const handleSearch = () => {
+        if (isCurrent) {
+            return;
+        }
+
+        onSearch(searchText);
+    };
+
+    const handleTextChange = (e) => {
+        setSearchText(e.target.value);
```
kellyi: Forgot to mention this but there's a `getValueFromEvent` utility method for this. It's debatable whether this is more readable but this could potentially just be

```
const handleTextChange = flow(getValueFromEvent, setSearchText);
```

rajadain:
> Does the new textbox include auto-suggest completion?

You can't really type in it. Focusing on it opens the menu.

rajadain:
There is an option to render with `Chip`s:

image

but we'll probably go multi-line with it.

kellyi:
> You can't really type in it. Focusing on it opens the menu.

Could we create a custom input component for the search instead that has styling to match React Select? That would circumvent having to try to style the material UI component and make it so we just have to style an input element.

rajadain:
> Could we create a custom input component for the search instead that has styling to match React Select? 

Yes I can try that.

rajadain:
React Select does go multiline itself when given many options:

image

rajadain:
Alright, I've moved the rewrite to [`tt/free-text-search-2`](https://github.com/open-apparel-registry/open-apparel-registry/tree/tt/free-text-search-2). I took another pass at styling React Select, and had more success this time:



I think this covers most things requested so far. This is ready for another look.

jwalgran:
Thanks for taking the time to work through this. Looking at this now.

src/app/src/components/FacilityListItemsFilterSearch.jsx
```diff
@@ -0,0 +1,96 @@
...
+    const handleReset = () => {
+        setSearchText('');
+        onSearch('');
+    };
+
+    const handleTextKeyDown = (e) => {
+        if (e.keyCode === ENTER_KEYCODE) {
+            handleSearch();
+        } else if (e.keyCode === ESCAPE_KEYCODE) {
+            handleReset();
```
jwalgran: In Chrome and Firefox on my macOS 10.14 workstation pressing the <kbd>esc</kbd> just makes the field loose focus.





In Safari it does clear the text box, but:
- If the search had not been applied the page reloads when it is not nessesary
- This does not match the behavior of pressing <kbd>esc</kbd> while the status dropdown has focus (which is removing focus from that field)

Consider dropping any special <kbd>esc</kbd> handling.

src/app/src/components/FacilityListItemsFilterSearch.jsx
rajadain: I find the current version to be more readable, will keep that. Thanks for surfacing that though!

src/app/src/components/FacilityListItemsFilterSearch.jsx
rajadain: Understood. Dropped in 37c518b.

rajadain:
Removed whitespace in 8c20130. Ready for one final look.

jwalgran:
Looking at this now

rajadain:
Thanks for the thorough reviews! Squashed fixups and rebased on `develop`, also added a line to the `CHANGELOG`. Will wait until it's green before merging.

Issue #549, opened 2019-05-30
Enable hot reloading in the React app


kellyi:
## Overview

Enable hot reloading in the React app. Having this work would save quite a bit of development time overall as it'd supplant us having to fully refresh the page and recreate the app state to test out a code change.

### Is your feature request related to a problem? Please describe.

The development loop between making and seeing frontend changes feels pretty long, as it involves a full refresh each time. Having the app patch changes without refresh via hot reloading would save time.

### Describe the solution you'd like

- Enable hot reloading
- Do not eject from Create React App to accomplish this

### Additional Context

There seem to be some quirks involved in setting up hot reloading with Create React App. I think it is desirable *not* to eject from CRA, so we should figure out the correct workaround.

I'm pointing this at 1 because I think we should spend no more than 4 hours of time trying to accomplish this right now.

kellyi:
Setting this up using https://github.com/cdharris/react-app-rewire-hot-loader was extremely easy.

However, it appears that our current `proxy` setting to use the API is breaking the websocket connection used by webpack-dev-server, so I'm going to fix that.

PR #556, opened 2019-06-03
Configure hot reloading for React app during development

kellyi:
## Overview

This PR configures hot reloading for the React app during development. To accomplish this without ejecting from CRA, I brought in

- React App Rewired (https://github.com/timarney/react-app-rewired) which enables customizing CRA webpack configs without ejecting
- React App Rewire Hot Loader (https://github.com/cdharris/react-app-rewire-hot-loader) which enables setting up hot reloading with react-app-rewired

Setting this up also required adjusting our development proxy settings because the previous setup would route all requests not caught by React Router to Django -- including the websocket requests used by Webpack Dev Server to push hot updates. (This was the source, btw, of the frequent error messages about the socket disconnecting during development). To fix that I followed the guide here -- https://facebook.github.io/create-react-app/docs/proxying-api-requests-in-development#configuring-the-proxy-manually -- to set up a manual proxy.

Since not all of our API routes are prefixed with `/api/`, I created a list of the routes from urls.py and then looped over them to set up proxies. It seemed like this change would have a smaller impact than trying to update the Django routing here. (I think we should eventually update the Django routing, but it's out of scope for this issue.)

The result is that we now have hot reloading in development.

Connects #549

## Notes

Second commit here drops an unused dependency on react-mapgl that I noticed when messing with the package.json file.

## Testing Instructions

- get branch, then `./scripts/update`
- once update has finished, visit the app at http://localhost:8081 and verify that it works as it did before
- `./scripts/server` then visit the development app at http://localhost:6543
- make some changes to the `FilterSidebarSearchTab` component locally, then save. Verify that after saving the changes are rendered without refreshing the page and that the previous app state is maintained

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/package.json
+++ b/src/app/package.json
@@ -25,7 +25,6 @@
         "react-copy-to-clipboard": "5.0.1",
         "react-dom": "16.8.6",
         "react-hot-loader": "4.9.0",
-        "react-map-gl": "3.2.8",
         "react-redux": "5.1.0",
         "react-router": "4.3.1",
         "react-router-dom": "4.3.1",
--- a/src/app/package.json
+++ b/src/app/package.json
@@ -5,6 +5,7 @@
     "private": true,
     "dependencies": {
         "@google/markerclusterer": "1.0.3",
+        "@hot-loader/react-dom": "16.8.6",
         "@material-ui/core": "3.1.0",
         "@material-ui/icons": "3.0.1",
         "@turf/turf": "5.1.6",
@@ -13,13 +14,17 @@
         "core-js": "2.6.4",
         "file-saver": "2.0.0",
         "google-map-react": "1.1.2",
+        "http-proxy-middleware": "0.19.1",
         "immutability-helper": "2.9.0",
         "lodash": "4.17.11",
         "moment": "2.24.0",
         "prop-types": "15.6.2",
         "react": "16.8.6",
+        "react-app-rewire-hot-loader": "2.0.1",
+        "react-app-rewired": "2.1.3",
         "react-copy-to-clipboard": "5.0.1",
         "react-dom": "16.8.6",
+        "react-hot-loader": "4.9.0",
         "react-map-gl": "3.2.8",
         "react-redux": "5.1.0",
         "react-router": "4.3.1",
@@ -36,11 +41,11 @@
         "viewport-mercator-project": "5.1.0"
     },
     "scripts": {
-        "start": "PORT=6543 react-scripts start",
-        "build": "react-scripts build",
-        "test": "CI=1 react-scripts test --env=node",
+        "start": "PORT=6543 react-app-rewired start",
+        "build": "react-app-rewired build",
+        "test": "CI=1 react-app-rewired test --env=node",
         "lint": "eslint src/ --ext .js --ext .jsx",
-        "eject": "react-scripts eject"
+        "eject": "react-app-rewired eject"
     },
     "devDependencies": {
         "@babel/cli": "7.0.0-beta.49",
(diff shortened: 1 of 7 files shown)


jwalgran:
How concerned should we be about the 🚨, ☠️, and ⚠️ at the top of the rewired README?

kellyi:
> How concerned should we be about the 🚨, ☠️, and ⚠️ at the top of the rewired README?

Seems like the ⚠️ is the only one that's not just decorative, and that's essentially a "you break it, you bought it" caveat. An alternative possibility for this would be that we could eject from Create React App, but we can also eject later if we encounter something insurmountable with rewired.

FWIW apparently Raster Foundry made use of react-app-rewired in a production app used internally, so there's some experience with it here already.

rajadain:
Taking a look now.

src/app/package.json
```diff
@@ -6,21 +6,25 @@
     "dependencies": {
         "@google/markerclusterer": "1.0.3",
         "@material-ui/core": "3.1.0",
+        "@hot-loader/react-dom": "16.8.6",
```
rajadain: Can this  be moved up a line to maintain alphabetical ordering?

src/app/package.json
```diff
@@ -6,21 +6,25 @@
...
         "immutability-helper": "2.9.0",
         "lodash": "4.17.11",
         "moment": "2.24.0",
         "prop-types": "15.6.2",
         "react": "16.8.6",
+        "react-app-rewire-hot-loader": "2.0.1",
+        "react-app-rewired": "2.1.3",
         "react-copy-to-clipboard": "5.0.1",
         "react-dom": "16.8.6",
-        "react-map-gl": "3.2.8",
```
rajadain: Is the removal of `react-map-gl` intentional?

src/app/package.json
kellyi: Oh yeah -- I originally did have it alphabetically but forgot to make it so again after rebasing on develop.

src/app/package.json
kellyi: Yeah -- it is d13e49e5aee6d50d98f2b901be94be3f92243416

react-map-gl was used for the old Mapbox map but since we no longer have that we no longer need the library. I removed it here because I saw it but kept it to a separate commit.

kellyi:
Thanks! https://github.com/open-apparel-registry/open-apparel-registry/issues/558 is for the URL routes.

src/app/package.json
kellyi: I moved this in a new fixup commit then squashed it back into the original.

kellyi:
Thanks for taking a look!

src/app/src/setupProxy.js
```diff
@@ -0,0 +1,24 @@
...
+]);
+
+const djangoProxyTarget = Object.freeze({ target: 'http://django:8081' });
+
+const createProxies = app => _.forEach(
+    pathsToProxy,
+    path => app.use(proxy(path, djangoProxyTarget)),
+);
+
+module.exports = createProxies;
```
jwalgran: I see statements like

```
[HPM] Proxy created: /api  ->  http://django:8081
```

being printed to my console, but I don't see in this PR where `createProxies` is being called. How is this wired in to the dev server?

kellyi: If a `src/setupProxies.js` file exists, the default export gets called as a function by Create React App: https://facebook.github.io/create-react-app/docs/proxying-api-requests-in-development#configuring-the-proxy-manually

jwalgran: 👍 Thanks for providing that pointer to the documentation.

Issue #547, opened 2019-05-29
Superusers can view all lists


jwalgran:
## Overview

Allow superusers to view all lists but not modify them.

### Is your feature request related to a problem? Please describe.

It is difficult to answer support questions that contributors have about their uploaded lists without being able to see them.

### Describe the solution you'd like

- Allow superusers to browse any list detail page.
- Add a superuser dashboard page.
  - After choosing a contributor from a single-select dropdown links to all of the contributors lists are displayed. Show upload date for each and order by upload date descending.

kellyi:
https://github.com/open-apparel-registry/open-apparel-registry/issues/546 was merged so this is no longer blocked. However, we should coordinate routing because https://github.com/open-apparel-registry/open-apparel-registry/issues/517 will also involve working with this dashboard route.

I was thinking that `/dashboard` could be a page that had links to two subpages:

- `/dashboard/claims` for working with facility claims
- `/dashboard/lists` for working with facility lists

These routes will have to be set up in a switch in the current Dashboard component, which catches `/dashboard` at the top level. We'll also need some kind of minimal interface in the Dashboard component which has two react-router-dom `Link`s to `/dashboard/claims` and `/dashboard/lists`

kellyi:
https://github.com/open-apparel-registry/open-apparel-registry/pull/557 adds the routing described above for the dashboard page

Issue #517, opened 2019-05-15
Enable users to annotate, verify, reject, or revoke facility claims through React dashboard


kellyi:
## Overview

Following #516 we should enable users to make notes about, verify, or reject a facility claim.

Notes should potentially be a separate model, many-to-one with FacilityClaim.

This work should remain behind a feature flag.

### Describe the solution you'd like

#516 should add a table listing facility claims. For this feature, clicking on one of the items should go to a `/dashboard/claims/claimID` type route which will be a placeholder for displaying data about a facility claim so that dashboard users can annotate, verify, reject, or revoke facility claims.

For annotations, add a FacilityClaimNote model which is many-to-one with the claim. This model will be used by dashboard users to document actions taken around claiming a facility.

Add an POST/PUT API endpoint for saving annotations along with endpoints for accepting, rejecting, or revoking a claim.

It probably also makes sense to log accept/reject/revoke type actions like we do in Credit/Appeals.

PR #557, opened 2019-06-04
Add /dashboard/lists and /dashboard/claims routing

kellyi:
## Overview

To split work on #547 from work on #517 this PR sets up some routing for `/dashboard/lists` and `/dashboard/claims` routes, with the claims routes behind a feature flag. It also adds some placeholder components for each route.

- add /dashboard/lists and /dashboard/claims routing
- add placeholder DashboardLists and DashboardClaims components to
disentangle work on viewing contributor lists from work on viewing
facility claims
- add routing & links on main Dashboard components to lists and claims,
placing the claims route behind a feature flag

Connects #517 

## Demo

feature switch:

Screen Shot 2019-06-04 at 12 25 55 PM

with feature switched off:

Screen Shot 2019-06-04 at 12 26 03 PM

with feature switched on:

Screen Shot 2019-06-04 at 12 26 10 PM

/dashboard/lists:

Screen Shot 2019-06-04 at 12 26 19 PM

/dashboard/claims:

Screen Shot 2019-06-04 at 12 26 13 PM

## Testing Instructions

- serve this branch, `./scripts/manage resetdb`, `./scripts/manage createsuperuser`, then sign in to the app as your superuser on 6543 and visit `/dashboard`
- if you haven't turned the feature flag for claim a facility on, you should only see the lists link on the dashboard page. Clicking it should take you to the placeholder component and update the route to `/dashboard/lists`
- with the feature flag still turned *off*, visit `/dashboard/claims` directly and verify that you just see the main dashboard page again
- turn the CLAIM_A_FACILITY switch *on* in the django-admin, save the change, then refresh `/dashboard/claims`. You should now see the claims placeholder component
- visit `/dashboard` again and verify that you now see two links, both of which work

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/components/Dashboard.jsx
+++ b/src/app/src/components/Dashboard.jsx
@@ -2,10 +2,23 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { bool } from 'prop-types';
 import CircularProgress from '@material-ui/core/CircularProgress';
+import { Link, Route, Switch } from 'react-router-dom';
+
+import DashboardLists from './DashboardLists';
+import DashboardClaims from './DashboardClaims';
+import FeatureFlag from './FeatureFlag';
 
 import { checkWhetherUserHasDashboardAccess } from '../util/util';
+import { CLAIM_A_FACILITY, dashboardListsRoute, dashboardClaimsRoute } from '../util/constants';
 
 import AppGrid from './AppGrid';
+
+const dashboardStyles = Object.freeze({
+    linkSectionStyles: Object.freeze({
+        display: 'flex',
+        flexDirection: 'column',
+    }),
+});
 
 function Dashboard({
     userWithAccessHasSignedIn,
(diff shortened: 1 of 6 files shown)


src/app/src/components/Dashboard.jsx
```diff
@@ -27,9 +40,38 @@ function Dashboard({
...
+                <Route path={dashboardListsRoute} component={DashboardLists} />
+                <Route
+                    path={dashboardClaimsRoute}
+                    render={
+                        () => (
+                            <FeatureFlag
+                                flag={CLAIM_A_FACILITY}
+                                alternative={linkSection}
+                            >
+                                <Route component={DashboardClaims} />
```
kellyi: The nesting here is to enable using the `FeatureFlag` component while keeping react router props in DashboardClaims

src/app/src/components/Dashboard.jsx
```diff
@@ -27,9 +40,38 @@ function Dashboard({
         );
     }
 
+    const linkSection = (
+        <div style={dashboardStyles.linkSectionStyles}>
+            <Link to={dashboardListsRoute}>
+                View Contributor Lists
+            </Link>
+            <FeatureFlag flag={CLAIM_A_FACILITY}>
```
kellyi: Without an `alternative` prop this will just be `null`.

kellyi:
Oops, I just realized that the `alternative` prop for `FeatureFlag` doesn't exist in develop yet because it is still on #540. I'll bring that into this.

kellyi:
9accec5 pulls the `alternative` prop for `FeatureFlag` in.

rajadain:
With the feature flag turned off, going to `/dashboard/claims` shows the regular dashboard, but the URL is not updated:

image

Can we make it redirect to `/dashboard`?

kellyi:
> Can we make it redirect to /dashboard

I'd prefer not to have it redirect -- the `/dashboard/claims` behavior with the feature flag off is how it'd work with any other random route, too, like `/dashboard/hello` since the last `Route` in the `Switch` statement catches all nonmatches by virtue of not having a `path` declared. It'd probably possible to have `/claims` redirect, but the other routes still won't.

rajadain:
> how it'd work with any other random route, too, like `/dashboard/hello` since the last `Route` in the `Switch` statement catches all nonmatches by virtue of not having a path declared

Oh hmm. That's a little confusing. As a user I expect the URL routes to be meaningful, and if parts of it don't apply then to have they discarded. Changing this behavior from our routing system is outside the scope of this card, and I agree in its presence we shouldn't make a one-off redirect. But ideally the URL would faithfully represent the current state of the application. I'll make a card for it.

kellyi:
> But ideally the URL would faithfully represent the current state of the application.

Makes sense. I think we can accomplish this by:

- replacing the defaults for current `Switch`es with declared, specific `path`s with an `exact` prop
- generally adding the `NoRouteMatch` component as the new default, non-`path`/non-`exact` case

kellyi:
Cool -- thanks for testing this!

Issue #545, opened 2019-05-29
Add user/brand verification process


jwalgran:
## Overview

Add a process by which a contributor can be marked as verified by a system administrator. 

### Describe the solution you'd like

- Opt the `Contributor` model into tracking via `django-simple-history`. 
  - Update the next release issue with a step to run the management command to insert initial history records 
- Add `is_verified` (bool) and `verification_notes` (long text) columns to `Contributor` and make sure they appear properly in the Django admin. 
- Show a verified badge on profile page
- Show a verified badge next to the contributor name on the facility detail map sidebar.

### Additional Context

We are starting with the minimum viable process for marking a brand as verified and recording the reason why.

PR #563, opened 2019-06-06
Allow contributors to be verified

jwalgran:
## Overview

Allow superusers to mark contributors as verified and show that verification status on the facility detail page and the contributor profile page.

Connects #545

### Public Profile

Screen Shot 2019-06-06 at 1 40 18 PM

## My Profile Is Unchanged

Screen Shot 2019-06-06 at 1 40 32 PM

## Facility Details

Screen Shot 2019-06-06 at 1 40 58 PM

## Admin

Screen Shot 2019-06-06 at 1 42 10 PM

Screen Shot 2019-06-06 at 1 44 26 PM

## Notes

Adding a verified flag and a description box accessible via the Django Admin is the simplest possible version of this feature. We can grow a more complicated process from this if we discover a need. We opt the model into history tracking so that we have a log of changes to the verified status.

We have opted to show the name of the contributor as the title on the profile page rather than the static string "Profile" to emphasize what the page is about, to save some vertical space, and to provide a high-profile place on the page to show the verification badge.

We modified the `title` property on `AppGrid` to accept `node`s so we can include the icon component.

Showing the verified badge next to a contributor on the facility details page
requires us to have the verified status of contributors available in the
facility details API response. The previous implementation of the
`Facility.contributors` model method was returning a string representation of
the contributor. To better separate concerns, this helper now returns a list of
`FacilityList` objects that contributed the Facility, leaving the rendering up
to the serializer. We are now returning a list of objects that include an
`is_verified` flag that we use to conditionally render a verified SVG icon.

In order to reuse the same icon we used on the profile page, we are now passing
through the `style` prop from `BadgeVerified` to the internal `SvgIcon` in order
to adjust the size, margins, etc.

This PR also includes a unrelated change to the `Contributor` serializer that avoids multiple database calls. This was made before realizing that we would not need to change the `Contributor` serializer to implement the feature, but it is a useful change anyway.

## Testing Instructions

These instructions assume your database contains fixture data.

* Run migrations with `./scripts/manage migrate`
* Seed the new history data with `./scripts/manage populate_history --auto`
* Log in as a superuser, browse http://localhost:8081/admin/api/contributor/2/change/, and mark the contributor as verified.
* Browse http://localhost:6543/profile/2 and verify that the verified icon is displayed and that a "Verified" tooltip is shown on hover
* Browse http://localhost:6543/profile/3 and verify that no icon is displayed
* Browse http://localhost:6543/ and filter by contributor 2. Select a facility and verify that the verified icon is shown next to the contributor line on the detail page
* Filter by a different contributor and verify that there are no badges on the facility detail pages.
* Verify that the sample response on http://localhost:8081/api/docs/#!/facilities/facilities_read matches the actual API response shape.

## Checklist

- [x] CI passes after rebase
- [x] Create a release issue and include a step for running `./scripts/manage populate_history --auto`


kellyi:
> AppGrid only accepts strings for its title property so we are manually
creating an <h2> with the same style so we can show an icon as well as text.

In another context I was thinking that we may want to consider changing the `title` prop on this component to be a `node` which could either be a string or another React component. Alternatively, we could just give it a `component` prop that would render in a way that superseded the `title` if it was there.

I don't know that we have to do this here if the `h2` works well enough though.

kellyi:
Consider giving the verification badge span element a small amount of padding:

Screen Shot 2019-06-07 at 10 29 01 AM

jwalgran:
I have a migration conflict. Will be rebasing and fixing.

kellyi:
I tested out the main workflows for this and everything's working as expected!

jwalgran:
I pushed a rebase:
- Fixed the migration conflict
- Updated AppGrid to accept `node` for `title`
- Added padding around the verified badge on the profile page
Ready for a rereview

jwalgran:
Thanks for the review.

PR #568, opened 2019-06-07
Brand verification fixes

jwalgran:
## Overview

Corrects 2 issues with the initial brand verification implementation

* Links to profile pages from the facility details page were incorrect because we were using Contributor ID instead of User ID
* The verified badge was not shown on the profile of the currently logged in user.

Connects #545

## Demo

Screen Shot 2019-06-07 at 12 23 50 PM

## Testing Instructions

This assumes you have the fixture data loaded

* Log in as and admin (i.e. c1@example.com)
* Browse http://localhost:8081/admin/api/contributor/2/change/ and mark the contributor as verified.
* Log in as `c2@example.com` and browse http://localhost:6543/profile/2. Verify that the verified badge is displayed next to "My Profile"
* Browse http://localhost:6543 and open a facility detail page. Verify that the link to the profile page is correct.

## Checklist

- [x] CI passes after rebase
- [ ] ~CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines~ Skipped because this is a fix to a previous feature.

### DIFF
--- a/src/app/src/components/UserProfile.jsx
+++ b/src/app/src/components/UserProfile.jsx
@@ -12,6 +12,7 @@ import FacilityListSummary from './FacilityListSummary';
 import UserProfileField from './UserProfileField';
 import UserAPITokens from './UserAPITokens';
 import BadgeVerified from './BadgeVerified';
+import ShowOnly from './ShowOnly';
 import COLOURS from '../util/COLOURS';
 
 import '../styles/css/specialStates.css';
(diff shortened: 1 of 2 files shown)


jwalgran:
I discovered the link issue while testing on staging, where user ID and contributor ID are not the same value for each contributor.

kellyi:
Tested this out and everything is working as expected except that the verified badge doesn't seem to be shown on public profile pages when a contributor is verified, although it will show on the editable profile page when the user is logged in:

Screen Shot 2019-06-10 at 1 12 18 PM

Screen Shot 2019-06-10 at 1 15 27 PM

jwalgran:
Silly mistake. Pushed a fixup. Ready for another look.

jwalgran:
Thanks for review and catching that mistake.

Issue #543, opened 2019-05-29
Change "My Facilities" from link to uploaded facilities to list of claimed facilities


kellyi:
## Overview

Currently the "My Facilities" link leads to a facilities search pre-populated with the contributor id. However, we want to repurpose the link to make it go to a list of a user's successfully claimed facilities.

We also want to move the current "My facilities" functionality to the /lists page

### Describe the solution you'd like

- add a link on the /lists page which has the current "My Facilities" behavior
- update the "My Facilities" link on the tab to lead to a new route, like `/claimed` or similar, which will display a list of successfully claimed facilities each of which is linked to the claimed facility details page.

Issue #518, opened 2019-05-15
Add minimal claimed facility profile page and form


kellyi:
## Overview

Add a minimal claimed facility profile page and form.

Enable facility owners who have claimed facilities -- and had those claims accepted -- to edit the profile form.

This work should all live behind a feature flag.

## Desired Solution

- add a minimal facility profile page and form, to live on a route like `/facililty/OARID/details` or similar
- for most users this should just show a non-editable profile page
- for the user who has claimed that facility, it should be a form which can be used to edit the details
- editing should save to whatever model. We don't want to edit the Facility model's name, address, and location fields directly so we will want to add `claimed_name`, `claimed_address`, and `claimed_point` fields.
- add API endpoints for serving and saving details
- on saving updated details, send emails to users who are associated with that facility because they have uploaded it on a list to let them know that the data has changed.

This work should live behind a feature flag. We should also confer with @designmatty

PR #572, opened 2019-06-10
Add claimed/ routing

kellyi:
## Overview

- add React routing for `/claimed` and `/claimed/:claimID`, along with
components for routing
- update FeatureFlag component & Redux actions to display nothing while
the fetching is underway. This fixes a bug whereby the `alternative`
node would get rendered while the request to get all feature flags was
underway.

Connects #543 
Connects #518 

## Demo

Screen Shot 2019-06-10 at 2 12 52 PM
Screen Shot 2019-06-10 at 2 13 07 PM

## Testing Instructions

- serve this branch then visit `/claimed` and `/claimed/hello`, respectively, with the claim a facility feature flag on and verify that the routes both work -- and that nothing renders until the feature flag request returns
- switch off the claim a facility feature, then refresh the `/claimed` route and verify that you see the route not found message:

Screen Shot 2019-06-10 at 2 14 58 PM

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Enable submitting claim a facility form [#540](https://github.com/open-apparel-registry/open-apparel-registry/pull/540)
 - Allow contributors to be verified [#563](https://github.com/open-apparel-registry/open-apparel-registry/pull/563)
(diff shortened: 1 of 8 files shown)


kellyi:
> What do you think about nesting claimed under facilities

Happy to make this change if we'd like -- originally I was thinking that `/claimed` was a sort of top-level concern for users and not a resource under facilities, but the mapping between React routes & resources is already kind of tenuous anyway so if `/facilities/claimed` won't conflict with the facility details route then we can use that.

kellyi:
Cool -- I'm going to leave it at the top-level then. Thanks for the rapid review!

PR #573, opened 2019-06-10
Show a list of facilities successfully claimed by the current contributor

jwalgran:
## Overview

Change the "My Facilities" link in the navbar to link to a new page showing a list of all approved facility claims for the currently logged in contributor.

Connects #543

### My Facilities

Screen Shot 2019-06-12 at 4 03 58 PM

### Empty List

Screen Shot 2019-06-12 at 4 03 38 PM

### Claim A Facility Disabled

Screen Shot 2019-06-12 at 4 06 08 PM

## Notes

The wireframes include showing the status of a claim and removing the claim but these are out of scope of #543 which specifies showing only approved claims. 

## Testing Instructions

This assumes that the fixture data is loaded and processed
* Disable claim a facility with `./scripts/manage waffle_switch claim_a_facility off`. Restart `./scripts.server.
* Log in as `c2@example.com` and verify that "My Facilities" does not appear in the navbar dropdown.
* Enable claim a facility with `./scripts/manage waffle_switch claim_a_facility on`. Restart `./scripts.server.
* Log in as `c2@example.com`, search for a facility, and submit a claim.
* Select "My Facilities" from the navbar dropdown and verify that a "no approved claims" message is shown.
* In a separate browser log in as `c1@example.com` and browse http://localhost:6543/dashboard/claims. Approve the facility claim.
* Back in the original browser select "My Facilities" from the navbar dropdown and verify that the approved facility now appears and that clicking on it navigates to a placeholder details page.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add routing to enable users to view claimed facilities [#572](https://github.com/open-apparel-registry/open-apparel-registry/pull/572)
 - Add claim a facility dashboard [#559](https://github.com/open-apparel-registry/open-apparel-registry/pull/559)
 - Add profile page and update form for approved facility claims [#575](https://github.com/open-apparel-registry/open-apparel-registry/pull/575)
+- Show a list of facilities successfully claimed by the current contributor [#573](https://github.com/open-apparel-registry/open-apparel-registry/pull/573)
 
 ### Changed
 - Adjust /claimed routing container [#574](https://github.com/open-apparel-registry/open-apparel-registry/pull/574)
--- a/src/app/src/actions/claimedFacilities.jsx
+++ b/src/app/src/actions/claimedFacilities.jsx
@@ -1,0 +1,28 @@
+import { createAction } from 'redux-act';
+
+import csrfRequest from '../util/csrfRequest';
+
+import {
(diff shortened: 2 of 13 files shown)


kellyi:
I made https://github.com/open-apparel-registry/open-apparel-registry/issues/586 for the wire-frame-related updates.

kellyi:
I tested this out and it's working almost perfectly! There are two small UI things that may not be blocking on merging, but which I did want to surface in case we would like to address them here or subsequently:

First: in Firefox the "no claimed facilities" message doesn't place the button beneath the message. I think this can be fixed just by putting the two elements in a common container, then giving it `display: flex` and `flex-direction: column`:

Screen Shot 2019-06-12 at 9 17 53 PM

Second, also in Firefox: when I click a row to go to a facility details page and then click the back button to return to the list page, for a split second I can see the "no claimed facilities" message before searching commences and it changes to the CircularProgress spinner. The back button seemed to be the only condition in which I could create the split-second quirk. It's not a show-stopper, though, so we could live with it if it's not easy to remedy.

jwalgran:
Thanks for review and pointing out those two issues.
I was able to fix the button-on-same-line issue but just changing the fragment to a div. No style needed.

We paired on the flash of submit button issues and solved it by paying attention to whether data is null or an empty list.

kellyi:
Ooof -- sorry about this but I introduced a merge conflict by merging #575

PR #584, opened 2019-06-12
Superusers Can View All Lists

rajadain:
## Overview

Adds a list view in the dashboard for Superusers that allows them to see a read-only view of any list any contributor has uploaded. The view is very similar to that of My Lists, with the inclusion of the Date Created column. Upon clicking through to a list, they are presented with a "read-only" view of the list, that allows for searching and filtering as usual, but does not allow confirming or rejecting potential matches.

Connects #547 

## Demo

image

## Notes

To facilitate the front-end, a few fields were exposed in the back-end that were previously not included: `is_superuser` for the user, `created_at` and `contributor_id` for the lists. If there are any security implications for exposing these fields, we should consider and address them now.

Much of the front-end was copied from other parts of the app, leading to a lot of duplication and redundancy in terms of styling and table markup. This is a good candidate for refactoring and abstracting, that can be done in a future pass.

There is an edge case when a superuser accesses their own list from the Dashboard Lists page. In this case, when they click "Back to Lists", they will be taken to their My Lists page instead. This is unlikely to be a huge issue immediately, since most Superusers don't have their own lists, but should be kept in mind. This can be addressed by changing "Back to Lists" from a link to simulate the browser's "Back" button, but I don't know if React Router provides such facility. It could also be implemented by setting a flag somewhere (either in the URL or in Redux), but I did not deem it important enough to include in this pass. Please let me know if I should.

## Testing Instructions

- Check out this branch and `server`
- Go to [:6543](http://localhost:6543/) and sign in as a Superuser
  - A Superuser is any user that has `is_superuser=True` in the database
- Go to [:6543/dashboard/lists/](http://localhost:6543/dashboard/lists/)
  - [ ] Ensure you see a table and a select
- Open the select box
  - [ ] Ensure it is populated with a list of contributors
- Select a contributor
  - [ ] Ensure the contributor's lists are fetched from the server and populate the table
  - [ ] Ensure the lists are sorted newest first
  - [ ] Ensure the URL is updated with the selected contributor
- Click a list
  - [ ] Ensure you are taken to that list's detail page
- Observe that the view is "read-only"
  - [ ] Ensure searching / filtering works as usual
  - [ ] Ensure you cannot confirm or reject potential matches
- Click "Back to lists"
  - [ ] Ensure you are taken to Dashboard Lists, rather than your My Lists page
  - [ ] Ensure Dashboard Lists has the correct contributor selected in the view and the URL
- Refresh the page
  - [ ] Ensure the contributor is picked up from the URL and used to populate the select and the table
- Select a contributor that hasn't uploaded any lists
  - [ ] Ensure you see an empty message

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/util/util.js
+++ b/src/app/src/util/util.js
@@ -83,6 +83,8 @@ export const makeFacilityListsURL = () => '/api/facility-lists/';
 export const makeSingleFacilityListURL = id => `/api/facility-lists/${id}/`;
 export const makeSingleFacilityListItemsURL = id => `/api/facility-lists/${id}/items/`;
 
+export const makeDashboardFacilityListsURL = contributorID => `/api/facility-lists/?contributor=${contributorID}`;
+
 export const makeAPITokenURL = () => '/api-token-auth/';
 
 export const makeGetContributorsURL = () => '/api/contributors/';
@@ -252,6 +254,18 @@ export const getTokenFromQueryString = (qs) => {
         : token;
 };
 
+export const getContributorFromQueryString = (qs) => {
+    const qsToParse = startsWith(qs, '?')
+        ? qs.slice(1)
+        : qs;
+
+    const {
+        contributor = null,
+    } = querystring.parse(qsToParse);
+
+    return parseInt(contributor, 10);
+};
+
 export const allFiltersAreEmpty = filters => values(filters)
     .reduce((acc, next) => {
         if (!isEmpty(next)) {
@@ -384,6 +398,8 @@ export const makeClaimFacilityLink = oarID => `${facilitiesRoute}/${oarID}/claim
 export const makeApprovedClaimDetailsLink = claimID => `/claimed/${claimID}`;
 
 export const makeFacilityClaimDetailsLink = claimID => `/dashboard/claims/${claimID}`;
+
+export const makeDashboardContributorListLink = contributorID => `/dashboard/lists/?contributor=${contributorID}`;
 
 export const makeProfileRouteLink = userID => `/profile/${userID}`;
 
--- a/src/django/api/views.py
+++ b/src/django/api/views.py
@@ -39,6 +39,7 @@ from oar.settings import MAX_UPLOADED_FILE_SIZE_IN_BYTES, ENVIRONMENT
 
 from api.constants import (CsvHeaderField,
                            FacilitiesQueryParams,
+                           FacilityListQueryParams,
                            FacilityListItemsQueryParams,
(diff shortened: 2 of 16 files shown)


kellyi:
> This can be addressed by changing "Back to Lists" from a link to simulate the browser's "Back" button, but I don't know if React Router provides such facility.

I believe `history.pop()` should work.

src/app/src/util/util.js
```diff
@@ -248,6 +250,18 @@ export const getTokenFromQueryString = (qs) => {
...
+    const qsToParse = startsWith(qs, '?')
+        ? qs.slice(1)
+        : qs;
+
+    const {
+        contributor = null,
+    } = querystring.parse(qsToParse);
+
+    return parseInt(contributor, 10);
+};
```
kellyi: Consider adding a test for this new utility function.

src/django/api/views.py
```diff
@@ -796,9 +798,29 @@ def list(self, request):
             ]
         """
         try:
-            contributor = request.user.contributor
-            queryset = FacilityList.objects.filter(contributor=contributor)
-            response_data = self.serializer_class(queryset, many=True).data
+            if request.user.is_superuser:
```
kellyi: I can't remember if we have tests for these endpoints already, but consider adding tests for each of the codepaths.

rajadain:
I'll see if I can make the table wider. Currently there's a lot of whitespace around it.

kellyi:
Heads up that there is a merge conflict.

kellyi:
This commit reference was a mistake, and I amended it to refer to the correct issue:

Screen Shot 2019-06-13 at 1 33 09 PM

rajadain:
I dropped the processing detail columns, which makes it more manageable:

image

Unfortunately, padding comes from further up the DOM, and I didn't want to pop the table out of that arrangement (using something like absolute positioning) just for this. Is this good enough for now?

jwalgran:
Definitely. I am ok with deferring it to an app-wide table redesign in the future.

rajadain:
> How will this work with `?contributor=5&contributor=10`?

Added a test case for that. It'll take the first one and ignore the rest.

rajadain:
Added API tests in 3661555. I'm going to make follow up issues for making the table wider, and for fixing the back button. Will merge this shortly after.

rajadain:
Made #602 and #603 for the follow up issues. Going to merge this now. Thank you both for the thorough reviews!

Issue #581, opened 2019-06-12
Ensure only one Facility Claim can be approved per Facility


kellyi:
## Overview

Each Facility should be able to have at most one FacilityClaim approved at any time. We should enforce this in the `approve` endpoint handler and elsewhere.

### Describe the solution you'd like

Each Facility should be able to have at most one FacilityClaim approved at any time. We should enforce this in the `approve` endpoint handler and elsewhere.

PR #585, opened 2019-06-13
Ensure at most one approved claim per facility

kellyi:
## Overview

When attempting to approve a facility claim, check whether there are
already any other approved facility claims and, if so, return a bad
request response. Also add a test to demonstrate the feature
functionality.

Connects #581 

## Testing Instructions

It should suffice just to run the tests and verify that they continue to pass.

You can also go through the loop of:

- turn on claim a facility
- sign in as `c2@example.com`
- submit the claim form 2x for the same facility
- sign out, then sign in as a superuser
- visit /dashboard/claims
- approve one of the facilities and verify that it works
- attempt to approve the other facility and verify that it does not work and that you see an error

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/tests.py
+++ b/src/django/api/tests.py
@@ -1394,6 +1394,49 @@ class FacilityClaimAdminDashboardTests(APITestCase):
         self.assertEqual(400, error_response.status_code)
 
     @override_switch('claim_a_facility', active=True)
+    def test_can_approve_at_most_one_facility_claim(self):
+        response = self.client.post(
+            '/api/facility-claims/{}/approve/'.format(self.facility_claim.id)
+        )
+
+        self.assertEqual(200, response.status_code)
+
+        updated_facility_claim = FacilityClaim \
+            .objects \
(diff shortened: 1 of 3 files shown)


src/django/api/tests.py
```diff
@@ -1393,6 +1393,47 @@ def test_approve_facility_claim(self):
 
         self.assertEqual(400, error_response.status_code)
 
+    @override_switch('claim_a_facility', active=True)
+    def test_can_approve_at_most_one_facility_claim(self):
+        self.client.post(
```
kellyi: For the sake of completeness, I will also add a check that the response status from this first request is 200.

kellyi: Done.

rajadain:
Taking a look now.

kellyi:
Thanks for testing this! For this:

> It would've been nicer if the error explained that the reason for failure was the presence of an approve claim, rather than the generic "there was an error"

The server error will be logged to the browser console, so we can always get it that way we encounter this again.

PR #592, opened 2019-06-13
Add .prettierrc.json config file

kellyi:
## Overview

Add a .prettierrc.json configuration file with rules which are
compatible with our existing set of ESLint rules. This enables
developers to run Prettier manually in their editors on new files while
working, saving the burden of having to manually format the code.

Connects #549 

## Notes

I intentionally did not set this up to run prettier on file saves,
although that is theoretically possible now that we are using
react-app-rewired. Setting it up to run on saves would cause a large
initial changeset because many of the project's files have not been
formatted with Prettier.

## Testing Instructions

Try running Prettier locally on this branch in one of the JSX files and verify that it reformats according to the formatting rules here.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add profile page and update form for approved facility claims [#575](https://github.com/open-apparel-registry/open-apparel-registry/pull/575)
 - Show a list of facilities successfully claimed by the current contributor [#573](https://github.com/open-apparel-registry/open-apparel-registry/pull/573)
 - Add GitHub issue template for "draft" issues [#590](https://github.com/open-apparel-registry/open-apparel-registry/pull/590)
+- Add .prettierrc.json configuration file for JS code formatting during development [#592](https://github.com/open-apparel-registry/open-apparel-registry/pull/592)
 
 ### Changed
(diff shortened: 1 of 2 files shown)


kellyi:
I've been using this file locally for the last sprint, btw, and it's worked well. Sometimes it does introduce slight quirks but those can be remedied manually if we want to keep the prior idiom for certain things.

jwalgran:
Looking at this now

kellyi:
Oh interesting. Thanks for catching those. In other projects we have removed the ESLint rules related to those errors and let Prettier take over. However, here I think we want to adjust the Prettier config to keep ESLint in place. I'll spend a very small amount of time seeing if I can remedy this and push an update if so.

kellyi:
Unfortunately I think this is as close as we can get unless we want to adjust some ESLint rules, since the `arrowParens` options for Prettier are either on or off:

https://prettier.io/docs/en/options.html#arrow-function-parentheses

If we want to adjust the ESLint rules and make the associated code formatting changes, we should wait to do so until next month when there's a lull in development.

For this PR we can either wait until then to complete it or merge it now and live with the errors. Either is fine for me; I guess the question around merging is whether the value of having a mostly-working Prettier config file more than the problem with having a partly-broken Prettier config file.

jwalgran:
I would rather not commit a formater config that is inharmonious with the linter. I'd rather wait and implement the broad change to our eslint config.

kellyi:
Sounds good to me.

Issue #555, opened 2019-06-03
Prevent users who have a pending claim on a facility from submitting another claim on that facility


kellyi:
## Overview

Prevent users who have a pending claim on a facility from submitting another claim on that facility

Depends on #515 

### Describe the solution you'd like

Two options I imagined:

1. adjust the facility details serializer to also send back a list of contributor IDs who have pending claims on the facility, then show a "you have already submitted a claim for this facility" notice if the currently-signed-in contributor is in that list

2. adjust the user serializer to include a list of the facilities a user has claimed and then show the same "you have already submitted a claim" notice, but this time checking the OAR ID against the current facility details ID

Option 1 is technically easier but unfortunately seems to leak some information, so option 2 might be preferable.

jwalgran:
I agree that "adjust the user serializer to include a list of the facilities a user has claimed" is a preferable approach.

kellyi:
As part of this work we should also make the following updates to the API endpoint:

- the server endpoint should reject posted facility claims when the user sending the claim already has a pending claim
- the server endpoint should also reject posted facility claims when there is already an approved facility claim for that facility

PR #601, opened 2019-06-14
Prevent submitting > 1 pending facility claim

kellyi:
## Overview

- update API to return a bad request error if a user attempts to claim a
facility for which that user already has a pending claim
- update API to send back a lists of OAR IDs for which users have
pending and approved facility claims, both on sign in and also on
submitting a new claim
- adjust the facility details UI to prevent a user from seeing claim a
facility links when a claim is pending
- add facility details sidebar message indicating "You have a pending claim on this facility" when
appropriate

Connects #555 

## Demo

Screen Shot 2019-06-14 at 11 45 08 AM

## Testing Instructions

- sign in as c8@example and try to submit more than one claim for the same facility
- you cannot

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/serializers.py
+++ b/src/django/api/serializers.py
@@ -103,16 +103,33 @@ class UserSerializer(ModelSerializer):
 
     def get_claimed_facility_ids(self, user):
         if not switch_is_active('claim_a_facility'):
-            return []
+            return {
+                'approved': None,
+                'pending': None,
+            }
 
         try:
-            return FacilityClaim \
+            approved = FacilityClaim \
                 .objects \
                 .filter(status=FacilityClaim.APPROVED) \
                 .filter(contributor=user.contributor) \
                 .values_list('facility__id', flat=True)
+
+            pending = FacilityClaim \
+                .objects \
+                .filter(status=FacilityClaim.PENDING) \
+                .filter(contributor=user.contributor) \
+                .values_list('facility__id', flat=True)
+
+            return {
+                'pending': pending,
+                'approved': approved,
+            }
         except Contributor.DoesNotExist:
-            return []
+            return {
+                'approved': None,
+                'pending': None,
+            }
 
 
 class UserProfileSerializer(ModelSerializer):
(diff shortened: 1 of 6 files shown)


src/django/api/serializers.py
```diff
@@ -103,16 +104,33 @@ def get_contributor_id(self, user):
...
                 .values_list('facility__id', flat=True)
+
+            pending = FacilityClaim \
+                .objects \
+                .filter(status=FacilityClaim.PENDING) \
+                .filter(contributor=user.contributor) \
+                .values_list('facility__id', flat=True),
+
+            return {
+                'pending': list(chain(*pending)),
```
kellyi: This is here to flatten the lists; somehow they became nested like [[123], [456]] when making this change.

jwalgran:
Looking at this now.

src/django/api/views.py
```diff
@@ -608,7 +621,22 @@ def claim(self, request, pk=None):
...
+                .objects \
+                .filter(status=FacilityClaim.APPROVED) \
+                .filter(contributor=contributor) \
+                .values_list('facility__id', flat=True)
+
+            pending = FacilityClaim \
+                .objects \
+                .filter(status=FacilityClaim.PENDING) \
+                .filter(contributor=contributor) \
+                .values_list('facility__id', flat=True),
```
jwalgran: We were bitten by the Python! There is a trailing comma hiding at the end of this line which ended up creating a tuple. I was able to remove that comma and the `list(chain(*))` wrappers and see things working as expected.

src/django/api/serializers.py
jwalgran: We were bitten by the Python! There is a trailing comma just above at the end of the `pending =` assignment which ended up creating a tuple. I was able to remove that comma and the list(chain(*)) wrappers and see things working as expected.

kellyi: Wow, thanks for catching that! I dropped the trailing comma and the `list(chain` call and everything worked as before. Will amend the commit to include the changes before merging.

src/django/api/views.py
kellyi: Thanks for catching that! I fixed this in amended commit a6dd71f, along with the instance in the serializers file.

kellyi:
Thanks again for catching the trailing comma thing! I fixed it and am going to merge this.

Issue #605, opened 2019-06-14
Create a FacilityAlias model [points covered by dependant issues]


jwalgran:
## Overview

To support merging create a `FacilityAlias` model that will allow for redirects from old OAR IDs to new ones.

### Is your feature request related to a problem? Please describe.

When merging facilities we want the OAR ID of the merged facility to redirect to the new facility.

When deleting a facility with matches, one of the matches will be promoted to being a new facility. The deleted OAR ID should redirect to the new facility.

### Describe the solution you'd like

- [ ] Create a FacilityAlias table
  - `oar_id` char field
  - `facility` foreign key to Facility with PROTECT
  - `reason` choice field ('MERGE', 'DELETE')

PR #612, opened 2019-06-17
 Add FacilityAlias model 

jwalgran:
## Overview

This tables allows us to keep track of the OAR IDs for facilities that are merged with other facilities or deleted so that we can report on their removal and redirect to their replacement facilities.

We applied the same primary key field settings to `oar_id` that were used for `Facility.id` because we only want a single alias for each OAR ID and we do not want it to change.

Connects #605

## Testing Instructions

* Verify that running `./scripts/update` applies the migration and creates the new table.

## Checklist

- [ ] CI passes after rebase
- [X] ~CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines~ Skipped because this is an internal change.


src/django/api/models.py
```diff
@@ -740,3 +740,37 @@ class Meta:
     def __str__(self):
         return '{0} - {1} - {2}'.format(self.facility_list_item, self.facility,
                                         self.status)
+
+
+class FaclilityAlias(models.Model):
```
kellyi: I set this up and it worked except that when I ran this in `shell_plus` it reported an error.

```
>>> FacilityAlias.objects.all()
  File "<console>", line 1, in <module>
NameError: name 'FacilityAlias' is not defined
```

Turns out there's a typo in the name of this model: `FaclilityAlias` -> `FacilityAlias`

jwalgran: Sorry for not giving this a more thorough test. Will fix.

jwalgran: I fixed the typo in 974d8da.

jwalgran:
This is ready for another look. I added a `__str__` method to make shell testing more clear.

src/django/api/models.py
kellyi: No worries! Checked out the newer commits and everything's working as expected now.

jwalgran:
Thanks for catching my goof.

Issue #597, opened 2019-06-13
Add a facility delete API


jwalgran:
## Overview

Depends on #604 "Opt FacilityMatch into django-simple-history tracking"
Depends on #605 "Create a FacilityAlias model"

Add a private superuser-only API that allows for the deletion of a `Facility`.

### Is your feature request related to a problem? Please describe.

Sometimes lists contain mistakes. We need to provide a way to get rid of junk data that should never have been uploaded.

### Describe the solution you'd like

- [x] Add an `DELETED` status for `FacilityListItem`
- [x] Implement the delete action of the `FacilityViewSet`
  - [x] check that the request is from a superuser
  - [x] Set the status the `FacilityListItem` that created the `Facility` to `DELETED`
  - [x] Add an entry to the `processing_results` array on the list item documenting the delete.
  - Delete the `FacilityMatch` row.
    - [x] Set the changeReason to "Deleted ${OAR ID}" https://django-simple-history.readthedocs.io/en/2.7.2/historical_model.html?#change-reason
  - [x] If there are other `FacilityMatch` rows that point to `Facility` being deleted
    - [x] Pick the one with the highest match confidence. 
    - [x] Create a new facility. Use `processing.save_match_details` as a reference.
    - [x] Update all remaining `FacilityMatch` rows to point to the new facility
      - [ ] Set the changeReason to "Deleted ${deleted OAR ID} and promoted ${promoted OAR ID}" https://django-simple-history.readthedocs.io/en/2.7.2/historical_model.html?#change-reason
    - [x] Update the `FacilityListItem` to point to the new facility
      - [x] Add an entry to the `processing_results` array on the list item documenting the promotion.
    - [x] Write a row into the `FacilityAlias` table to link the old OAR ID to the new one with a reason code `DELETE`

PR #616, opened 2019-06-18
Add Facility delete API

jwalgran:
## Overview

Add a superuser-only API for deleting facilities and updating any related matches.

Connects #597

### Delete method does not appear in the API docs

Screen Shot 2019-06-19 at 1 11 23 PM

### Deleted items are searchable on the list details page

Screen Shot 2019-06-19 at 1 14 48 PM

## Notes

Redirection using `FacilityAlias` is not implemented in this PR.

To allow for deletes we need to convert from using `ReadOnlyModelViewSet` to manually opting in to the list, retrieve, and delete methods.

We will be reserving this action for superusers, so we exclude it from the public API documentation by defining a custom `AutoSchema` subclass and overriding the appropriate method.

Adds a `DELETED` status for `FacilityListItem`s so we can correctly display and filter for them on the contributor facility list detail page.

Adds a `DELETE_FACILITY` processing action type. The delete operation appends an entry to the `processing_results` array on the `FacilityListItem` that includes the ID of the deleted `Facility`. This is not strictly required, but continues the theme of recording all the events related to the `FacilityListItem`.

Adds helper methods to `Facility` which fetch `FacilityMatch` records to help decrease the size of the `destroy` view function.

Returns a bad request response if there is an approved `FacilityClaim` object related to the `Facility`. We want to force superusers to revoke an existing claim before deleting a `Facility` to trigger a notification to the claimant.

Change foreign keys on FacilityMatch from CASCADE to PROTECT. Whenever we delete a `FacilityMatch` row we want to use the `changeReason` attribute provided by `django-simple-history`. Setting `PROTECT` on the relationship requires us to manually manage our delete process.

## Testing Instructions

* Test that migrations apply to an older DB version
  * `./scripts/manage migrate`
* Reset the DB
  * `./scripts/manage resetdb`
  * `./scripts/manage loadfixtures`
  * `./scripts/manage processfixtures`
* Log in as `c1@example.com` and generate an API key on the profile page. Save this for later.
* Log in as `c15@example.com`
* Browse http://localhost:6543/lists/15, search for `Regina Miracle Intimate` and verify that the status is `MATCHED`.
* On the map page, search for `Regina Miracle Intimate`. Verify that the facility has three contributors.
* Copy the OAR ID and submit a delete API request 
```
curl --request DELETE \
  --url http://localhost:8081/api/facilities/{{OAR ID}} \
  --header 'authorization: Token {{API KEY}}'
```
* Refresh the facility details page and verify it says `Not Found`
* Search for `Regina Miracle Intimate` and verify that the facility now has only 2 contributors.
* Browse http://localhost:6543/lists/15 and verify that the item now has a status of `NEW_FACILITY`

## Checklist

- [x] CI passes after rebase


kellyi:
Taking a look at this!

src/django/api/views.py
```diff
@@ -554,6 +569,131 @@ def retrieve(self, request, pk=None):
...
+            'started_at': now,
+            'error': False,
+            'finished_at': now,
+            'deleted_oar_id': facility.id,
+        })
+        list_item.facility = None
+        list_item.save()
+
+        match = facility.get_created_from_match()
+        match.changeReason = 'Deleted {}'.format(facility.id)
```
kellyi: Out of curiosity, is there a reason that `changeReason` is in camel case rather than snake case?

kellyi: (I may have missed this in a prior migration. I think it's probably fine to leave; just wanted to be deliberate about having it.)

kellyi: Oh -- I just realized that we got this from django-simple-history. Interesting choice on the library's part but I guess it could be a way to signal that it's a slightly different kind of field.

jwalgran: Yeah. It hurts a little :-)

Strange choice by the library author(s).

kellyi:
Small quirk we might want to fix here or else later: The main lists endpoint sends back a dictionary of status counts, and I think we need to add DELETED to it, as it's not coming through:

Screen Shot 2019-06-20 at 12 39 11 PM

Screen Shot 2019-06-20 at 12 40 22 PM

jwalgran:
>The main lists endpoint sends back a dictionary of status counts, and I think we need to add DELETED to it

Excellent catch. I pushed b213f3b to address this by adding a column to the facility list table.

Screen Shot 2019-06-20 at 10 45 48 AM

kellyi:
> by adding a column to the facility list table

👍

jwalgran:
Thanks for the review.

Issue #594, opened 2019-06-13
Allow a contributor to remove a matched or new facility item from a list


jwalgran:
## Overview

Depends on #604 	

Allow a contributor to remove an item from a list.

Remove is not delete. It behaves more like replacing a list, where the list item is no longer associated with the contributor but the data remains.

### Is your feature request related to a problem? Please describe.

If a brand breaks ties with a single facility we do not want to require them to upload and reprocess an entire list.

### Describe the solution you'd like

- [x] Add an `is_active` column to `FacilityMatch` with a default value of true.
- [x] Create a `remove` action on the `FacilityListViewSet` that takes the `list_item_id` as a param (similar to match confirmation/rejection)
  - [x] Set `is_active` to false on the `FacilityMatch` row
- [x] Show a "Remove" button for each **matched or new facility** row on the facility list details table.On click:
  - [x] Call the remove action
- [x] Style removed rows differently (consult with design). Do not show the remove button.
- [x] Incorporate the `is_active` flag into the facility details query so that the facility is not associated with the contributor.

PR #619, opened 2019-06-19
Enable contributors to remove individual facility list items from public display

kellyi:
## Overview

This PR enables list contributors to remove list items. Removing an item will keep it on a list but suppress it from displaying on the facility details page in any way. I designed this in line with the specs detailed in #594.

Connects #594 

## Demo

Screen Shot 2019-06-20 at 5 37 38 PM

Screen Shot 2019-06-20 at 5 37 46 PM

## Notes

In principle there is no reason why we cannot also remove list items with a status of "POTENTIAL_MATCH". However I did not have implement a UI for doing that, partly because the actions area for those rows is already kind of busy and partly because list items that are not yet matched don't show up on the facility details page anyway.

## Testing Instructions

- get branch then `./scripts/manage resetdb`, `./scripts/manage processfixtures`, `./scripts/server` and then sign in as c6@example
- in one tab, search for the facility named "Cotia". On its details page verify that there is a link to the contributor page for c6
- visit the facility lists for c6 and search for "Cotia". Use the UI to remove the facility and verify that the row changes its opacity on success and that you see the "REMOVED" status.
- try out the status filters and verify that "REMOVED" and the others work correctly
- look at the details page for "Cotia" and very that the contributor is no longer associated with the facility
- try removing some other facilities and verify that that also works. You may want to sign in as c2 or c8 and remove facilities from those lists, as you should test out that you can also remove "MATCHED" and "CONFIRMED_MATCH" list items -- and not just "NEW_FACILITY" items.

## Checklist

- [ ] CI passes after rebase


jwalgran:
Looking at this now.

src/app/src/components/FacilityListItemsTableRow.jsx
```diff
@@ -1,27 +1,43 @@
...
-const propsAreEqual = (prevProps, nextProps) =>
-    isEqual(prevProps.rowIndex, nextProps.rowIndex)
-    && isEqual(prevProps.countryCode, nextProps.countryCode)
-    && isEqual(prevProps.name, nextProps.name)
-    && isEqual(prevProps.address, nextProps.address)
-    && isEqual(prevProps.status, nextProps.status)
-    && isEqual(prevProps.hover, nextProps.hover)
-    && isEqual(prevProps.newFacility, nextProps.newFacility)
-    && isEqual(prevProps.oarID, nextProps.oarID);
+const makeTableRowStyles = cond([
```
jwalgran: I like the power that `cond` and `conforms` give us, but we should consider that in cases like this where we are mapping 2 values that don't have any complex interactions, a lower-order solution is more compact and arguably more easily understood.

```javascript
const makeTableRowStyles = ({ handleSelectRow, isRemoved}) => {
    const pointerStyle = isFunction(handleSelectRow) ? { cursor: 'pointer' } : {};
    const opacityStyle = isRemoved ? { opacity: '0.6' } : {};
    return merge(pointerStyle, opacityStyle);
}
```

src/app/src/components/FacilityListItemsDetailedTableRowCell.jsx
```diff
@@ -20,11 +22,50 @@ export default function FacilityListItemsDetailedTableRowCell({
...
+                    justifyContent: 'space-between',
+                    alignItems: 'center',
+                    width: '100%',
+                }}
+            >
+                <span style={{ marginRight: '5px' }}>
+                    {title}
+                </span>
+                <Button
+                    variant="contained"
```
jwalgran: Consider using a stock button. The full column of blue buttons draws a lot of focus. Contrast these with the approve/reject buttons, which we absolutely need the contributor to click on. The remove buttons are an optional feature that we may want to call less attention to.

Screen Shot 2019-06-24 at 12 33 51 PM


```diff
diff --git a/src/app/src/components/FacilityListItemsTableRow.jsx b/src/app/src/components/FacilityListItemsTableRow.jsx
index 55936de..8d0748e 100644
--- a/src/app/src/components/FacilityListItemsTableRow.jsx
+++ b/src/app/src/components/FacilityListItemsTableRow.jsx
@@ -113,7 +113,6 @@ const FacilityListItemsTableRow = ({
                                 {newFacility ? 'NEW_FACILITY' : status}
                             </span>
                             <Button
-                                variant="contained"
                                 color="primary"
                                 onClick={handleRemoveItem}
                                 style={{ marginLeft: '5px', marginRight: '5px' }}
```

kellyi: Sounds good -- I adjusted these buttons in 2a796e6430bf7879205973107e624f569d824495

src/app/src/components/FacilityListItemsTableRow.jsx
kellyi: I changed this in 90eb513ee385aaec7cc9275ca5a12b9f6731323c to be:

```
const makeTableRowStyles = ({ handleSelectRow, isRemoved }) => ({
    cursor: isFunction(handleSelectRow) ? 'pointer' : 'auto',
    opacity: isRemoved ? '0.6' : '1.0',
});
```

src/app/src/components/FacilityListItemsTable.jsx
```diff
@@ -498,6 +555,98 @@ class FacilityListItemsTable extends Component {
...
+                            ? (
+                                <>
+                                    <DialogTitle>
+                                        Remove {listItemToRemove.name}?
+                                    </DialogTitle>
+                                    <DialogContent>
+                                        <Typography
+                                            style={facilityListItemsTableStyles.dialogTextStyles}
+                                        >
+                                            Do you really want to remove this list item?
```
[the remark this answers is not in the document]
kellyi: Good idea -- I adjusted this in 6912094398cb11d84b3efc2d78a79fff5c12d50e

Screen Shot 2019-06-24 at 4 35 01 PM

src/django/api/views.py
```diff
@@ -1542,6 +1542,47 @@ def reject_match(self, request, pk=None):
...
+                .objects \
+                .filter(facility_list=facility_list) \
+                .get(pk=request.data.get('list_item_id'))
+
+            FacilityMatch \
+                .objects \
+                .filter(facility_list_item=facility_list_item) \
+                .update(is_active=False)
+
+            updated_facility_list_item = FacilityListItem \
```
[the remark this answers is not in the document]
kellyi: :+1: Replaced in 7322c564c71d8a5220dccf7f49dc0b5cfef77ba5

src/app/src/components/FacilityListItemsTableRow.jsx
```diff
@@ -1,27 +1,43 @@
...
+import constant from 'lodash/constant';
+import stubTrue from 'lodash/stubTrue';
 
 import { facilityListItemStatusPropType } from '../util/propTypes';
 
 import { listTableCellStyles } from '../util/styles';
 
 import { makeFacilityDetailLink } from '../util/util';
 
-const propsAreEqual = (prevProps, nextProps) =>
```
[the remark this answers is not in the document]
kellyi: I'll add this when squashing the fixups. We dropped this because we're no longer using React.memo on this component.

kellyi:
Oof, thanks for catching the missing filter on the facilities search endpoint. I added it in 151f6a416f44c7f84d9c9598d0381d4e1a4fc520 which will suppress matches with `is_active=False` from appearing in the contributors & contributor types results.

kellyi:
This is ready for another look!

kellyi:
Rebased this on `develop`

kellyi:
Awesome! Thanks for your help with this!

Issue #600, opened 2019-06-13
Add an API to merge 2 facilities


jwalgran:
## Overview

Depends on #604 "Opt FacilityMatch into django-simple-history tracking"
Depends on #605 "Create a FacilityAlias model"

Add a private API to merge 2 facilities

### Is your feature request related to a problem? Please describe.

If the dedupe process misses a match we need to merge the facilities together.

### Describe the solution you'd like

- [ ] Add a 'MERGED' status to `FacilityMatch`
- [ ] Create a private `merge` API that accepts 2 OAR IDs, a primary and a secondary
  - [x] Verify that a superuser is making the request.
  - [x] Verify that two unique OAR IDs were specified.
  - [ ] Update all the `FacilityMatch` rows of the secondary to point to the primary and set status to `MERGED`.
    - [ ] Set the `changeReason` to "Merged with ${primary OAR ID}" https://django-simple-history.readthedocs.io/en/2.7.2/historical_model.html?#change-reason
  - [ ] Add a row to `FacilityAlias` so that the secondary OAR ID points to the primary
  - [ ] Update the secondary `FacilityListItem.facility` to point to the primary `Facility`. Add an object to the `processing_results` array to document the merge.
  - [ ] Delete the secondary `Facility`. Set the `changeReason` to "Merged with ${primary OAR ID}" https://django-simple-history.readthedocs.io/en/2.7.2/historical_model.html?#change-reason

PR #620, opened 2019-06-20
Add facility merge API

jwalgran:
## Overview

Add facility merge API

Connects #600

## Demo

Screen Shot 2019-06-21 at 11 40 26 AM

## Notes

We merge a facility by:
- Updating all the `FacilityMatch` records and their related `FacilityListItem`s to point to the the target `Facility`
- Changing the `FacilityMatch status to `MERGED` and adding a processing result item to the `FacilityListItem` with a `MERGE_FACILITY` action
- Creating a `FacilityAlias` with a type of `MERGE` pointing the merged OAR ID to the target facility
- Deleting the merged facility record.

We needed to update the filters used by the `Facility` serialization process so that `MERGED` facilities would show up properly in on the facility details page.

## Testing Instructions

This assumes fixture data is loaded.

* Verify that the merge API does not appear on http://localhost:8081/api/docs/#/facilities
* Browse the map and copy 2 OAR IDs from different contributors.
* Call the merge API
```
curl --request POST \
  --url 'http://localhost:6543/api/facilities/merge/?target={{TARGET OAR ID}}&merge={{MERGE OAR ID}}' \
  --header 'authorization: Token {{TOKEN}}'
```
* Verify that `http://localhost:6543/facilities/{{MERGE OAR ID}}` is not found.
* Verify that `http://localhost:6543/facilities/{{TARGET OAR ID}}` shows the details of the merged facility.

## Checklist

- [x] CI passes after rebase
- [x] ~CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines~ SKIPPED because this is a non-public API and the CHANGELOG item for the merge UI covers the announcement of the new feature.

### DIFF
--- a/src/django/api/views.py
+++ b/src/django/api/views.py
@@ -425,6 +425,8 @@ class FacilitiesAutoSchema(AutoSchema):
     def get_link(self, path, method, base_url):
         if method == 'DELETE':
             return None
+        if 'merge' in path:
+            return None
 
         return super(FacilitiesAutoSchema, self).get_link(
             path, method, base_url)
@@ -849,6 +851,15 @@ class FacilitiesViewSet(mixins.ListModelMixin,
             raise NotFound(
                 'The current User does not have an associated Contributor')
         return Response(FacilityClaimSerializer(claims, many=True).data)
+
+    @action(detail=False, methods=['POST'],
+            permission_classes=(IsRegisteredAndConfirmed,))
+    @transaction.atomic
+    def merge(self, request):
+        if not request.user.is_superuser:
+            raise PermissionDenied()
+
+        return Response(status=status.HTTP_501_NOT_IMPLEMENTED)
 
 
 class FacilityListViewSetSchema(AutoSchema):
--- a/src/django/api/views.py
+++ b/src/django/api/views.py
@@ -44,6 +44,7 @@ from api.constants import (CsvHeaderField,
                            FacilitiesQueryParams,
                            FacilityListQueryParams,
                            FacilityListItemsQueryParams,
+                           FacilityMergeQueryParams,
                            ProcessingAction)
 from api.models import (FacilityList,
                         FacilityListItem,
@@ -66,7 +67,8 @@ from api.serializers import (FacilityListSerializer,
                              UserProfileSerializer,
                              FacilityClaimSerializer,
                              FacilityClaimDetailsSerializer,
-                             ApprovedFacilityClaimSerializer)
+                             ApprovedFacilityClaimSerializer,
+                             FacilityMergeQueryParamsSerializer)
 from api.countries import COUNTRY_CHOICES
 from api.aws_batch import submit_jobs
 from api.permissions import IsRegisteredAndConfirmed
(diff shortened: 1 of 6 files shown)


src/django/api/views.py
```diff
@@ -425,6 +427,8 @@ class FacilitiesAutoSchema(AutoSchema):
     def get_link(self, path, method, base_url):
         if method == 'DELETE':
             return None
+        if path[-6:].lower() == 'merge/':
```
kellyi: Out of curiosity, would `if 'merge' in path` work? I used `if 'claim' in path` for suppressing the claim route, but I don't know whether there are cases when the path would be not in lowercase or if 'claim' (or 'merge') could appear validly in the path string prior to the end.

jwalgran: `if 'merge' in path` is probably sufficient.

I added call to `lower()` instinctually because we are string comparing URLs which are case insensitive. However, because the value of `path` is not coming from a user, it will probably never be anything other than lowercase.

It's not likely that we will ever have another route under `/api/facilities` with `merge` in the path, but anchoring the match to the end of the path is a bit more explicit. That said, this would also match `/api/facilities/undo-merge`, so it is not actually as protective as it should be.

I'll switch to the simpler `if 'merge' in path` for consistency.

kellyi: That makes sense -- happy to use whichever idiom is best for suppressing `claim` also.

jwalgran: Updated in FIXUP 9aedbdc

kellyi:
Taking a look at this!

kellyi:
Ran through the testing instructions and the merge was successful! I'm going to rebase the merge UI PR on this and test it out there a bit more.

kellyi:
I rebased the UI PR on this and made a commit to use the new endpoint. It's working pretty well! I did encounter this bug after:

- Merged a facility into facility A
- Merged another facility into facility A
- Tried to merge facility A into fourth facility:

```
("Cannot delete some instances of model 'Facility' because they are referenced through a protected foreign key: 'FacilityAlias.facility'", <QuerySet [<FacilityAlias: US20191728AME0J -> Anhui New L'idole Group Co. (CN2019172BQ3KRQ)>, <FacilityAlias: CN20191723MACN9 -> Anhui New L'idole Group Co. (CN2019172BQ3KRQ)>]>)
```

kellyi:
(Potentially we could catch and disallow that operation & send back a 400 but it may be a valid thing to allow.)

jwalgran:
We should be able to merge anything into anything. I'll work on resolving that error.

jwalgran:
Merging can result in the same contributor appearing twice in the facility details. I am going to make sure this list has unique entries.

Screen Shot 2019-06-21 at 12 25 51 PM

kellyi:
I also encountered a 500 when trying to delete a facility which I had merged some facilities into:

```
("Cannot delete some instances of model 'Facility' because they are referenced through a protected foreign key: 'FacilityAlias.facility'", <QuerySet [<FacilityAlias: CN2019172JYZ8KH -> BMD Printing Vina Co. Ltd. (VN2019172ZAG0WF)>, <FacilityAlias: KH20191726D9MXP -> BMD Printing Vina Co. Ltd. (VN2019172ZAG0WF)>, <FacilityAlias: BR20191721J88K0 -> BMD Printing Vina Co. Ltd. (VN2019172ZAG0WF)>]>)
```

jwalgran:
These failures should be easy to fix. Just need to rewrite the aliases.

jwalgran:
I rebased this to include the alias handling fixes included in `develop` and pushed a fixup to properly handle aliases in the merge API. This is ready for another look.

kellyi:
I tested this out with the UI changes and everything is working as expected. We did note that deleting a merged facility causes `HistoricalFacilityAlias.objects.all` to break, but other ORM operations work and the Django admin continues to work afterward.

kellyi:
Friendly reminder to squash before merging!

jwalgran:
Thanks for the thorough review and for catching a very important bug in this one.

Issue #613, opened 2019-06-17
Facility claim API does not have API documentation


jwalgran:
## Overview

The facility claim API does not have API documentation.

### Expected Behavior

The API documentation for the claim endpoint includes a description and a sample request

### Actual Behavior

The default text from the FacilitiesViewSet is shown

### Steps to Reproduce

* Browse http://localhost:8081/api/docs/#!/facilities/facilities_claim

### Demo

<img width="1174" alt="Screen Shot 2019-06-17 at 3 07 25 PM" src="https://user-images.githubusercontent.com/17363/59640003-ab8b1080-9111-11e9-9b52-13b6a7ab50e1.png">

kellyi:
Alternatively: we could also just remove the claimed endpoints from the API docs.

kellyi:
Shares points with https://github.com/open-apparel-registry/open-apparel-registry/issues/569

Issue #606, opened 2019-06-17
Contributor names and type in the filter dropdowns are being translated


jwalgran:
## Overview

Contributor names and types in the filter dropdowns are being translated. The translate library tries to translate all text on a page unless it encounters a a `notranslate` class.

### Expected Behavior

Contributor names and types are not translated in the map page filter dropdown.

### Actual Behavior

Contributor names and types are translated in the map page filter dropdown.

### Steps to Reproduce

* Browse the map page
* Choose a non-English language
* Open the contributor filter dropdown
* Open the contributor type dropdown

### Demo

<img width="528" alt="Screen Shot 2019-06-17 at 7 43 16 AM" src="https://user-images.githubusercontent.com/17363/59614607-f5560580-90d5-11e9-9e10-29dacf585c1a.png">

<img width="237" alt="Screen Shot 2019-06-17 at 7 57 24 AM" src="https://user-images.githubusercontent.com/17363/59614701-17e81e80-90d6-11e9-8464-0812f52c9789.png">

### Additional context

It looks like setting the `notranslate` class on parent element applies to all child elements, so this should be a relatively simple fix.

kellyi:
Shares points with https://github.com/open-apparel-registry/open-apparel-registry/issues/569

Issue #569, opened 2019-06-07
In app back buttons do not route correctly after claiming a facility


jwalgran:
## Overview

The in app back buttons return to the claim a facility form rather than navigating back to facility search results.

### Expected Behavior

After searching for a facility, selecting it, and successfully running through the claim forms, clicking the two in-app back buttons should return to the facility search results.

### Actual Behavior
After searching for a facility, selecting it, and successfully running through the claim forms, clicking the two in-app back buttons returns to the claim form.

### Steps to Reproduce

* Browse http://localhost:6543/
* Search
* Select a facility
* Choose "Claim this facility"
* Successfully complete the claim.
* Click the in-app <- button on the "successful" claim screen. You are returned to the facility details page.
* Click the in-app <- button on the facility details page

### Demo

![2019-06-07 14 50 19](https://user-images.githubusercontent.com/17363/59135560-462a6900-8934-11e9-89c7-d3b6eeab4860.gif)

Issue #565, opened 2019-06-07
The profile page for a non-existant contributor ID should say not found


jwalgran:
## Overview

The profile page for a non-existent contributor ID shows empty fields rather than not found

### Expected Behavior

Loading the profile page for non-existent contributor ID shows the standard not found page

### Actual Behavior

A few empty fields are shown

### Steps to Reproduce

* Browse http://localhost:6543/profile/987654321

### Demo

<img width="1056" alt="Screen Shot 2019-06-07 at 8 29 34 AM" src="https://user-images.githubusercontent.com/17363/59115623-840c9a80-88fe-11e9-89bc-78809d05c86d.png">

kellyi:
Shares points with https://github.com/open-apparel-registry/open-apparel-registry/issues/569

PR #622, opened 2019-06-21
Make miscellaneous bugfixes

kellyi:
## Overview

- show RouteNotFound for non-existent profiles
- don't translate contributor name and type in search menus
- suppress `claim/` route from facilities API documentation
- adjust facility details back button to go back to the main facilities/ route
- update contributor name hint text on registration form

Connects #565 
Connects #606
Connects #613 
Connects #569 
Connects #120 

## Testing Instructions

- serve this branch
- visit http://localhost:6543/profile/70000 and verify that you see the route not found component
- visit http://localhost:6543 and change the language to Esperanto and verify that the contributor names and contributor types sections aren't translated
- sign in as c1@example.com then visit localhost:8081/admin and turn on the claim a facility switch
- visit http://localhost:8081/api/docs/#/facilities and verify that you do not see a POST endpoint for `claim/`
- sign out then sign in as c2@example, search for a facility, and claim it
- fill out the claim form completely; when you reach the last page, click the in-app back button on the claim form
- when you return to the facility details page, click that in-app back button and verify that it takes you to the main facilities search

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Removed
 
 ### Fixed
+- Miscellaneous bugfixes [#622](https://github.com/open-apparel-registry/open-apparel-registry/pull/622)
 
 ### Security
 
--- a/src/app/src/App.css
+++ b/src/app/src/App.css
@@ -136,7 +136,7 @@ li.dropdown-list-item button {
 .form__hint {
   margin: 8px 0;
   font-weight: 400;
-  font-size: 12px;
+  font-size: 15px;
 }
 
 .form__select-input-container {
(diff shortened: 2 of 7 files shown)


kellyi:
I also adjusted the "Contributor Name" hint on the registration form with the suggested text from #120:

Screen Shot 2019-06-21 at 2 12 39 PM

kellyi:
Looking at that again I realized I placed the new text on the wrong hint, so I'll adjust.

kellyi:
Alright:

Screen Shot 2019-06-21 at 2 20 16 PM

jwalgran:
Looking at this now.

jwalgran:
The `facility-claims` API endpoints are also undocumented. Do you think that we should add an issue to document them, or remove that viewset from the docs as well?

kellyi:
> The facility-claims API endpoints are also undocumented. Do you think that we should add an issue to document them, or remove that viewset from the docs as well?

I think there are a couple that are administrator-only that we probably want to remove regardless, so my inclination is just to remove all of them for now and if we decide to restore some of them later we can document them then?

kellyi:
eb1af151326160b0b4c79abd3ac21099d6031fb4 removes the FacilityClaimsViewset from the API docs.

kellyi:
Thanks!

Issue #618, opened 2019-06-19
Add a parent company to claim and allow it to be edited on the profile


jwalgran:
## Overview

Add a "Parent Company" field to both the claim form and the profile form where the claimant can select a single, existing `Contributor` from a select box. Show the name of the contributor on the facility details page as a link to a search of all of the facilities for that Contributor.

### Describe the solution you'd like

- Add a nullable `parent_company` to the claim as a foreign key to `Contributor`. PROTECT on delete.
- Add a non-required choice field to the claim form to allow selecting a `Contributor`.
- Ad a non-required choice field to the profile form that allows editing the `Contributor`
- If a `parent_company` is specified, show a link on the facility details page. Clicking the link runs a search for all of that contributors facilities.

### Additional Context
<img width="513" alt="Screen Shot 2019-06-21 at 8 31 00 AM" src="https://user-images.githubusercontent.com/17363/59933944-3d8b6580-93ff-11e9-9229-43c1103bd033.png">

kellyi:
For this we're going to have to move the place where we call the fetch contributors options function. It currently is only called on the `/` and `/facilities/` routes, but not the facilities detail route or the claims route. I think we can just move it up to the top level of the app and have it live there.

PR #626, opened 2019-06-21
Add parent company to facility claim

kellyi:
## Overview

This PR adds a parent_company field to the FacilityClaim model as a FK to Contributor, and also enables selecting, displaying, and updating that field from the list of contributorOptions.

Connects #618 

## Demo
Screen Shot 2019-06-22 at 10 44 55 AM
Screen Shot 2019-06-22 at 10 48 10 AM
Screen Shot 2019-06-22 at 10 47 23 AM
Screen Shot 2019-06-22 at 10 45 08 AM

## Notes

Adding the new field makes the number of items displayed in the dashboard claim details page uneven, which breaks using having the two text area fields sit adjacent at the bottom of the list. We can probably wait to fix this when and if it is an actual issue.

## Testing Instructions

- get this branch, then `resetdb` and `processfixtures` and `server`
- turn on claim a facility
- visit 6543 and verify that the sidebar select dropdown filters still work as they did before
- sign in as c2 and claim a facility. Verify that you can select a parent company from the select menu
- sign out and sign back in as c1, then visit /dashboard/claims and click on the facility claim
- verify that you see the parent company listed there and that it is linked to the correct profile
- approve the claim, then sign out, and sign back in as c1
- visit "My Facilities" and verify that you see the "Parent Company" field on the form and that you can change and save the "Parent Company" value
- visit the facility details page for the facility you've claimed and verify that you see the parent company linked in the claim info

## Checklist

- [ ] CI passes after rebase


kellyi:
> contributors without lists will not be selectable in the UI. If this is problematic in some way, I can create another private endpoint for the claim a facility stuff.

Turns out we do want contributors without lists to be selectable so I'll make the necessary adjustments here.

jwalgran:
Looking at this now.

src/app/src/util/constants.js
```diff
@@ -434,6 +434,10 @@ export const claimAFacilityFormFields = Object.freeze({
         id: 'company-name',
         label: 'Official name of LLC or company registered',
     }),
+    parentCompany: Object.freeze({
+        id: 'parent-company',
+        label: 'Parent company of facility, selected from contributors list',
```
jwalgran: Consider simplifying this label to "Parent company" and adding an `aside`. This was a bit of early feedback when reviewing the feature on a previous checkin call. Something like:

>If you cannot find the parent company in this list consider inviting them to register with the Open Apparel Registry.

Screen Shot 2019-06-24 at 1 58 47 PM

kellyi: That makes sense. I adjusted this on both the initial form and the update page in 364ce9d79f00f30b38865c63e3f3f3fc78df982d


Screen Shot 2019-06-24 at 5 39 11 PM
Screen Shot 2019-06-24 at 5 47 36 PM

src/django/api/serializers.py
```diff
@@ -427,13 +427,22 @@ def get_claim_info(self, facility):
                 .filter(status=FacilityClaim.APPROVED) \
                 .get(facility=facility)
 
+            if claim.parent_company:
+                parent_company = {
+                    'id': claim.parent_company.id,
```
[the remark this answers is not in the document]
kellyi: Thanks for catching that -- I remember being slightly confused by this in instances when the contributor is and the contributor.admin id are the same, so it doesn't show the discrepancy. I adjusted this in efee42f.

kellyi:
Rebased this on `develop`.

kellyi:
Rebased on `develop` again & also adjusted the counter stamp on the migration.

kellyi:
Thanks! Going to merge and then I'll run the migration on staging.

Issue #593, opened 2019-06-13
Allow an incorrect match to be split


jwalgran:
## Overview

- Opt the `FacilityMatch` into django-simple-history tracking.
- Create a private split API that accepts a facility list item id
  - Verify that the list item is in a matched status
  - Verify that a facility was not already created from the list item.
  - Create a new `Facility` for the list item
  - Update the `FacilityMatch` row to point to the newly created facility.
- Create a `/dashboard/split` page 
  - Submit an OAR ID. 
  - Show the name/address/country of the facility matching the OAR id
  - Show the name/address/country of each list item matched to it.
      - Show a "Create Separate Facility" button next to each. On click:
        - call the split API
        - On success, disable the button and show the OAR ID of the new facility as a link to the new details page

PR #633, opened 2019-06-25
Enable administrators to split facility matches into new facilities

kellyi:
## Overview

This PR enables administrators to split facility matches into new facilities via a new `/dashboard/splitfacilities` route. The new page has the same facility card as is displayed on the merge and delete UIs, but also displays a "Matches to Split" card on the righthand side which displays details about each match along with a "Split" button used to open a confirmation dialog before creating a new facility. On successful splits, we update the UI to display the new OAR ID and also to filter the match out of the source facility's "Contributors" list.

Connects #593 

## Demo

Screen Shot 2019-06-25 at 7 33 40 PM

## Notes

~Although this might be an edge case I noticed a bug in what I think should be a valid operation: after splitting a match from Facility A into Facility B, I attempted to merge Facility A into Facility B and encountered this error:~

```
("Cannot delete some instances of model 'Facility' because they are referenced through a protected foreign key: 'FacilityListItem.facility'", <QuerySet [<FacilityListItem: FacilityListItem 594 - MATCHED>, <FacilityListItem: FacilityListItem 277 - MATCHED>]>)
```

~Merging B back into A worked as expected, so I'm not quite sure what causes the error above. Any ideas?~

I believe I fixed this and that the problem was that I had neglected to update the facility_list_item.facility to the new facility.
## Testing Instructions

- serve this branch then `resetdb` & `processfixtures`
- sign in as c1 then search for facility name "Regina" in country "China" and copy the OAR ID
- visit the dashboard page, then click on the "Split Facility Matches" link
- search for the OAR you copied
- split the matches and verify that they create a new facility
- you can also try out merging the new facilities back into the original facility, or merging facilities then splitting them then merging them etc

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/views.py
+++ b/src/django/api/views.py
@@ -985,6 +985,8 @@ class FacilitiesViewSet(mixins.ListModelMixin,
                 .objects \
                 .get(pk=match_id)
 
+            old_facility_id = match_for_new_facility.facility.id
+
             list_item_for_match = match_for_new_facility.facility_list_item
 
             new_facility = Facility \
@@ -1005,7 +1007,17 @@ class FacilitiesViewSet(mixins.ListModelMixin,
 
             match_for_new_facility.save()
 
+            now = str(datetime.utcnow())
+
             list_item_for_match.facility = new_facility
+            list_item_for_match.processing_results.append({
+                'action': ProcessingAction.SPLIT_FACILITY,
+                'started_at': now,
+                'error': False,
+                'finished_at': now,
+                'previous_facility_oar_id': old_facility_id,
+            })
+
             list_item_for_match.save()
 
             return Response({
--- a/src/django/api/views.py
+++ b/src/django/api/views.py
@@ -941,38 +941,81 @@ class FacilitiesViewSet(mixins.ListModelMixin,
         response_data = FacilityDetailsSerializer(target).data
         return Response(response_data)
 
-    @action(detail=True, methods=['GET'],
+    @action(detail=True, methods=['GET', 'POST'],
             permission_classes=(IsRegisteredAndConfirmed,))
     @transaction.atomic
     def split(self, request, pk=None):
         if not request.user.is_superuser:
-            return NotFound()
+            raise PermissionDenied()
 
         try:
-            facility = Facility.objects.get(pk=pk)
+            if request.method == 'GET':
+                facility = Facility.objects.get(pk=pk)
 
-            facility_data = FacilityDetailsSerializer(facility).data
+                facility_data = FacilityDetailsSerializer(facility).data
 
-            facility_data['properties']['matches'] = [
-                {
-                    'name': m.facility_list_item.name,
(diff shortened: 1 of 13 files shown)


jwalgran:
Looking at this now.

src/django/api/views.py
```diff
@@ -941,38 +941,81 @@ def merge(self, request):
...
+            match_for_new_facility.status = FacilityMatch.CONFIRMED
+            match_for_new_facility.results = {
+                'match_type': 'split_by_administator',
+                'split_from_oar_id': match_for_new_facility.facility.id,
+            }
+
+            match_for_new_facility.save()
+
+            list_item_for_match.facility = new_facility
+            list_item_for_match.save()
```
jwalgran: The delete and merge APIs add an item to `processing_results`, which acts as an audit trail of all the mutations that have been applied to the line item. Consider continuing that pattern here following the example of merge:

https://github.com/open-apparel-registry/open-apparel-registry/blob/95395af3c6d2a30dedc350902fe27ac5453e82cc/src/django/api/views.py#L909-L915

We would want to add a new `ProcessingAction.SPLIT_MATCH` constant.

kellyi: Sounds good! I added this in fc74dbe and the entry now shows "split_match" as the action and also indicates the "previous_facility_oar_id":

```
>>> f = Facility.objects.get(pk='CN2019178FQ1RKZ')
>>> f.c
f.check(           f.clean(           f.clean_fields(    f.contributors(    f.country_code     f.created_at       f.created_from     f.created_from_id  
>>> f.created_from
<FacilityListItem: FacilityListItem 594 - MATCHED>
>>> f.created_from.processing_results
[{'error': False, 'action': 'parse', 'started_at': '2019-06-26 18:54:46.654964', 'finished_at': '2019-06-26 18:54:46.659191', 'is_geocoded': True}, {'error': False, 'action': 'geocode', 'started_at': '2019-06-26 18:54:51.556838', 'finished_at': '2019-06-26 18:54:51.556877', 'skipped_geocoder': True}, {'error': False, 'action': 'match', 'started_at': '2019-06-26 18:54:52.275275', 'finished_at': '2019-06-26 18:55:10.561837'}, {'error': False, 'action': 'merge_facility', 'started_at': '2019-06-26 18:59:01.103755', 'finished_at': '2019-06-26 18:59:01.103755', 'merged_oar_id': 'CN2019177A7GXDQ'}, {'error': False, 'action': 'merge_facility', 'started_at': '2019-06-26 19:25:29.522594', 'finished_at': '2019-06-26 19:25:29.522594', 'merged_oar_id': 'CN2019177ZT0K3V'}, {'error': False, 'action': 'split_facility', 'started_at': '2019-06-27 01:08:15.123850', 'finished_at': '2019-06-27 01:08:15.123850', 'previous_facility_oar_id': 'CN2019177D9KY2N'}]
```

kellyi:
Thanks!

Issue #637, opened 2019-06-26
Log all non-app API requests in the database by  user ID


jwalgran:
## Overview

Depends on #636

To allow reporting on API usage, log all requests made with an API token to the database.

### Describe the solution you'd like
- Create a `RequestLog` table
  - ~`requested_at`~ `created_at` timestamp (opted to reuse our conventional timestamp field name)
  - `user_id` FK to api_user, indexed, cascade delete
  - `token` indexed ~(both alone and compound index with user_id)~ (Opting to not add speculatively add a multicolumn index) NOT an FK because tokens can be rotated.
  - `url` string
  - `response_code` int
- Create middleware (https://docs.djangoproject.com/en/2.2/topics/http/middleware/#writing-your-own-middleware) that writes a row to the table if an API request is made with a token

hectcastro:
Do we have a sense for how many requests this will be? Surfacing minor concern over the possibility of:

- Adding functionality that could lead to a large database table (over time)
- Introduces a step in the request/response cycle that has the potential to compromise the success of a request if this particular step fails

jwalgran:
Thanks for raising these concerns. 

We don't know just how successful our APIs will be, but we are actively engaged with multi-supplier initiatives to build integrations with their systems. I would guess the order of magnitude would be thousands per week.

We are obligated to report on API usage to our stakeholders, so we need to keep a count somewhere. I don't think we can avoid introducing a step in the request/response cycle, but I would definitely want to implement it so that a failure to count does not fail the request. 

I'm open to using an alternate data store. We don't need real time access to these counts so that could give us some flexibility.

One advantage to using the database is that we can maintain a foreign key to the user table and cascade delete when and if a user exercises their "right to be forgotten" under the GDPR.

hectcastro:
Thanks for the follow-up. Not super interested in pursuing an alternate data store at this time. GDPR makes it difficult to use the existing logging structure, which would have been great to reuse. I guess we just need to keep an eye on the size of this table over time.

PR #646, opened 2019-07-02
Log requests made with token authentication

jwalgran:
## Overview

Log all token-authenticated requests to the database for reporting purposes.

Connects #637

## Notes

This PR also includes a migration mistakenly omitted from a previous PR (5a66f74).

We do not want a failure in the request logging to disrupt the handling of a response to we catch all exceptions and send them to Rollbar, failing silently if the call to Rollbar, in turn, raises an exception.

## Testing Instructions

This assumes you have fixtures loaded and processed.

* Run `./scripts/manage dbshell`
* Log in to the web app as `c2@example.com`. 
* Browse http://localhost:6543/lists
* Run `select * from api_requestlog;` in `dbshell`. Verify that the session authenticated web app request for the facility lists was _not_ logged.
*  Make an authenticated request with the token and verify that it succeeds.
```
curl \
  -H "Authorization: Token {{API-TOKEN}}" \
  http://localhost:6543/api/facility-lists/
```
* Run `select * from api_requestlog;` in `dbshell`. Verify that the token authenticated web app request for the facility lists _was_ logged.
*  Make an authenticated request for a resource that returns 404.
```
curl \
  -H "Authorization: Token {{API-TOKEN}}" \
  http://localhost:6543/api/facility-lists/3/
```
* Run `select * from api_requestlog;` in `dbshell`. Verify that the 404 request was also logged.

## Checklist

- [x] CI passes after rebase


src/django/api/middleware.py
```diff
@@ -0,0 +1,45 @@
...
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+
+        auth = None
+        try:
+            if request.user and request.user.is_authenticated:
+                auth = get_authorization_header(request)
+                if auth and auth.split()[0].lower() == 'token'.encode():
```
jwalgran: This code is based on
https://github.com/encode/django-rest-framework/blob/e4e75f1c7ceb053bffb8031e0468000c4d9c3cec/rest_framework/authentication.py#L170-L189

src/django/api/tests.py
```diff
@@ -2881,3 +2882,36 @@ def check_host(url):
...
+        self.user.set_password(self.password)
+        self.user.save()
+
+        Contributor.objects.create(name=self.name, admin=self.user)
+
+    def test_request_without_token_is_not_logged(self):
+        self.client.login(email=self.email, password=self.password)
+        response = self.client.get(reverse('facility-list-list'))
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(0, RequestLog.objects.all().count())
```
rajadain: This implicitly relies on the assumption that the test is operating on a fresh database. While this will mostly be true, to be a more specific test consider something like:

```python
self.assertEqual(0, RequestLog.objects.filter(user=self.user).count())
```

src/django/api/tests.py
```diff
@@ -2881,3 +2882,36 @@ def check_host(url):
...
+
+    def test_request_with_token_is_logged(self):
+        token = Token.objects.create(user=self.user)
+        path = reverse('facility-list-list')
+        response = self.client.get(
+            path,
+            HTTP_AUTHORIZATION='Token {0}'.format(token))
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(1, RequestLog.objects.all().count())
```
rajadain: Same here.

src/django/api/models.py
```diff
@@ -825,3 +825,32 @@ class Meta:
...
+        max_length=2083,
+        null=False,
+        help_text='The requested resource path'
+    )
+    response_code = models.IntegerField(
+        null=True,
+        help_text='The HTTP status code returned to the requester'
+    )
+    created_at = models.DateTimeField(auto_now_add=True, db_index=True)
+    updated_at = models.DateTimeField(auto_now=True)
```
rajadain: Can't imagine this field being used very much in a table like this, but I suppose it doesn't hurt to have it for consistency.

jwalgran: In addition to consistency, it is also good for retroactively identifying bad app behavior. If `updated_at` does not match `created_at` something that should not have happened did.

jwalgran:
Adding a `method` column is a great idea.

src/django/api/tests.py
jwalgran: Changed in 19a2eb2

src/django/api/tests.py
jwalgran: Changed in 19a2eb2

jwalgran:
I added the `method` field in edd8071 and 19a2eb2

jwalgran:
Thanks for the excellent suggestions.

Issue #587, opened 2019-06-13
Validate claimed facility form website & make website linked in facility details display


kellyi:
## Overview

Validate claimed facility profile form website & make website linked in facility details display

### Describe the solution you'd like

Validate claimed facility profile form website & make website linked in facility details display

PR #647, opened 2019-07-03
Validate claimed facility website field and show as hyperlink

jwalgran:
## Overview

Adds client validation of the facility claim website field, changes the related model field to a URLField, and makes valid URLs clickable on the facility details sidebar.

Connects #587 

### Form validation

Screen Shot 2019-07-02 at 4 56 40 PM

### Clickable link

Screen Shot 2019-07-02 at 4 56 17 PM

## Testing Instructions

Assumes that the fixtures have been loaded and processed.

* Run `./scripts/update`
* Make sure facility claiming is enabled
  * `./scripts/manage waffle_switch claim_a_facility on`
  * Restart `./scripts/server`
* Log in as `c2@example.com`
* Claim a facility
  * On the "Facility Information" step, verify that you can only proceed to the next step if the website field is either empty or a valid url (with or without a scheme)
* Log in as `c1@example.com` and approve the claim via http://localhost:6543/dashboard/claims
* Log back in as `c2@example.com` and open the facility claim details page via http://localhost:6543/claimed
* Verify that you can only enter valid URLs into the website field (protocol is optional)
  * Verify that the save button is disabled until the website field is valid
* Enter a valid URL in the website field and save the form.
* Browse the facility details page and verify that the website is a clickable link that opens the URL in a new tab/window.

## Checklist

- [ ] CI passes after rebase


rajadain:
We should probably add similar validation on this page:

image

which currently allows free text input.

jwalgran:
Thanks you for catching that. I overlooked the fact that there are two separate website fields, one for claim research and one for public display on the facility page. Will fix.

rajadain:
And should we also add similar validation in the profile page:

image

or wait until a later issue for it?

rajadain:
I also noticed that when I had a tab previously open at the search home page, searching for the facility and going to its details page did not update with the claim (bottom), but the details page on a full page refresh did (top):

image

Is that a known issue? If not, I'll make a fresh one for it.

jwalgran:
>should we also add similar validation in the profile page

I think so. No need to make a separate issue.

>searching for the facility and going to its details page did not update with the claim...Is that a known issue? If not, I'll make a fresh one for it.

I don't think this is a known issue and yes please make a fresh one for it.

jwalgran:
I added validation to the claim a facility wizard in c5d77ff 

I changed my my about also including changes to the user/contributor profile page. It is out of scope of this claim a facility specific issue. I created #656 for it.

This is rebased on develop and ready for re-review.

rajadain:
Taking another look now.

rajadain:
When there is an error in the "Official Name" field, in addition to the "Some required fields are missing or invalid" message in the bottom, the field itself has a red highlight:

image

However, when the "Facility Website" field has invalid data, it stays blue, instead of red:

image

Making it red would help identify which field has the invalid data.

kellyi:
re the URL field on the ClaimFacility...Step file, I believe you can just add a line like this one -- which I used to validate that the LinkedIn URL was a real URL -- to have the field be outlined in red when it has an error state:

https://github.com/open-apparel-registry/open-apparel-registry/pull/650/files#diff-cca727c1524411105c81408bc15d1331R85

rajadain:
When viewing a Pending facility claim as an admin, the website is not clickable, even when valid:

image

rajadain:
After a claim is approved, information submitted in the claim (such as website, contact person name, etc) are not pre-filled in the Claimed Facility Details page:

image

Is this expected / is there a separate issue for this?

rajadain:
The red highlight works correctly on the Claimed Facility Details page:

image

kellyi:
> information submitted in the claim (such as website, contact person name, etc) are not pre-filled in the Claimed Facility Details page

This one is intentional. The Claim Form is only for evaluating the claim; there are separate fields for the public profile info. They are separate so that the claim form info remains non-public.

rajadain:
Link works correctly from the UI once saved in the Claimed Facility Details form:

rajadain:
> I also noticed that when I had a tab previously open at the search home page, searching for the facility and going to its details page did not update with the claim (bottom), but the details page on a full page refresh did (top)

I am unable to reproduce this right now, was probably doing something wrong during testing. Going to ignore it for now.

rajadain:
The two things this needs for approval:

- [ ] Make the website textbox be red when the URL is incorrect in the Claim application form
- [ ] Make the website clickable when reviewing a claim as an admin

jwalgran:
Thanks for the thorough review

Issue #621, opened 2019-06-21
Ensure client uses `RouteNotFound` component for all unmatched routes.


kellyi:
## Overview

As described in #560, some react-router `Switch` blocks do not use the `RouteNotFound` component as their default case. As a consequence, some not-declared URLS direct to existing components rather than `RouteNotFound`. For example: `/dashboard/hello/`

We should ensure that the default cases for all react-router `Switch` blocks render the `RouteNotFound` component.

### Desired Behavior

Ensure that the default cases for all react-router `Switch` blocks render the `RouteNotFound` component.

Ensure that all other routes continue to work as before -- and that the API routes also continue to work.

### Additional context

#560 proposes a more complicated system of redirects. However for simplicity's sake we should stick with the default case handling.

Supersedes #560

kellyi:
I ran `rg "<Switch>"` in the OAR dir on my local and these are the places where it occurs:

```
src/app/src/App.jsx
86:                            <Switch>

src/app/src/components/FacilityListItems.jsx
218:                                    <Switch>

src/app/src/components/MapAndSidebar.jsx
39:                        <Switch>
53:                        <Switch>

src/app/src/components/Dashboard.jsx
87:                        <Switch>
135:                <Switch>

src/app/src/components/ClaimedFacilities.jsx
21:                        <Switch>
31:                <Switch>
```

PR #657, opened 2019-07-09
Display RouteNotFound component on unmatched routes

kellyi:
## Overview

Make react-router setup adjustments to ensure we display RouteNotFound component on unmatched routes.

Connects #621 

## Testing Instructions

- serve branch, `./scripts/manage resetdb && ./scripts/manage processfixtures`
- turn on claim a facility
- visit the app on 6543 and navigate through routes as:
   - c1@example
   - c2@example
   - non-signed-in user

...and verify that the routing now:

- displays a component when appropriate match occurs
- displays the RouteNotFound component (which is still not very well-styled) for routes which should not match)

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/App.jsx
+++ b/src/app/src/App.jsx
@@ -86,11 +86,6 @@ class App extends Component {
                             <Switch>
                                 <Route
                                     exact
-                                    path={mainRoute}
-                                    component={MapAndSidebar}
-                                />
-                                <Route
-                                    exact
                                     path={claimFacilityRoute}
                                     render={() => (
                                         <FeatureFlag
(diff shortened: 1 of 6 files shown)


src/app/src/App.jsx
```diff
@@ -85,12 +85,6 @@ class App extends Component {
                         <main style={appStyles.mainPanelStyle}>
                             <Switch>
                                 <Route
-                                    exact
-                                    path={mainRoute}
-                                    component={MapAndSidebar}
-                                />
-                                <Route
-                                    exact
```
rajadain: `claimFacilityRoute` lost the `exact` here. Was this intentional?

kellyi: This may have been an oversight; I'll double check before merging!

kellyi: I believe this was unintentional. Adding it back did not prevent `/facilities/OARID/claim` from loading as usual. I'll restore the `exact` here.

kellyi: (Conversely, without it `/facilities/OARID/claim/helllo` did match.)

Issue #654, opened 2019-07-09
Add Affiliations field to claim a facility profile


kellyi:
## Overview

Add Affiliations field to claim a facility profile

### Describe the solution you'd like

Add Affiliations field to claim a facility profile

This should be a select/options input using React Select, likely a multiselect. ~We're awaiting the list of options to include~, but I believe we should treat it as a closed/fixed set and not allow free-form user input.

Per #617 this should be displayed between "Certifications" and "Office Info" on the facility details screen's claim info section.

kellyi:
This is blocked because we are awaiting the list of options.

dboyer:
These options should be available in a drop-down. Users cannot add other options.

Benefits for Business and Workers (BBW)
Better Mills Programme
Better Work (ILO)
Canopy
Ethical Trading Initiative
European Outdoor Group
Fair Labor Association
Fair Wear Foundation
SEDEX
Social and Labor Convergence Plan (SLCP)
Sustainable Apparel Coalition
Sweatfree Purchasing Consortium

dboyer:
@kellyi The field name should be changed to Affiliations

Issue #653, opened 2019-07-09
Add Certifications to claimed facility profile


kellyi:
## Overview

Add Certifications to claimed facility profile

### Describe the solution you'd like

Add Certifications to claimed facility profile. This should be a select/option input using React-Select with autocomplete. I believe it should also be a multiselect.

~We're waiting on the enumerated set of options, and~ I think we should treat it as a closed set for now.

Per #617 this should be displayed between "Percentage of Female Workers" and "Membership Bodies" in the facility details screen's claim info section.

kellyi:
This is blocked because we are still awaiting the list of options.

dboyer:
The following options should be listed in a drop-down. Users cannot add new options.

BCI
B Corp
Bluesign
Canopy
Cradle to Cradle
EU Ecolabel
FSC
Global Recycling Standard (GRS)
GOTS
Green Screen for Safer Chemicals
Higg Index
IMO Control
International Wool Textile Organisation (IWTO)
ISO 9000
IVN leather
Leather Working Group
Nordic Swan
Oeko-Tex Standard 100
Oeko-Tex STeP
Oeko-Tex Made in Green
Oeko-Tex Eco Passport
PEFC
REACH
Responsible Down Standard (RDS)
Responsible Wool Standard (RWS)
SA8000
Soil Association Organic
Tessile e Salute
USDA Biobased
Woolmark
Worldwide Responsible Accredited Production (WRAP)
Zero Discharge of Hazardous Chemicals (ZDHC) Gateway

dboyer:
@kellyi The field name should be changed to Certifications/Standards/Regulations

PR #671, opened 2019-07-12
Add affiliations and certifications fields to facility claim profiles

kellyi:
## Overview

Add affiliations and certifications fields to facility claim profiles. The fields are ArrayFields in Django and are chosen using react-select multiselects in the UI. In the facility details page we display the selections as lists.

Connects #654 
Connects #653 

## Demo

Screen Shot 2019-07-12 at 1 51 15 PM
Screen Shot 2019-07-12 at 1 51 19 PM

## Testing Instructions

- serve branch, then `resetdb` && `processfixtures`
- claim a facility and then approve the claim
- on the claim profile form, select some affiliations and certifications options, then save
- verify that the save is persisted and that the affiliations and certs are visible in the UI
- verify that the affiliation and certification changes appear in the update email

## Checklist

- [ ] CI passes after rebase


jwalgran:
Looking at this now.

src/app/src/components/FacilityDetailSidebarClaimedInfo.jsx
```diff
@@ -156,6 +158,44 @@ export default function FacilityDetailsSidebarClaimedInfo({
...
+                }
+            />
+            <ClaimInfoSection
+                label="Certifications/Standards/Regulations"
+                value={
+                    facility.certifications && facility.certifications.length
+                        && (
+                            <ul>
+                                {
+                                    orderBy(facility.certifications, identity)
```
jwalgran: We are using `orderBy` both here and in `ClaimedFacilityDetailsReducer.js`. Consider removing the sort from one or the other.

jwalgran:
Wait

kellyi:
Oh, good catch!

kellyi:
I fixed the array field serialization in 
d2884c4

src/app/src/components/FacilityDetailSidebarClaimedInfo.jsx
kellyi: The orderBy here is to ensure that the Facility Details screen display is in alpha order; the reducer version is to ensure that the list sent to the API on profile updates is alphabetical. In general the API doesn't really bother to alphabetize things on its own, where the lists are effectively just sets -- but it will send an update email if the list has the same elements but in a different order, which is why I added the guard on the reducer.

I think I may leave the orderBy in both places in the client just for some display guarantees. If it becomes problematic we can move the alphabetization into the API.

jwalgran: 👍

kellyi:
> My initial serialization suggestion had a bug. We need to handle None.

Sounds good -- this is 7d569532640dfeba28359fa15b542965cb9375fe

kellyi:
Fixed the lint and squashed the fixups! This is ready to go if it's good to go in.

jwalgran:
👍

kellyi:
Thanks for your help with this! Going to merge & will then run the migration on staging.

Issue #652, opened 2019-07-09
Add Production Type field to claim a facility profile


kellyi:
## Overview

Add Production Type field to claim a facility profile

### Describe the solution you'd like

Add Production Type field to claim a facility profile.

We'll need to clarify how this should work: will it be a select/option input from which the user chooses one of a pre-existing set of options or will it be a user-generated set of tags with pre-seeded values.

Per #617 the field should appear between "Product Type" and "Minimum Order Quantity" on the facility details screen's claimed section.

dboyer:
The field should function the same as Product Type with a pre-seeded list of options to choose, the ability to create new options, and an auto-complete feature where we suggest previously chosen options when the user begins typing.

dboyer:
Seeded list of options. Users can add others.

Cutting 
Denim Services 
Embroidery 
Fabric Dyeing 
Fabric Sourcing 
Fabric Weaving
Ginning 
Hand Sewing 
Knitting 
Laundry
Marker Making 
Packing
Pattern Grading 
Pattern Making 
Pleating 
Printing 
Sample Making 
Scouring
Sewing 
Shoe stitching
Spinning
Tanning
Trimming/Embellishment

PR #680, opened 2019-07-16
Add product and production type to claimed facility profile

jwalgran:
## Overview

Brief description of what this PR does, and why it is needed.

Connects #652 

## Demo

## Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description
* Start after checking out this branch
* Include any setup required, such as bundling scripts, restarting services, etc.
* Include test case, and expected output

## Checklist

- [x] CI passes after rebase


src/app/src/components/ClaimedFacilitiesDetails.jsx
```diff
@@ -177,9 +178,39 @@ const InputSection = ({
...
+
+            if (isCreatable && isMultiSelect) {
+                const [
+                    optionsInExistingSet,
+                    newlyCreatedOptionsPrime,
+                ] = partition(
+                    value,
+                    v => includes(map(selectOptions, 'value'), v),
+                );
+                window.console.dir(value);
```
kellyi: Oops, I missed this one. Will push another fixup.

src/app/src/components/ClaimedFacilitiesDetails.jsx
```diff
@@ -173,9 +178,39 @@ const InputSection = ({
...
+
+                const mapStringToOption = s => ({ value: s, label: s });
+
+                return concat(
+                    map(optionsInExistingSet, mapStringToOption),
+                    map(newlyCreatedOptionsPrime, mapStringToOption),
+                );
+            }
+
+            window.console.warn('isCreatable && !isMultiSelect case is not yet implemented');
```
kellyi: We do not currently have a select box with these settings here but I left open the possibility that we could.

src/django/api/serializers.py
```diff
@@ -632,6 +640,62 @@ def get_affiliation_choices(self, claim):
...
+        ]
+
+        union_of_seeds_and_values = list(set(values).union(seeds))
+        union_of_seeds_and_values.sort()
+
+        return [
+            (choice, choice)
+            for choice
+            in union_of_seeds_and_values
+        ]
```
kellyi: I revised how these serializers were working to make them return a list of tuples like `[("Accessories", "Accessories")]` etc. This matches how the other choice serializers work here which makes it easier to reuse the existing `mapDjangoChoiceTuples` stuff in the client.

src/django/api/serializers.py
```diff
@@ -632,6 +640,68 @@ def get_affiliation_choices(self, claim):
...
+            .values_list('facility_product_types', flat=True)
+
+        if new_values.exists():
+            new_values = new_values[0]
+
+        values = [
+            new_value
+            for new_value
+            in new_values or []
+        ]
```
kellyi: I'd like to check my reasoning here a little bit:

- the `values_list` call for `seeds` will always return a list because of our data migration
- the `values_list` call for `facility_product_types` will return `None` if no custom types have been entered
- the `values_list` call for `facility_product_types returns a list of lists when it does return values

So we check whether `new_values.exists()` here to see whether it is not `None` and if it is not, we remove one level of nesting by getting the 0th index. If it is `None`, that's okay because the `or []` thing in the list comprehension above will just make `values` into an empty list.

Subsequently we union the seeds and custom values together before making the tuples to send back as choices.

(Same logic for production types, btw). Does this seem reasonable? I tested it out and it worked both when there were no existing `facility_product_type` values and when there were some existing values.

jwalgran: I tested this reasoning by adding products to 2 different profiles. It does not look like we want `new_values[0]`

```
>>> new_values = FacilityClaim.objects.all().values_list('facility_product_types', flat=True)
>>> new_values
<QuerySet [['Bedskirts', 'Foo'], ['Chaps', 'Foo', 'Spats']]>
>>> new_values[0]
['Bedskirts', 'Foo']
```

We can combine set and itertools.chain to get a set that we can union with our seeds.
```
>>> from itertools import chain
>>> set(chain.from_iterable(new_values))
{'Bedskirts', 'Foo', 'Spats', 'Chaps'}
```

Going to rework this if you agree.

src/django/api/serializers.py
jwalgran: :+1:

src/django/api/serializers.py
kellyi: Sounds fine to me -- the main thing I was interested in doing is to make sure we return the list of choices formatted like `[('A', 'A')]` from the API since that's generally what the client expects for these. Feel free to adjust how we get there, though.

jwalgran: Will do

src/app/src/components/ClaimedFacilitiesDetails.jsx
```diff
@@ -173,9 +178,39 @@ const InputSection = ({
...
+                return filter(
+                    selectOptions,
+                    ({ value: option }) => includes(value, option),
+                );
+            }
+
+            if (isCreatable && isMultiSelect) {
+                const [
+                    optionsInExistingSet,
+                    newlyCreatedOptionsPrime,
```
kellyi: Just noticed that I made a naming mistake here by leaving `Prime` on the end of this, from when I was trying out a couple ways to generate these partitions. I guess it doesn't hurt anything to leave it but if we'd like we can adjust the name to be just `newlyCreatedOptions`.

src/app/src/util/propTypes.js
```diff
@@ -283,4 +283,6 @@ export const approvedFacilityClaimPropType = shape({
     facility_types: arrayOf(arrayOf(string)).isRequired,
     affiliation_choices: arrayOf(arrayOf(string)).isRequired,
     certification_choices: arrayOf(arrayOf(string)).isRequired,
+    product_type_choices: arrayOf(string).isRequired,
+    production_type_choices: arrayOf(string).isRequired,
```
kellyi: Just noticed that this needs to be `arrayOf(arrayOf(string))` now.

jwalgran:
Thanks for helping me finish this up. Sorry your contribution was lost to rebase.

PR #684, opened 2019-07-17
Make small Createable/Multiselect UI refactor

kellyi:
## Overview

I had some very tiny epiphanies about simpler ways to do the multiselect/creatable dropdowns for product types & production types, so I'm making this PR to update the code before we forget about it.

- drop partition/concat/map code to set up selected creatable
multiselect options: it's effectively useless because the createables
don't care whether the selected values are within the list of choices
sent from the API
- adjust conditional setup for Creatable/Select components
- update product_types and production_types propTypes

Connects #652 

## Testing Instructions

- serve this branch and verify that the claimed details form works the same as it does on develop

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/components/ClaimedFacilitiesDetails.jsx
+++ b/src/app/src/components/ClaimedFacilitiesDetails.jsx
@@ -19,8 +19,6 @@ import map from 'lodash/map';
 import filter from 'lodash/filter';
 import includes from 'lodash/includes';
 import isNull from 'lodash/isNull';
-import partition from 'lodash/partition';
-import concat from 'lodash/concat';
 import Select, { Creatable } from 'react-select';
 import { isEmail, isInt } from 'validator';
 import { toast } from 'react-toastify';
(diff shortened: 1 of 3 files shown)


src/app/src/components/ClaimedFacilitiesDetails.jsx
```diff
@@ -191,27 +189,18 @@ const InputSection = ({
...
-                    v => includes(map(selectOptions, 'value'), v),
-                );
-
-                const mapStringToOption = s => ({ value: s, label: s });
-
-                return concat(
-                    map(optionsInExistingSet, mapStringToOption),
-                    map(newlyCreatedOptionsPrime, mapStringToOption),
-                );
+                return map(value, s => ({ value: s, label: s }));
```
kellyi: The deleted bit was code I wrote that in retrospect was basically useless: the point of the Createable is that any option is valid, so there's no need to intersect with the existing choices. Since we know the structure of the choices, we can just map all the values to be that.

src/app/src/components/ClaimedFacilitiesDetails.jsx
```diff
@@ -191,27 +189,18 @@ const InputSection = ({
...
-
-            return [];
+            // isCreatable && !isMultiSelect creates an option object from the value
+            return {
+                value,
+                label: value,
+            };
         })();
 
+        const SelectComponent = isCreatable ? Creatable : Select;
```
kellyi: Since these two components get an identical set of props, we can cut some lines below by conditionally choosing the component here.

Issue #639, opened 2019-06-26
Allow a match to be promoted


jwalgran:
## Overview

Allow a matched list item to become the primary creator of a facility.

### Is your feature request related to a problem? Please describe.

Sometimes the facility details of a matched item are better than the original facility details.

### Describe the solution you'd like

- Create a private `promote` API that accepts a facility list item id
  - Verify that the list item is in a matched status
  - Copy the details from the list item into the facility (country, name, address, location if not null, created_from)
  - Update the facility match records for both list items
  - Return full facility details with additional private matches (like split)
- Alter the "split" dashboard page tile to "Split or Promote"
- Add a "Promote" button to each of the matches displayed on the existing dashboard page. On click:
  - Show a dialog explaining what will happen. On confirm:
    - Call the `promote` API
    - Refresh the details on the dashboard page.

PR #695, opened 2019-07-18
Enable admins to promote a facility match to become the canonical facility

kellyi:
## Overview

This PR enables admins to promote a facility match to become the canonical facility. I updated the split match user interface to become `Adjust` match, and added a `Promote` button for each of the matches.

Clicking `Promote` will call the new API endpoint which updates the facility to use the list item for the selected match as its `created_from` & also updates its name, address, and location data for the new list item. It also writes an entry to the processing results log. Finally, it sends back the facility data to the adjust match screen, updating everything so that it is reflective of the new facility data.

Connects #639 

## Demo

Screen Shot 2019-07-18 at 5 45 08 PM

Screen Shot 2019-07-18 at 5 45 13 PM

## Testing Instructions

- using the standard set of facility data, sign in as an administrator
- merge some facilities together and copy the OAR ID
- paste that OAR ID into the new adjust path's component, then verify that you can
   - promote a match to become the canonical facility, seeing the changes reflected in the UI
   - continue to split facility matches using the existing apparatus.

## Checklist

- [ ] CI passes after rebase


kellyi:
I still need to make an addition to the CHANGELOG for this.

jwalgran:
Looking at this now.

kellyi:
I remembered the q I was trying to recall earlier:

Should I also adjust the match statuses and/or confidence scores in some way for the two exchanged matches?

src/django/api/views.py
```diff
@@ -1057,14 +1057,86 @@ def split(self, request, pk=None):
...
+
+        try:
+            facility = Facility \
+                .objects \
+                .get(pk=pk)
+
+            match = FacilityMatch \
+                .objects \
+                .get(pk=match_id)
+
```
jwalgran: Consider raising `BadRequestException` in these cases
- `match.facility != facility`
We should only allow promotion if the match already points to the facility
- `facility.created_from == match.facility_list_item`
We don't need to promote if the match is already the "winner"

src/django/api/views.py
```diff
@@ -1057,14 +1057,86 @@ def split(self, request, pk=None):
...
+            previous_created_from_id = facility.created_from.id
 
-        return Response('ok')
+            facility.name = match.facility_list_item.name
+            facility.address = match.facility_list_item.address
+            facility.country_code = match.facility_list_item.country_code
+            facility.location = match.facility_list_item.geocoded_point
+            facility.created_from = match.facility_list_item
+            facility.save()
```
jwalgran: Consider setting `facility.changeReason` before saving to something like "Promoted list item {new list item id} from {new list as string} over list item {previous list item id} from {previous list as string}"

src/django/api/views.py
kellyi: Good idea -- I added these checks and also tests for them in 4c8f50d7ab258527c7915cdd825cbfccae7ccf5f

src/django/api/views.py
kellyi: I added this in 5b5ce2bee073396f3fc9875aabadc7c59420dd36

Issue #668, opened 2019-07-10
Require authentication for CSV downloads and log the requests


jwalgran:
## Overview

In concert with requiring authentication for all API endpoints and tracking requests for metrics purposes add a similar authentication requirement and logging system for CSV downloads from the map page.

### Describe the solution you'd like

- Require authentication for CSV downloads from the map page.
  - If user is not logged in, show a dialog prompting them to log in or register when clicking "Download"
- Create a `DownloadLog` table
  - `user` FK to user
  - `path` null False, blank false
  - ~`result_count`~ `record_count` non-null Int
  - `created_at` and `updated_at`
- Add a ~`download`~ `log_download` action to the Facilities API 
  - Responds to `POST` only
  - Do not show it in the swaggar docs
- POST to `download` before creating CSV
  - Pass the path including the query string arguments use to filter the facility search
  - Include a result count argument.

PR #697, opened 2019-07-19
Require authentication for facility CSV downloads and log requests

jwalgran:
## Overview

Require a login to download a facility CSV from the map page and log all CSV download requests to the database. 

Connects #668

## Demo

Screen Shot 2019-07-23 at 7 10 43 PM

## Notes

We have kept the method, model, and model attribute names relatively generic so we have the option of logging other downloads using the same code while avoiding the need to change names.

## Testing Instructions

This assumes the fixture data has been loaded and processed with `./scripts/resetdb`

* Log in as `c2@example.com`
* Search for `shoes`
* Click "Download CSV" and verify that is behaves as before.
* Run `./scripts/manage dbshell`
* Run `select * from api_downloadlog` and verify that the `path` and `record_count` columns match.
* Stop `./scripts/server`. Click "Download CSV" and verify that error toast is shown. Start `./scripts/server`
* Log out
* Search for "shoes" again and click "Download CSV." Verify that the "Log In To Download" dialog is shown.
  * Verify each button behavior (Cancel, Register, and Log In). 

## Checklist

- [ ] CI passes after rebase


kellyi:
Setting this up to take a look!

src/app/src/actions/logDownload.js
```diff
@@ -0,0 +1,32 @@
...
+
+export const startLogDownload =
+    createAction('START_LOG_DOWNLOAD');
+export const failLogDownload =
+    createAction('FAIL_LOG_DOWNLOAD');
+export const completeLogDownload =
+    createAction('COMPLETE_LOG_DOWNLOAD');
+
+export function logDownload(rows, callback) {
+    return (dispatch) => {
```
kellyi: Since the facilities data lives in the app's Redux store state, conceivably we could change this function to actually perform the download directly rather than having to pass `rows` and `callback` in as arguments. If we handled it this way, we could get the facilities data here via `getState` and then directly import the `downloadFacilitiesCSV` utility function to execute it in the promise chain.

src/app/src/actions/logDownload.js
```diff
@@ -0,0 +1,32 @@
...
+        dispatch(startLogDownload());
+
+        const path = `${window.location.pathname}${window.location.search}${window.location.hash}`;
+        const recordCount = rows ? rows.length : 0;
+        return apiRequest
+            .post(makeLogDownloadUrl(path, recordCount))
+            .then(() => {
+                callback(rows);
+                return dispatch(completeLogDownload());
+            })
```
kellyi: Nitpick but consider

```
.then(() => callback(rows))
.then(() => dispatch(completeLogDownload())
.catch....
```

Or whatever similar construction would work with a potential revision to use getState.

src/django/api/models.py
```diff
@@ -1190,3 +1190,26 @@ class ProductionType(models.Model):
...
+    Log CSV download requests from the web client
+    """
+    user = models.ForeignKey(
+        'User',
+        null=False,
+        on_delete=models.CASCADE,
+        help_text='The User account that made the request'
+    )
+    path = models.CharField(
+        max_length=2083,
```
kellyi: Is 2083 a maximum value for paths?

kellyi:
I tested this, btw without reconciling the merge conflict, which is just that one of the reducers on develop now has a different name than it does on this branch.

jwalgran:
Rebased to resolve the merge conflict.

src/django/api/models.py
jwalgran: There is a well researched StackOverflow answer that contributed to this choice of maximum value
https://stackoverflow.com/a/417184

In summary, URLs larger than this bump into some limits in browsers and search crawling infrastructure so it is a reasonable maximum value.

src/app/src/actions/logDownload.js
jwalgran: 👍

src/app/src/actions/logDownload.js
jwalgran: We discussed this in person and it does make sense given the future refactoring of the search result handling to have this action fetch the results from the state rather than having the caller pass them in.

jwalgran: Implemented in fixup c53103b

src/app/src/actions/logDownload.js
jwalgran: Implemented in fixup c53103b

jwalgran:
Thanks for the great architecture suggestions. The implementation is better for it.

Issue #708, opened 2019-07-25
Gather additional metrics


jwalgran:
## Overview

Gather additional metrics.

### Describe the solution you'd like

- Average number of affiliations across all contributor types per facility
- Average number of brand contributors per facility
- Average number of X type contributor per facility
- New contributors by organization type
- Average number of lists per user
- Number of public lists vs private lists
- Number of lists updated

PR #709, opened 2019-07-29
Add admin reports

jwalgran:
## Overview

Provides a quick, simple way to do ad-hoc reporting via raw SQL queries.

Connects #708

## Demo
Screen Shot 2019-07-29 at 9 24 42 AM

Screen Shot 2019-07-29 at 9 13 39 AM

## Notes

We add a pair of new admin views so that the reports are only accessible to superusers.

A new report can be added just by adding a `.sql` file to `/api/reports/`. The name of the report is derived from the name of the `.sql` file.

We return and display the raw SQL along with the results to make easy to debug and answer questions about how the report was generated.

We opt for raw SQL for the flexability it provides. We will never be writing data so the ORM would cause more friction than provide benefit.

## Testing Instructions

* Verify that the admin site at http://localhost:8081/admin works as before.
* Browse http://localhost:8081/admin/reports/ and verify that all the `.sql` files in `/src/django/api/reports/` appear.
* Verify that all of the reports run without error. It is expected that some of the reports will not produce results from the fixture data.
* Verify the logic of the report queries.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/reports/average_list_count_per_contributor.sql
+++ b/src/django/api/reports/average_list_count_per_contributor.sql
@@ -1,0 +1,17 @@
+SELECT
+  CAST (month AS int),
+  round(avg(list_count), 2)
+FROM (
+  SELECT
+    date_part('month', l.created_at) AS month,
+    c.id,
+    COUNT(*) AS list_count
+  FROM api_facilitylist l
+  JOIN api_contributor c ON c.id = l.contributor_id
+  WHERE date_part('month', l.created_at) != date_part('month', now())
+  AND l.is_public = 't'
+  AND l.is_active = 't'
+  GROUP BY date_part('month', l.created_at), c.id
+) s
+GROUP BY month
+ORDER BY month;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Add admin reports [#709](https://github.com/open-apparel-registry/open-apparel-registry/pull/709)
 
 ### Changed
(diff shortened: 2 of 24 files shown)


jwalgran:
I realized while preparing the PR notes that it will help the users of the reports to have the list of available reports sorted by name so I pushed a fixup.

src/django/api/reports/average_list_count_per_contributor.sql
```diff
@@ -0,0 +1,14 @@
+select
```
kellyi: I don't have a real preference but I feel like I've usually seen SQL written with keywords capitalized. However, lowercase is clear enough.

Now that we have raw SQL files in the repository (and the possibility of adding more in the future, or changing these in some way), would it make sense to add a SQL linter to run as part of the `lint` script -- assuming one exists?

kellyi:
This is working well! I do have a question about whether we're sure the output is correct for these queries using the standard set of fixtures. The following reports don't have any results after running `resetdb` when I expect they might:

- public contributors
- unique public list count

There are also a few other reports without any results, but they're reports I would expect not to have any, like reports related to manual matches.

src/django/api/reports/average_list_count_per_contributor.sql
jwalgran: It's faster to write things when you never worry about the shift key, but I checked the Postgres docs and the SQL generated by the Django ORM and they both use all-caps keywords. I'll go ahead and upcase them for consistency.

I did some Googling but did not find a command line SQL linter.

jwalgran:
Both `Public Contributors` and `Unique Public List Count` use an email address string match that looks for our staging/production convention of uploading public lists under openapparel.org email addresses.

```sql
and u.email like '%openapparel.org%'
```

Our fixture data does not have any of these addresses.

jwalgran:
Thanks for the review.

Issue #706, opened 2019-07-24
Replace Google Map component with React-Leaflet using Google basemap


kellyi:
## Overview

Replace Google Map component with React-Leaflet using Google basemap.

This is prep work for displaying more facilities on the map, but it will also make it easier to undertake future map-related changes like polygonal search.

### Describe the solution you'd like

Replace Google Map component with React-Leaflet using Google basemap.

Ensure that the new map component has feature parity with the existing Google Map component.

Ensure that the new component requests its tiles from the correct URL, based on the user's country.

### Additional context

As much as is possible, I think we should stick with using React-Leaflet idioms for handling map interactions rather than going through refs & using imperative code. The current Google Map component uses a lot of imperative code and it's therefore comparatively difficult to reason about.

kellyi:
We also need to ensure that the basemap request routes correctly by country so that the map will continue to work in China.

kellyi:
Looks like this library does what we need as far as setting the region conditionally:

https://www.npmjs.com/package/react-leaflet-google-layer

jwalgran:
👍 	

On the surface it looks like a well managed solo project that is a simple mash up of Carrooi/Js-GoogleMapsLoader and IvanSanchez/Leaflet.GridLayer.GoogleMutant

The GoogleMutant docs say that it needs a promise polyfill for IE 11. Presumably we already have this in our pipline since we are making use of promises already.

kellyi:
For my reference here is what the existing multiple-facilities-at-point popup looks like:

<img width="604" alt="Screen Shot 2019-07-29 at 4 33 21 PM" src="https://user-images.githubusercontent.com/4165523/62080440-a6e06e80-b21e-11e9-982d-3a042146a2ba.png">

PR #710, opened 2019-07-29
Replace map component

kellyi:
## Overview

This PR replaces the existing Google Map component with a React Leaflet map using a Google Basemap (via GoogleMutant, via some React wrappers). I believe it has feature parity with the Google Map component version on staging, although the cluster marker sizes may be slightly different.

Connects #706 

## Notes

Leaflet.markercluster wants a peer dependency of Leaflet 1.3.something. There's an open PR to fix this by adjusting how it declares its peer dependencies, but the yarn install output here will currently report an incorrect peer dependency. Here's the PR:

https://github.com/Leaflet/Leaflet.markercluster/pull/946

The library seems to work regardless of the peer dependency mismatch, however.

## Testing Instructions

- get this branch, then `./scripts/update` and `./scripts/resetdb` and `./scripts/server`
- use the map and verify that searching for facilities works as before and the clusters work as before
- using `shell_plus`, set two facilities to live at `Point(0, 0)`. I did this just to make them easily searchable:

```
>>> from django.contrib.gis.geos import Point
>>> f = Facility.objects.first()
>>> f.location = Point(0, 0)
>>> f.name = "Hello"
>>> f.save()
>>> f = Facility.objects.last()
>>> f.location = Point(0, 0)
>>> f.name = "Hello world"
>>> f.save()
```

- search for `Hello` in the app and verify that you see two facilities at the same point and that the disambiguation thing works the same as it does on staging

## Checklist

- [ ] CI passes after rebase


jwalgran:
Sorry. I was trying to click on the details of the failed PR build and the content moved. I accidentally clicked "ready for review"

kellyi:
No worries!

src/app/src/components/FacilitiesMap.jsx
```diff
@@ -1,519 +1,276 @@
-import React, { Component, Fragment } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
+import { bool, func, number, shape, string } from 'prop-types';
```
kellyi: The diff here isn't really meaningful -- this is an entirely new file, but I deleted the prior one and replaced it with the React-Leaflet implementation.

src/app/src/components/FacilitiesMap.jsx
```diff
@@ -1,519 +1,276 @@
...
+                            return ['cluster-icon-four', [78, 78]];
+                        }
+
+                        return ['cluster-icon-five', [90, 90]];
+                    })();
+
+                    return L.divIcon({
+                        className: iconClassName,
+                        iconSize,
+                        html: `<span>${clusterCount}</span>`,
```
jwalgran: Consider using the application font and tweaking the span style so that the numbers appear more centered.

```diff
diff --git a/src/app/src/components/FacilitiesMap.jsx b/src/app/src/components/FacilitiesMap.jsx
index 8669cc1..ca794f1 100644
--- a/src/app/src/components/FacilitiesMap.jsx
+++ b/src/app/src/components/FacilitiesMap.jsx
@@ -214,9 +214,9 @@ function FacilitiesMap({
                     })();
 
                     return L.divIcon({
-                        className: iconClassName,
+                        className: `cluster-icon ${iconClassName}`,
                         iconSize,
-                        html: `<span>${clusterCount}</span>`,
+                        html: `<span style="margin: -0.1rem; display: block;">${clusterCount}</span>`,
                     });
                 }}
                 onClusterClick={({
diff --git a/src/app/src/styles/css/leafletMap.css b/src/app/src/styles/css/leafletMap.css
index 71c6d98..aea2022 100644
--- a/src/app/src/styles/css/leafletMap.css
+++ b/src/app/src/styles/css/leafletMap.css
@@ -1,3 +1,7 @@
+.cluster-icon {
+    font-family: ff-tisa-sans-web-pro, sans-serif;
+}
+
 .cluster-icon-one {
     width: 53px;
     height: 53px;
```

#### Before
Screen Shot 2019-07-30 at 16 13 52

#### After

Screen Shot 2019-07-30 at 16 13 30

src/app/src/util/constants.facilitiesMap.js
```diff
@@ -7,29 +7,6 @@ export const initialCenter = Object.freeze({
     lng: 5,
 });
 
-export const initialZoom = 1;
+export const initialZoom = 3;
```
jwalgran: Consider 2 as an initial zoom rather than 3. It shows more of the world.

#### Zoom 3

Screen Shot 2019-07-30 at 15 58 59

#### Zoom 2

Screen Shot 2019-07-30 at 15 59 14

src/app/src/components/FacilitiesMap.jsx
kellyi: Cool -- I added this in b615822262e038e6fff06ba846b086e09a285f39

Screen Shot 2019-08-01 at 12 52 18 PM

src/app/src/util/constants.facilitiesMap.js
kellyi: Changed this in fcf2a55b16753cc32160ac0e74a71ba06eb2277e

kellyi:
I set the minZoom to 1 in dd37625c434ecfb613e2b88315e1c1861af9b4c6

src/app/src/components/FacilitiesMap.jsx
```diff
@@ -1,519 +1,276 @@
...
+                            position={[
+                                get(f, 'geometry.coordinates[1]', null),
+                                get(f, 'geometry.coordinates[0]', null),
+                            ]}
+                            icon={
+                                f.id === oarID
+                                    ? selectedMarkerIcon
+                                    : unselectedMarkerIcon
+                            }
+                            onClick={() => navigateToFacilityDetails(f.id)}
```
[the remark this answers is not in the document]
kellyi: Handled this (and also continued adjusting the minZoom setup) in b9d1915ccdde20c77b4f582820a92801de31b79b

src/app/src/components/FacilitiesMap.jsx
```diff
@@ -1,519 +1,276 @@
...
+                        return ['cluster-icon-five', [90, 90]];
+                    })();
+
+                    return L.divIcon({
+                        className: iconClassName,
+                        iconSize,
+                        html: `<span>${clusterCount}</span>`,
+                    });
+                }}
+                onClusterClick={({
```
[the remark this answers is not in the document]
kellyi: I think this is taken care of by b9d1915ccdde20c77b4f582820a92801de31b79b as well: we no longer adjust the zoom level on moving toward the details page, so it'll just keep the existing zoom level.

kellyi:
fd63ec1fdb58f7802d3f7651b31425a8f5f4de01 adjusts the map panning behavior:

Previously we were setting the view anew each time a new facility details page got pushed onto the stack. Now I've adjusted things so that we only set the view around the facility if user has navigated directly to the details screen without traversing the main facilities search.

We discussed this behavior in Slack and determined that it was the preferred behavior.

src/app/src/components/FacilitiesMap.jsx
```diff
@@ -1,519 +1,276 @@
...
+                        ),
+                        get(
+                            facilitiesToDisambiguate,
+                            '[0].geometry.coordinates[0]',
+                            null,
+                        ),
+                    ]}
+                    onClose={() => setFacilitiesToDisambiguate(null)}
+                >
                     <FacilitiesMapPopup
```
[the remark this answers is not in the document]
kellyi: Changed this in 6ebbd43c6207b31830051476be17a1d53b1540a3

Screen Shot 2019-08-01 at 4 20 00 PM

kellyi:
@jwalgran I think this is ready for another look! Apologies for the `fixup! fixup! fixup! etc` stuff.

src/app/src/components/FacilitiesMap.jsx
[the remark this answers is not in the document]
kellyi: Ended up solving this in a different way: viz., we only set the map view for a details page if it is an initial page load, navigating directly to the details page.

kellyi:
> where you don't move the map if you click a result on the map, but do move the map if you click an item in the sidebar

That makes sense to me -- I'll add this interaction before merging.

kellyi:
Going to squash this momentarily but 5e69460675a9c1bb04a294d29e63c9cd20e1ac0e has the changes to make the map panning work like:

- if the selected facility is in the viewport's map bounds when it is selected (technically: when its data returns from the API), do not pan the map
- if the selected facility is not in the viewport's map bounds when it is selected (technically: when its data returns from the API), do pan the map to center it on the facility

I also commented the `useEffects` so that we'll have an easier time understanding them if we ever, for instance, switch to another mapping library.

Issue #705, opened 2019-07-24
Add react-infinite for scrolling facility results


kellyi:
## Overview

In preparation for showing more facilities on the map & having the sidebar results be paginated, replace the current facilities results list with an implementation using react-infinite:

https://github.com/seatgeek/react-infinite

This is a library we used on GARP which supports infinite scrolling by rendering only a subset of a large list into DOM elements.

### Describe the solution you'd like

In preparation for showing more facilities on the map & having the sidebar results be paginated, replace the current facilities results list with an implementation using react-infinite:

https://github.com/seatgeek/react-infinite

This is a library we used on GARP which supports infinite scrolling by rendering only a subset of a large list into DOM elements.

The list should otherwise continue to look and behave as it did prior to switching components.

### Additional Context

Optional. Add any other context, screenshots, gifs, etc about the feature request here.

PR #711, opened 2019-07-29
Wrap facility search results with react-infinite via react-infininte-any-height

jwalgran:
## Overview

Wraps search facility results with `react-infinite` via `react-infininte-any-height` in preparation for loading more than 500 results.

Connects #705

## Notes

React infinite requires that we specify the height of the list item container. We accomplish this by listening for window resize events, tracking the window size in the Redux state, then calculating the proper container size.

`react-infinite-any-height` is a wrapper around `react-infinite` that wraps each list item with a span and uses that span to calculate the height, preventing the need to force all list items to be the same height. There is a lot of variability in the length of facility names and addresses. On narrow screens some list items can be twice as tall as others.

## Testing Instructions

* Run `./scripts/update` to install the new packages
* Verify that the facility search UI works as before

## Checklist

- [x] CI passes after rebase


src/app/package.json
```diff
@@ -25,6 +25,8 @@
         "react-copy-to-clipboard": "5.0.1",
         "react-dom": "16.8.6",
         "react-hot-loader": "4.9.0",
+        "react-infinite": "0.13.0",
+        "react-infinite-any-height": "^2.3.0",
```
kellyi: Consider removing `^` here just to stay in tune with the other imports.

src/app/src/components/FilterSidebarFacilitiesTab.jsx
```diff
@@ -256,40 +259,54 @@ function FilterSidebarFacilitiesTab({
...
     );
 
+    const componentSizes = [
+        64, // Navbar
+        107, // Sidebar heading
+        48, // Sidebar tabs
+        138, // Sidebar controls
+        51, // Footer
+        8, // Additional padding
+    ];
```
kellyi: Wonder if we should set these values as constants somewhere if they're hard-coded in multiple places?

src/app/src/reducers/UIReducer.js
```diff
@@ -20,6 +21,10 @@ const initialState = Object.freeze({
         filterText: '',
         resetButtonClickCount: 0,
     }),
+    window: Object.freeze({
+        innerHeight: typeof window === 'object' ? window.innerHeight : null,
+        innerWidth: typeof window === 'object' ? window.innerWidth : null,
```
kellyi: There is also https://lodash.com/docs/4.17.15#isObject but it probably does the same thing.

kellyi:
We may not need to worry about them but it appears that react-infinity & react-infinite-any-height cause some peer dependency warnings on yarn install.

src/app/package.json
jwalgran: Good idea. Will do.

src/app/src/components/FilterSidebarFacilitiesTab.jsx
jwalgran: Good idea. Since we are styling with JS objects, we should have the luxury of just moving the values out to a shared constant.

src/app/src/reducers/UIReducer.js
jwalgran: Interestingly there is a subtle difference. `_.isObject` considers a function to be an "object"

https://github.com/lodash/lodash/blob/9f865886f8ae22b205107b00b7b65871bfc9a2f0/lodash.isobject/index.js#L35-L38

Good to know, but it doesn't matter in this case. Switching to `isObject` will make things a bit more readable.

src/app/src/components/FilterSidebarFacilitiesTab.jsx
jwalgran: Turns out it is not that easy. There are `min-height` and `padding` directives coming in from our top level stylesheet as well as Material UI that are complicating things. I am going to instead add class tags to the elements that we need to measure and do the height calculation on the fly.

src/app/package.json
jwalgran: Pinned the `react-infinite-any-height` version in fixup a9c5d89

src/app/src/components/FilterSidebarFacilitiesTab.jsx
jwalgran: Implemented a class-based height calculation in fixup abdafcf

src/app/src/reducers/UIReducer.js
jwalgran: Switched to using `isObject` in fixup 2952bdf

jwalgran:
Thanks for the excellent suggestions.

PR #719, opened 2019-08-05
Only pan map if facility not in viewport bbox

### LINKED ISSUE #718: Don't pan map when clicking on a facility marker
## Overview

The map pans when clicking on a visible facility marker. The FacilitiesMap component has logic which was intended to pan the map only when the newly-selected facility is off-screen -- and not when the facility's marker is on-screen -- but the conditional checks in the hook must not have been correct.

### Expected Behavior

Map pans when selecting a facility that is currently NOT within the viewport bbox

Map does NOT pan when selecting a facility that is already within the viewport bbox

### Actual Behavior

Map pans when selecting a facility that is NOT within the bbox

Map ALSO pans when selecting a facility that is within the bbox

kellyi:
## Overview

- pan the map if the newly selected facility is not in the viewport bbox
- don't pan the map if the newly selected facility is already visible in
the viewport bbox

Fixes #718 

## Testing Instructions

- get this branch and verify that this fixes #718 & hasn't introduced other regressions

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Removed
 
 ### Fixed
+- Pan map on selecting a new facility only when the selected facility is off-screen [#719](https://github.com/open-apparel-registry/open-apparel-registry/pull/719)
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


jwalgran:

kellyi:
Awesome, thanks!

Issue #704, opened 2019-07-24
Create map tiling ADR


jwalgran:
## Overview

Write an ADR to document the outcome of a meeting to discuss how to display more facilities on the map simultaneously.

PR #723, opened 2019-08-08
Add vector tiling ADR

kellyi:
## Overview

Adds an ADR documenting our decision to add an `ST_AsMVT`-backed vector tile endpoint to the existing Django web app.

Connects #704 

## Testing Instructions

- read the ADR & verify that it documents our decision accurately and that it is an acceptable decision.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Add vector tile ADR [#723](https://github.com/open-apparel-registry/open-apparel-registry/pull/723)
 
 ### Changed
 
--- a/doc/arch/adr-002-decide-how-to-display-more-facilities.md
+++ b/doc/arch/adr-002-decide-how-to-display-more-facilities.md
@@ -1,0 +1,249 @@
+# Determine How to Display All Facilities on the Map
+
+## Context
+
+The Open Apparel Registry currently includes more than 18,000 facilities. For
+performance reasons, we have paginated the facilities data API endpoint data so
+that it will [return a maximum of 500 results][pagination-pr] for any single
+request. In turn this means that the frontend client will only ever display a
+maximum of 500 facilities at a time, rendered as clustered Leaflet markers via
+React-Leaflet. Facilities API requests are currently filtered using Django
+querysets whose inputs are querystring parameters included in the API requests.
+
+To enable users to view all of the OAR's facilities on the map simultaneously,
+we'll need to update how the API returns facilities for display and how the
+client renders them on the map. At present this means updating the application
+so that it can display 18,000+ facilities simultaneously. Following upcoming MSI
+integration work, we anticipate that the number of OAR facilities will increase
+to around 100,000 -- which the application should be able to map. In addition,
+we also want users to be able to filter these vector tiles by query parameters
+like contributor, facility name, and country, along with the map bounding box.
+
+To accomplish this we have decided to use vector tiles generated, ultimately,
+by PostGIS's [`ST_AsMVT`][st-asmvt] function, rendering them in the frontend
+with [Leaflet Vector Grid][leaflet-vector-grid] (possibly via
+[react-leaflet-vector-grid][react-leaflet-vector-grid]). We've decided to have
+the vector tiles cluster facilities by zoom level, which would limit the number
+of actual points the frontend needs to display at any given time.
+
+This ADR documents a subsequent decision between setting up a dedicated
+`ST_AsMVT`-based vector tile server, like [Martin][martin] or adding a new
+vector tile endpoint to the existing Django web application which would make
+the `ST_AsMVT` query.
+
+## Four Rejected Options
+
+Before landing on an `ST_AsMVT`-based solution, we did consider a few other
+options which we ultimately rejected outright:
+
+### Reusing Existing /facilities API Endpoint
+
+In theory we could remove the `MAX_PAGE_SIZE` limit on the `/facilities` API
+endpoint. In practice this would cause performance problems as the size of the
+GeoJSON response -- and the number of Leaflet markers -- increased. The web app
+would have to serialize tens of thousands of markers for each request, the
+GeoJSON payload for each request could be several megabytes in size, and the
+client would have to put tens of thousands of Leaflet markers in browser
+memory.
+
+### Using Windshaft
+
+While we could potentially use a combination of [Windshaft][windshaft] and
+[Leaflet.utfgrid][leaflet-utfgrid] to render facilities, there wasn't much
+enthusiasm for setting up and maintaining a Windshaft tiler for a few specific
+reasons:
+
+- using Windshaft requires adding another service
+- Windshaft is pretty costly to configure and maintain
+- Windshaft's documentation isn't great
+
+### Creating Static Vector Tiles
+
+We ruled out the idea of creating a static set of vector tiles because the
+OAR's facilities data changes frequently.
+
+### Using a Lambda Function Tiler
+
+Azavea has undertaken some research work to determine the viability of using
+a tiler based on a Lambda function which can connect to PostGIS and call
+`ST_AsMVT`. However, the research has discovered a few limits on this approach,
+such as dealing with function warmup times and handling concurrent database
+connections.
+
+## Two `ST_AsMVT`-Based Approaches
+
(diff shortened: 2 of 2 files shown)


doc/arch/adr-002-decide-how-to-display-more-facilities.md
```diff
@@ -0,0 +1,232 @@
...
+at different zoom levels, tile request traffic will likely be more frequent and
+sustained than requests to the current `/facilities` endpoint.
+
+## Decision
+
+We have decided to add a vector tile endpoint to the existing Django app.
+
+While Martin, in particular, seemed like a compelling solution, we had enough
+questions about using it to discourage us from taking on the complexity of
+using it here.
```
fungjj92: I am very curious about Martin (et al), and I think others are too! We discussed that this would be a good R&D opportunity. Can those who are more involved with this project explain why this project in and of itself can't be that R&D opportunity? What are the limitations I don't know about, and can they be further discussed/captured?

doc/arch/adr-002-decide-how-to-display-more-facilities.md
```diff
@@ -0,0 +1,232 @@
...
+
+Martin's `query_params` appear to be passed in to the database as strings, which
+opens a security hole. While we could create a PostGIS role or user with a
+limited, readonly set of permissions to use solely for the Martin instance,
+doing so requires taking on some additional risk and complexity.
+
+##### Unfamiliarity
+
+We don't have any experience running Martin in production. We've also got
+limited experience using Rust. Together this means a Martin-based tile server
```
fungjj92: A sentence here or at the top of the section that Martin is written in Rust would help clarify

doc/arch/adr-002-decide-how-to-display-more-facilities.md
```diff
@@ -0,0 +1,232 @@
...
+END
+$$ LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
+```
+
+#### Pros
+
+##### Configuration
+
+Martin appears fairly straightforward to configure and its documentation
+encompassed most of what we'd want to do.
```
fungjj92: the power of good docs 🌦

doc/arch/adr-002-decide-how-to-display-more-facilities.md
kellyi: Oh, great point -- I'll add a phrase somewhere to note that Martin is written in Rust.

doc/arch/adr-002-decide-how-to-display-more-facilities.md
kellyi: @jwalgran may amplify or correct this but I think one main reason not to have it here is that the client needs to be able to request subsets of facilities using query parameters like contributor id or facility name etc.

If we added Martin, making changes to these queries (like adding an ability to search on facility description or certifications) would require us to update both the Django queryset filter code and also the PL/pgSQL code used in Martin.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Add vector tile ADR [#723](https://github.com/open-apparel-registry/open-apparel-registry/pull/723)
```
rajadain: Consider linking to the ADR itself

doc/arch/adr-002-decide-how-to-display-more-facilities.md
jwalgran: Yep. If we didn't have to filter, Martin would be a more attractive choice. Writing our own endpoint also allows us to bail out of using `ST_AsMVT` if it proves to be a problem and do our tile rendering on an app server rather than in the database.

fungjj92: Thanks, makes sense. Consider adding the above to the ADR

doc/arch/adr-002-decide-how-to-display-more-facilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+Our main apprehension about adding a tile endpoint to the existing web app is
+that it'll mingle tile requests with other requests in a way that could cause
+performance problems. However, given the size of the OAR's traffic and the
+possibility of addressing traffic increases by scaling the number of app
+instances, this seemed like an acceptable tradeoff.
+
+## Consequences
+
+As a consequence of this decision, we will need to:
```
hectcastro: Potentially too low detail for the ADR, but we may also need to consider:

- Tweaking the number of Gunicorn workers per Gunicorn instance, in addition to the Gunicorn worker type (synchronous vs. Gevent). This would be in service of trying to increase the concurrent request capabilities of a single Gunicorn instance vs. spinning up more of them.
- Looking at caching HTTP headers used by Whitenoise for static assets and evaluating which make sense to replicate for `/tile`.
- Extending [caching behaviors](https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/deployment/terraform/cdn.tf#L43-L62) in CloudFront so that they respect any caching header changes made at the Django level.

CHANGELOG.md
kellyi: Good suggestion -- I'm going to leave this link as is since custom has been to link to the PR.

doc/arch/adr-002-decide-how-to-display-more-facilities.md
kellyi: Sounds good -- I'll add line items about Gunicorn configuration & caching to the list.

doc/arch/adr-002-decide-how-to-display-more-facilities.md
```diff
@@ -0,0 +1,233 @@
...
+
+##### PL/pgSQL Function Sources
+
+Since Martin uses PL/pgSQL functions for its filtering, we would have to
+rewrite some facility filtering logic that currently exists in Django querysets
+in the web app to work in PL/pgSQL. Moreover, each time we added a new filter
+or search option to the web application, we'd have to write a version of the
+same query in PL/pgSQL for the tiler.
+
+##### Security
```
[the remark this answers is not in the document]
kellyi: Going to add this to the `Security` section:

```
Likewise, adding PostGIS-based security just for the tiler may also compel us to
have to figure out how to duplicate features like API key authentication or
facilities-data request logging -- which we've already written once in Django.
```

doc/arch/adr-002-decide-how-to-display-more-facilities.md
```diff
@@ -0,0 +1,233 @@
...
+To enable users to view all of the OAR's facilities on the map simultaneously,
+we'll need to update how the API returns facilities for display and how the
+client renders them on the map. At present this means updating the application
+so that it can display 18,000+ facilities simultaneously. Following upcoming MSI
+integration work, we anticipate that the number of OAR facilities will increase
+to around 100,000 -- which the application should be able to map. In addition,
+we also want users to be able to filter these vector tiles by query parameters
+like contributor, facility name, and country, along with the map bounding box.
+
+To accomplish this we have decided to use vector tiles generated, ultimately,
```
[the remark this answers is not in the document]
kellyi: Makes sense -- I'll pare this down a bit here and move things into "Decision" or whatever is the relevant section.

kellyi:
Thanks for the thoughtful reviews!

doc/arch/adr-002-decide-how-to-display-more-facilities.md
```diff
@@ -0,0 +1,249 @@
...
+an alternate Python library for generating vector tiles.
+
+##### Doesn't Require Creating & Deploying a Different Service
+
+Adding Martin or another vector tile server would increase the number of
+different kinds of services running as part of the OAR, which adds to the
+application's complexity. Keeping the tile endpoint in Django does not require
+adding a new service.
+
+##### Allows Scaling by Increasing the Number of App Instances
```
mmcfarland: I don't think this warrants a change to this ADR nor the decision you've arrived at, but it may be valuable to point out that this may not be a pro (it may not be a con either, but is certainly a consequence). I understood the implication of this statement to be that you _lose_ the ability to scale the the tiling and the API endpoints separately, they must be done in tandem. You also lose the ability to provision the API infrastructure independently of the tiler. For example, if the tiling endpoints end up requiring higher memory allocations, you will be obligated to provision all API instances with the extra resources, and will take that hit if you need to scale on the API-only dimension.

Issue #655, opened 2019-07-09
Add ability for Superusers to download list of contributors from Dashboard


katieashaw:
## Overview

There is currently no way for a Superuser to download a list of OAR contributors from the back end of the map

### Expected Behavior

Ability to download a list of OAR contributors should be made possible from the dashboard, and include a date column for when they subscribed / contributed to the platform, and whether or not they opted into the OAR newsletter.

### Actual Behavior

It's currently not possible for Superusers to download a list of contributors from their dashboard.

### Steps to Reproduce

* instructions for reproducing the bug
* prefer bulleted list
* describe the environment: device, os, browser/runtime

### Demo

Optional. Screenshots, gifs, `curl` examples, console errors, log files, etc.

### Additional context

Optional. Add any other context about the problem here.

PR #726, opened 2019-08-12
Add contributor and mailing list admin reports and the ability to download admin reports

jwalgran:
## Overview

We regularly need to get a list of contributors who have opted into the newsletter. Including the unfiltered list of non-public contributors is a convenience.

All the data is already available so it is easy to include a download link with a `data` url. This does create duplicate data on the page, effectively doubling the size of the report, but this system is designed to be simple, not efficient.

Connects #655

## Testing Instructions

This assumes the fixture data has been loaded.

* Use `shell_plus` or `dbshell` to update some of the "c*@example.com" contributors and change `should_receive_newsletter` to true.
* Log in as `c1@example.com`
* Browse http://localhost:8081/admin/reports/mailing-list-recipients/ and verify that the users changed in the first step appear.
* Browse http://localhost:8081/admin/reports/countributors/ and verify that all the contributors appear.
* Verify that the "Download" link on the report page works.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 ### Added
 - Add vector tile ADR [#723](https://github.com/open-apparel-registry/open-apparel-registry/pull/723)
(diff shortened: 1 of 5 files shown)


jwalgran:
Thanks for the review.

Issue #734, opened 2019-08-16
TypeError: Cannot read property 'lat' of null


jwalgran:
## Overview

"Cannot read property 'lat' of null" client exception raised when clicking on a cluster of facilities on the map

### Expected Behavior

Clicking any cluster zooms the map.

### Actual Behavior

The map shows an error message until refresh

### Steps to Reproduce

* On production, browse https://openapparel.org/?q=Saitex
* Click the 7 cluster, then the 6 cluster

### Demo

![image](https://user-images.githubusercontent.com/17363/63189220-53d92a80-c018-11e9-9e6b-7b0c06eb60d5.png)

### Additional context

View details in Rollbar: [https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/217/](https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/217/)

```

TypeError: Cannot read property 'lat' of null
  File "../node_modules/leaflet/dist/leaflet-src.js", line 1541, in Object.project
              lat = Math.max(Math.min(max, latlng.lat), -max),
  File "../node_modules/leaflet/dist/leaflet-src.js", line 1400, in Object.latLngToPoint
          var projectedPoint = this.projection.project(latlng),
  File "../node_modules/leaflet/dist/leaflet-src.js", line 3752, in e.project
          return this.options.crs.latLngToPoint(toLatLng(latlng), zoom);
  File "../node_modules/leaflet/dist/leaflet-src.js", line 3771, in e.latLngToLayerPoint
          var projectedPoint = this.project(toLatLng(latlng))._round();
  File "../node_modules/leaflet/dist/leaflet-src.js", line 9060, in _updateContent
          var pos = this._map.latLngToLayerPoint(this._latlng),
  File "../node_modules/leaflet/dist/leaflet-src.js", line 8965, in setContent
          this._updatePosition();
  File "../node_modules/leaflet/dist/leaflet-src.js", line 8899, in e.onAdd
          this.update();
  File "../node_modules/leaflet/dist/leaflet-src.js", line 9166, in e.onAdd
          DivOverlay.prototype.onAdd.call(this, map); // @namespace Map
  File "../node_modules/leaflet/dist/leaflet-src.js", line 6195, in _layerAdd
          this.onAdd(map);
  File "../node_modules/leaflet/dist/leaflet-src.js", line 4196, in e.whenReady
            callback.call(context || this, {
```

jwalgran:
[...]

PR #737, opened 2019-08-21
Prevent map crash by not using strict equality when comparing point coordinates

jwalgran:
## Overview

Points were not being combined into a disambiguation popup and throwing an exception when a precision or rounding error prevented the points from being strictly equal.

Connects #734

## Testing Instructions

* Run `./scripts/resetdb`
* Run `./scripts/manage dbshell` and set the same location for all facilities in VN.
```
UPDATE api_facility 
SET location = st_geomfromtext('POINT(106.8909948 10.9393085)', 4326) 
WHERE id IN (
  SELECT id 
  FROM api_facility 
  WHERE country_code = 'VN'
);
```
* Browse http://localhost:6543/?countries=VN
* Click the `97` cluster. Verify that the map component does not crash and a disambiguation popup is displayed.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - Restore django-waffle admin pages [#732](https://github.com/open-apparel-registry/open-apparel-registry/pull/732)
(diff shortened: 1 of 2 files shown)


jwalgran:
Thanks for the review.

Issue #742, opened 2019-08-21
Allow loading all search results into the sidebar a page at a time and remove the textbox filter from results tab


jwalgran:
## Overview

Allow loading all search results into the sidebar a page at a time and remove the textbox filter from results tab

### Describe the solution you'd like

- Allow loading all search results into the sidebar a page at a time and remove the textbox filter from results tab
- Keep functionality behind the `vector_tile` feature flag, reverting to old behavior when the feature flag is not active

PR #750, opened 2019-08-26
Fetch next page of facilities via infinite scroll component

kellyi:
## Overview

Adjust facilities tab so that it will fetch the next page of facilities
on reaching the bottom of the infinite scroll list.

Also:

- remove some now unused utility code & tests
- update facilities list endpoint to return results in alphabetical
order
- update url to retrieve facilities to request pages of 25 results at a
time
- add a loading indicator to display contextually at the temporary bottom of the list when there's another page of items to fetch

Connects #742

## Notes

I'm going to make a new issue for scroll restoration. I had worked out part of a solution using a custom hook, but it's most likely an enhancement that we can address later if desired.

## Testing Instructions

- serve this branch, then search for some facilities
- verify that the results return in alphabetical order in the sidebar and that a new page is fetched each time you reach the bottom of the list (until no pages remain)
- verify that any filters you've applied remain applied when fetching a new page of results

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 ### Added
 - Adjust marker icon on selecting a new facility on the vector tiles layer [#749](https://github.com/open-apparel-registry/open-apparel-registry/pull/749)
+- Fetch next page of facilities while scrolling through sidebar list [#750](https://github.com/open-apparel-registry/open-apparel-registry/pull/750)
 
 ### Changed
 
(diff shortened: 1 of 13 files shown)


kellyi:
Yep, feature flagging this should be easy. In this case what I will do is just add a temporary component that is named `NonVectorTileFilterSidebarFacilitiesTab` and just includes the old code, then set things up to display conditionally based on whether the feature flag is set.

kellyi:
@jwalgran 73ad944141eae0e6add44eaed1127d454547341c adds back the old facilities sidebar functionality so that we can toggle the behavior with the `vector_tile` switch. I opted to rename the old component rather than the new one, since the download CSV PR has some changes targeting the new component.

kellyi:
Thanks for your help with this!

Issue #743, opened 2019-08-21
Allow downloading all search results as CSV


jwalgran:
## Overview

Allow signed in users to download all search results as a CSV.

### Describe the solution you'd like

- Use facility list download as a reference
- Show spinner when clicking the button until all pages are loaded.
- Keep behind the vector_tile feature flag.

PR #752, opened 2019-08-29
Enable signed-in users to download all facility results as a CSV

kellyi:
## Overview

Using the machinery implemented for the infinite scroll sidebar, enable
signed-in users to download a CSV containing all of the facilities
matching any given search params, including all the OAR's facility data
when no search params are present.

Also increase the size of the facilities pages from 25 to 100. We want
to keep this value shared between the infinite scroll and the CSV
download so that we can reuse the existing machinery and so that after
downloading a CSV, the user can scroll through all the facilities in the
sidebar and can download the CSV again without undertaking new requests.

The change also adds a loading indicator to display in place of the
download CSV button while a download is underway.

Connects https://github.com/open-apparel-registry/open-apparel-registry/issues/743

## Testing Instructions

- serve this branch, then sign in and search for facilities
- click "Download CSV" and verify that it downloads all of them
- sign out, then search for facilities
- click "Download CSV" and verify that it fails but doesn't crash the app.

## Checklist

- [ ] CI passes after rebase


kellyi:
Rebased on develop and retargeted.

jwalgran:
Looking at this now.

kellyi:
Thanks!

Issue #740, opened 2019-08-21
DRAFT: Make disambiguation popup work with vector tile layer


jwalgran:
# Draft

This is a draft issue, and it will need to clarify specs and formalize what
counts as "done" before work begins on it.

Check with the issue's creator before starting work on this issue.

## Overview

A clear description of the feature request.

### Describe the solution you'd like

- Include name and address in the vector tile for map features.

### Additional Context

Optional. Add any other context, screenshots, gifs, etc about the feature request here.

kellyi:
We discussed this being very similar to the work for #741. It appears that we can use `ST_ClusterWithin` to do this via PostGIS in the tiler file, and then give clustered features an array property that would include dicts for each facility that has been clustered in that feature with the OAR ID, name, and address of the facility.

From there we can set up the react/leaflet component to conditionally render these kinds of markers in some way to indicate that they are clusters, attaching the disambiguation popup to have it work as before.

We also considered not limiting the disambiguation feature by zoom level -- if it's zoomed in closely enough to render a marker rather than the grid, clicking on a cluster could show the popup without zooming in. We could make clicking on an item in the popup also pan the map, if necessary, if we're not at the max zoom level.

PR #760, opened 2019-09-04
Add vector-tile/multifacility disambiguation popup

kellyi:
## Overview

Set up the multi-facility disambiguation popup to work with the new
vector tile layer component by making the following changes:

- add the `name` and `address` fields to the features returned in the
vector tile, along with the x, y and z arguments representing the
location of the tile
- use the x y and z tile arguments to find the vector tile containing a
clicked facility marker, then check whether there are any other facilities at
the same point using a `_.isEqual` check
- if a clicked facility marker only has one facility, just continue
selecting it as before
- if a clicked facility marker represents multiple facilities, display
the existing disambiguation popup component wired up to display the
facility data
- always highlight the marker for the currently selected facility if
viewing a details page

Connects #740 

## Demo

Screen Shot 2019-09-05 at 10 42 26 AM
Screen Shot 2019-09-05 at 10 42 33 AM

## Notes

The imperative code in the useEffect handlers for this file has become a little more difficult to reason about. This was also the case for the prior imperative class-based code for the Google Map component (and for the Mapbox map component before that). I'm not sure how best to make it more legible at a glance but I'm open to suggestions.

## Testing Instructions

- serve this branch with the standard set of fixtures and vector tiles turned on
- in the terminal run `manage shell plus` and then create a set of clustered facilities at point 0,0 by running:

```
from django.contrib.gis.geos import Point
null_island = Point(0, 0)
Facility.objects.filter(country_code="IT").update(location=null_island)
```

- use the map page and verify the following states:

1. when a facility is selected, its marker has the selected marker image (and only that marker has the selected marker image, all others being unselected -- excluding, of course, when a facility at the same point is selected)

2. when the marker at 0,0 is clicked, the popup displays with the facilities listed. Clicking on any of the facilities in it should select that facility, change that marker to be selected, and change any previously selected markers to unselected

3. navigating directly to an OAR ID details page should make that marker be selected automatically

4. clicking reset should close any open popups and reset the map state

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/components/VectorTileFacilitiesLayer.jsx
+++ b/src/app/src/components/VectorTileFacilitiesLayer.jsx
@@ -2,25 +2,39 @@ import React, { useEffect, useRef, useState } from 'react';
 import { bool, func, number, string } from 'prop-types';
 import { connect } from 'react-redux';
 import VectorGridDefault from 'react-leaflet-vectorgrid';
-import { withLeaflet } from 'react-leaflet';
+import { withLeaflet, Popup } from 'react-leaflet';
 import L from 'leaflet';
 import isEmpty from 'lodash/isEmpty';
 import get from 'lodash/get';
+import noop from 'lodash/noop';
+import filter from 'lodash/filter';
+import map from 'lodash/map';
+import isEqual from 'lodash/isEqual';
+import intersection from 'lodash/intersection';
+import sortBy from 'lodash/sortBy';
+
+import FacilitiesMapPopup from './FacilitiesMapPopup';
 
 import { createQueryStringFromSearchFilters } from '../util/util';
 
 const VectorGrid = withLeaflet(VectorGridDefault);
 
-const createMarkerIcon = iconUrl => L.icon({
-    iconUrl,
-    iconSize: [30, 40],
-    iconAnchor: [15, 40],
-});
+const createMarkerIcon = iconUrl =>
+    L.icon({
+        iconUrl,
+        iconSize: [30, 40],
+        iconAnchor: [15, 40],
+    });
 
 const selectedMarkerIcon = createMarkerIcon('/images/selectedmarker.png');
 const unselectedMarkerIcon = createMarkerIcon('/images/marker.png');
 
-function useUpdateTileURL(tileURL, performingNewSearch, resetButtonClickCount) {
+function useUpdateTileURL(
+    tileURL,
+    performingNewSearch,
+    resetButtonClickCount,
+    closeMultipleFacilitiesPopup,
+) {
     const [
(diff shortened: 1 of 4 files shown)


src/app/src/components/VectorTileFacilitiesLayer.jsx
```diff
@@ -77,12 +95,100 @@ const useUpdateTileLayerWithMarkerForSelectedOARID = (oarID) => {
             });
 
             setCurrentSelectedMarkerID(oarID);
+        } else if (includes(oarIDsAtSamePoint, oarID) || includes(oarIDsAtSamePoint, currentSelectedMarkerID)) {
```
kellyi: This can be `intersection(...).length)`

src/app/src/components/VectorTileFacilitiesLayer.jsx
```diff
@@ -108,30 +231,76 @@ const VectorTileFacilitiesLayer = ({
...
+                multipleFacilitiesAtPointPosition &&
+                multipleFacilitiesAtPoint &&
+                multipleFacilitiesAtPoint.length && (
+                    <Popup
+                        position={multipleFacilitiesAtPointPosition}
+                        onClose={closeMultipleFacilitiesPopup}
+                    >
+                        <FacilitiesMapPopup
+                            facilities={multipleFacilitiesAtPoint}
+                            closePopup={noop}
```
kellyi: Needs a function here

src/app/src/components/VectorTileFacilitiesLayer.jsx
```diff
@@ -108,30 +238,81 @@ const VectorTileFacilitiesLayer = ({
...
+            }
+
+            if (facilitiesAtSamePoint.length === 1) {
+                return handleMarkerClick(data);
+            }
+
+            setMultipleFacilitiesAtPoint(facilitiesAtSamePoint);
+            return setMultipleFacilitiesAtPointPosition(data.latlng);
+        } catch (e) {
+            window.console.error(e);
```
jwalgran: Is this left over from debugging? Does this propagate to rollbar or should we not catch the exception so that it does get logged to rollbar?

kellyi: I believe this is not propagated to Rollbar. I added this whole try/catch block because I don't want a mistake in the click code to crash the entire map, but I left the console.error line so that we could easily debug it if it occurred on production.

Would it be worthwhile to log it to Rollbar?

kellyi:
Going to merge this! If it turns out we want any hypothetical click error to go to Rollbar we can circle back and add that when completing the vector tiles map.

Issue #765, opened 2019-09-05
Show facility search results by default


jwalgran:
## Overview

Show facility search results by default

### Describe the solution you'd like

- Rearrange tabs so that results comes before search.
- When you arrive on the map page, always run a search.
- Add a loading state to the facilities list.

kellyi:
See https://github.com/open-apparel-registry/open-apparel-registry/issues/754 for more context

PR #771, opened 2019-09-05
Fetch all facilities by default on main page when vector tile feature is active

kellyi:
## Overview

When the vector tile feature is active, fetch all facilities by default
on the main map and sidebar page. Also make the "Facilities" tab the
initially displayed tab on loading the map page if the vector tile
feature is active.

Add guards to retain prior map behavior when the vector tile feature is
not active.

Connects #765 

## Demo

Screen Shot 2019-09-05 at 5 59 41 PM

## Notes

Making this work while keeping the same functionality as before when the feature flag is off was a little tricky, but I think I came up with a good way to handle it which was to defer rendering any components that need to know the state of feature flags until the feature flags have returned from the API.

## Testing Instructions

Using the standard set of fixtures, serve this branch, then...

With the `vector_tile` switch turned *off*, use the app and verify that it still works as it did before.

With the `vector_tile` switch turned *on*, use the app and verify that the facilities tab populates with data as expected.

In particular: clicking the "reset" button for the vector tiles version will now undertake a new search.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/components/FilterSidebar.jsx
+++ b/src/app/src/components/FilterSidebar.jsx
@@ -1,10 +1,12 @@
 import React, { Component } from 'react';
-import { func, oneOf } from 'prop-types';
+import { bool, func, oneOf } from 'prop-types';
 import { connect } from 'react-redux';
 import AppBar from '@material-ui/core/AppBar';
 import Tabs from '@material-ui/core/Tabs';
 import Tab from '@material-ui/core/Tab';
+import CircularProgress from '@material-ui/core/CircularProgress';
 import { Route } from 'react-router-dom';
+import get from 'lodash/get';
 
 import FilterSidebarGuideTab from './FilterSidebarGuideTab';
 import FilterSidebarSearchTab from './FilterSidebarSearchTab';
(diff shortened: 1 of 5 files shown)


jwalgran:
Looking at this now.

src/app/src/components/FilterSidebar.jsx
```diff
@@ -95,15 +103,27 @@ class FilterSidebar extends Component {
                 </p>
             </div>);
 
-        const activeTabIndex = filterSidebarTabs
+        const orderedTabsForSidebar = vectorTileFeatureIsActive
+            ? filterSidebarTabs.slice().reverse()
```
jwalgran: Clever use of `reverse` here. I think that having "Guide" last makes sense.

src/app/src/components/FilterSidebar.jsx
```diff
@@ -95,15 +103,27 @@ class FilterSidebar extends Component {
...
+        const activeTabIndex = orderedTabsForSidebar
             .findIndex(({ tab }) => tab === activeFilterSidebarTab);
 
         const handleTabChange = (_, value) => {
-            const changeTab = [
-                makeGuideTabActive,
-                makeSearchTabActive,
-                makeFacilitiesTabActive,
-            ][value];
+            const changeTabDictionary = vectorTileFeatureIsActive
```
jwalgran: Consider `changeTabList` to reflect the arrayness.

kellyi: Good catch -- I'll rename before merging.

Issue #729, opened 2019-08-13
The parent company link on the facility claim review page links to the wrong contributor


jwalgran:
## Overview

The parent company link on the facility claim review page links to the wrong contributor.

### Expected Behavior

Clicking the parent company link on the facility claim review page navigates to the correct contributor page.

### Actual Behavior

Clicking the parent company link on the facility claim review page navigates to the correct contributor page. The wrong ID is being used to build the link URL.

### Steps to Reproduce

This assumes the fixture data has been loaded

* Use `shell_plus` to create a user account without a related contributor. This ensures the ID sequences between the two model types are out of sync.
```
>>> u = User(email='test@example.com')
>>> u.set_password('testing123')
>>> u.save()
```
* Register and confirm a new account. 
* Log in with that account and open a facility claim. Set the parent company to the newly created contributor associated with the newly registered account
* Log in as `c1@example.com` and browse http://localhost:6543/dashboard/claims
* Open the new claim and verify that the parent company link goes to the wrong page.

### Additional context

We show the link correctly on the facility details page when there is an approved claim.

rajadain:
After following the steps to reproduce this, and viewing the created claim, I see this snippet retrieved from the `/api/facility-claims/1` endpoint:

```json
{
  "contributor": {
    "id": 102,
    "name": "TT",
    "description": "TT",
    "website": "",
    "contributor_type": "Civil Society Organization",
    "other_contributor_type": "",
    "facility_lists": [],
    "is_verified": false
  },
  "facility_parent_company": {
    "id": 100,
    "name": "TT"
  },
}
```

The `facility_parent_company.id` field is used to populate the Facility Parent Company link. Upon clicking the link, the page the user is taken to fetches its data from the `/user-profile/100` endpoint, which fetches the user's data:

https://github.com/open-apparel-registry/open-apparel-registry/blob/df05229571713239feb58a96cd6cd9a118921980/src/django/api/views.py#L214-L223

If, in the claim detail view, the user instead clicks the Claim Contributor link, they are taken to the same profile page, but for id 102 corresponding to the user.

---

The fundamental issue seems to be that there is no detail page for Facility Parent Company entities, as both Claim Contributors and Facility Parent Company links open the same page. A solution would be to :one: make a new page for Facility Parent Companies, rather than reusing the Claim Contributors profile page.

A simpler, more temporary alternative could be to :two: make both links point to the Claim Contributor page, which has the correct details.

[...]

PR #772, opened 2019-09-06
Correct Facility Parent Company link

rajadain:
## Overview

There is no detail page for Facility Parent Company entities, as both Claim Contributors and Facility Parent Company links open the same page. In cases when the IDs of the two are not in sync, clicking the Facility Parent Company link would go to the wrong page.

By making the link always point to the Claim Contributor's id, we ensure that it is always correct.

Connects #729 

## Testing Instructions

* Follow the instructions in the issue to simulate a situation where the user id and parent company id are out of sync
* Submit a claim
* Login as `c1@example.com` and go to the dashboard to view its details
* Click the Claim Contributor link. Ensure it works.
* Click the Facility Parent Company link. Ensure it works.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix a bug which prevented the facility claims dashboard page header from loading [#778](https://github.com/open-apparel-registry/open-apparel-registry/pull/778)
 - Fix a bug which prevented the vector tile marker layer from rendering [#782](https://github.com/open-apparel-registry/open-apparel-registry/pull/782)
 - Capture infinite-scroll emitted bug in an ErrorBoundary in order not to crash the map component [#777](https://github.com/open-apparel-registry/open-apparel-registry/pull/777)
+- Correct Facility Parent Company link [#772](https://github.com/open-apparel-registry/open-apparel-registry/pull/772)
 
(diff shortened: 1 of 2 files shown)


kellyi:
Testing this out and I don't think it's doing the correct thing:

For this link, it should go to the contributor page for "Civil Society Organization G" but instead it is going to "Factory A":

Screen Shot 2019-09-09 at 1 34 49 PM

kellyi:
Chatted about this and it seems like we need to update the serializer to return the correct id value. We'll also need to update https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/components/FacilityDetailSidebarClaimedInfo.jsx#L85 to match whatever changes we decide to make to the serializer and the dashboard claim details page.

rajadain:
I reverted the original front-end implementation and took another look at the serializer. Turns out we were inconsistent about the parent company's `id`, so I standardized to using the company's admin's `id`. This fixes the link issue on the front-end as well. Ready for another look.

kellyi:
Heads up that there's a merge conflict in the CHANGELOG file again.

kellyi:
This has another merge conflict.

rajadain:
That should fix it.

rajadain:
Thanks for taking a look, and for talking me through this. Will merge when green.

Issue #762, opened 2019-09-05
Incorporate search parameters into cache key


jwalgran:
# Draft

This is a draft issue, and it will need to clarify specs and formalize what
counts as "done" before work begins on it.

Check with the issue's creator before starting work on this issue.

## Overview

Include a new segment in the cache key that encodes the filter query string parameters.

### Is your feature request related to a problem? Please describe.

We want to cache tiles produced from searches, because they can be more expensive to build than unfiltered tiles.

### Describe the solution you'd like

- Sort query string filter arguments by key
- Sort any array values
- Stringify the normalized object, compute the SHA1 hash, and take the first 7 characters
- Include the short hash as another hypen-separated segment in the cache key.

kellyi:
I checked and the way the querystring builds in the client already guarantees the order the distinct parameters are listed, but it does not yet sort within a param. e.g. `q=hello` will always appear before `countries=AF` but `countries=AF` will not always appear before `countries=CN`

PR #773, opened 2019-09-06
Encode querystring params in vector tile cache key

kellyi:
## Overview

Encode qs params into vector tile cache key sent with client tile
requests. We do this client side because the client is the source of the
request and already knows what filters have been applied -- and because
the server sent cache key only encodes the version number and the
timestamp of the most recently updated facility.

The server already only checks for the presence or absence of a cache
key in the URL path, so this implementation leaves that unchanged. For future CF or S3
based caching, the path will still be set up to "just work" with
requests for specific sets of tiles.

Connects #762

## Testing Instructions

- serve the app with the vector tile feature switched on, and verify that everything still works as before but that we are now encoding the filters into the tile cache key used by the client

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/util/util.js
+++ b/src/app/src/util/util.js
@@ -28,6 +28,7 @@ import every from 'lodash/every';
 import { isEmail, isURL } from 'validator';
 import { featureCollection, bbox } from '@turf/turf';
 import { saveAs } from 'file-saver';
+import hash from 'object-hash';
 
 import env from './env';
 
@@ -130,6 +131,9 @@ export const makeLogDownloadUrl = (path, recordCount) => `/api/log-download/?pat
 
 export const getValueFromObject = ({ value }) => value;
 
+const createCompactSortedQuerystringInputObject = (inputObject = []) =>
+    compact(inputObject.map(getValueFromObject).slice().sort());
+
 export const createQueryStringFromSearchFilters = ({
     facilityFreeTextQuery = '',
     contributors = [],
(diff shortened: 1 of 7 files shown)


src/app/src/util/util.js
```diff
@@ -273,6 +276,12 @@ export const getContributorFromQueryString = (qs) => {
     return parseInt(contributor, 10);
 };
 
+export const createTileURLWithQueryString = (qs, key) =>
+    `/tile/facilitygrid/${key}/{z}/{x}/{y}.pbf`.concat(isEmpty(qs) ? '' : `?${qs}`);
+
+export const createTileCacheKeyWithQSParams = (qs, key) =>
+    `${key}-${window.btoa(qs).slice(0, 8) || 'all'}`;
```
jwalgran: Base64 encoding is not a good choice because "/" is one of the characters used in base64 (https://www.w3schools.com/jsref/met_win_btoa.asp). If we want to continue generating this key segment on the client we should consider:

- Requiring a hash function implementation (perhaps https://github.com/puleos/object-hash)
- Writing our own encoder.

kellyi: Oof, thanks for catching that. I pushed up 0e285a9699ba22647c828a3be9100105cbbaaa06
 which uses the object-hash library instead.

kellyi:
Great! Thanks for your help with this. Going to merge once the CI build is complete.

Issue #763, opened 2019-09-05
Write map tiles to S3


jwalgran:
## Overview

Write generated vector tiles to an S3 bucket so they can be served via Cloudfront (fallback from cloudfront/s3 404 to the tile endpoint will be provisioned in #764)

### Is your feature request related to a problem? Please describe.

Our data updates infrequently so caching tiles in S3 will take most of the burden off of the tile endpoint. The tile endpoint includes a cache key in the URL path so tiles can be effectively cached forever.

### Describe the solution you'd like

- Setup django storages with s3 backend (https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html)
- Write vector tiles to s3 after render via django storages
 - Don't write tile in dev, but allow dev to be set to use s3 for testing
 - Make tile request base url configurable per environment

rajadain:
Superseded by #779.

Issue #764, opened 2019-09-05
DRAFT: Provision cloudfront/s3/route 53 for vector tile serving


jwalgran:
# Draft

This is a draft issue, and it will need to clarify specs and formalize what
counts as "done" before work begins on it.

Check with the issue's creator before starting work on this issue.

## Overview

Required by #763

### Describe the solution you'd like

- Add terraform provisioning code to create a system like Model My Watershed. Cloudfront targeting an S3 bucket which falls back to the app tile endpoint (the app will write tiles back to S3 in #763)
- Forward original referrer and custom header to the tile endpoint if at all possible.

### Additional Context

Optional. Add any other context, screenshots, gifs, etc about the feature request here.

rajadain:
Superseded by #779.

PR #776, opened 2019-09-06
Cache vector tiles for a year

rajadain:
## Overview

Add a `Cache-Control: max-age=31536000` header to the vector tile responses. This allows the browser to cache the tile for a day.

This header will also be used by Cloudfront to cache the tiles after the first user request is made. If we'd like Cloudfront to have a separate cache duration than browsers, we can add `, s-maxage=CLOUDFRONT_CACHE_DURATION` to the end of the `Cache-Control` header. See [here for more details](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html#ExpirationDownloadDist).

The cache duration is currently set to be a day, but this can be changed on feedback. This can also be made into an environment variable if we'd like to be able to change this at run time.

Also add some configuration for Cloudfront that ensures it caches the tiles generated by a user request.

Request review from both app and ops teams since this touches both aspects. Need eyes on the Terraform configuration, which was updated [according to this](https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html#ordered_cache_behavior).

~Did not connect to #763 or #764 since this does not include S3 work.~

Connects #779 

## Demo

image

## Testing Instructions

* Check out this branch and ensure that the vector tile flag is turned on `manage waffle_switch vector_tile on`
* Run `server` and go to [:6543/](http://localhost:6543/)
* Open the developer tools and ensure that "Disable Cache" is unchecked
* Observe the vector tile responses. Wait for them to finish.
* Refresh the page. Ensure the browser uses cached responses the second time.
* Update a facility to change the cache key. Or artificially change it.
* Refresh the page. Ensure the browser fetches fresh tiles.

## Checklist

- [ ] CI passes after rebase


deployment/terraform/cdn.tf
```diff
@@ -61,6 +61,27 @@ resource "aws_cloudfront_distribution" "cdn" {
...
   }
 
+  ordered_cache_behavior {
+    path_pattern     = "tile/*"
+    allowed_methods  = ["GET", "HEAD", "OPTIONS"]
+    cached_methods   = ["GET", "HEAD", "OPTIONS"]
+    target_origin_id = "originAlb"
+
+    forwarded_values {
+      query_string = false
```
hectcastro: I think that we want this to be `true`. With this set to `false`, query string parameters won't get included as part of CloudFront's cache key (e.g., `tile/1?q=foo` and `tile/1?q=bar` will be processed as `tile/1`).

deployment/terraform/cdn.tf
```diff
@@ -61,6 +61,27 @@ resource "aws_cloudfront_distribution" "cdn" {
     max_ttl                = 300
   }
 
+  ordered_cache_behavior {
+    path_pattern     = "tile/*"
+    allowed_methods  = ["GET", "HEAD", "OPTIONS"]
+    cached_methods   = ["GET", "HEAD", "OPTIONS"]
+    target_origin_id = "originAlb"
+
+    forwarded_values {
```
rajadain: We may need to change this part to respect this:

> Forward original referrer and custom header to the tile endpoint if at all possible.

from #764

rajadain: This has been addressed in d4cf98c.

deployment/terraform/cdn.tf
```diff
@@ -61,6 +61,28 @@ resource "aws_cloudfront_distribution" "cdn" {
...
+      query_string = false
+      headers      = ["*"]
```
hectcastro: Slightly similar comment here, except that I don't think we want to pass any headers through. Any differences in the values of headers across requests will cause objects to be cached separately within CloudFront. We need to evaluate all of the relevant headers and only includes ones (if any) in this list that we want to be cached separately.

deployment/terraform/cdn.tf
```diff
@@ -61,6 +61,28 @@ resource "aws_cloudfront_distribution" "cdn" {
...
+
+      cookies {
+        forward = "all"
```
hectcastro: It is possible that this setting should be changed to `none` or a specific set. When set to `all`, any values that differ between requests including the `Set-Cookie` header will get cached as separate objects.

rajadain:
Made edits to the Cloudfront configuration as recommended. Also pushed up `test/tt/tile-cache` branch which should be deployed to Staging soon.

kellyi:
> Did not connect to #763 or #764 since this does not include S3 work.

We should just rewrite one of these issues to account for the points here, or make a new 2-3 point issue which supersedes one of those. This work solves the issue even if the technical solution was different than what was originally imagined.

deployment/terraform/cdn.tf
```diff
@@ -61,6 +61,28 @@ resource "aws_cloudfront_distribution" "cdn" {
...
 
+  ordered_cache_behavior {
+    path_pattern     = "tile/*"
+    allowed_methods  = ["GET", "HEAD", "OPTIONS"]
+    cached_methods   = ["GET", "HEAD", "OPTIONS"]
+    target_origin_id = "originAlb"
+
+    forwarded_values {
+      query_string = true
+      headers      = ["Referer", "X-OAR-Client-Key"]
```
hectcastro: I have not found anything to suggest that leaving these headers out of this list will prevent them from getting passed to the origin on cache misses. Do we want different objects to be cached based on referer and client API key? If there isn't anything user specific about the vector tiles, can different users see the same object?

src/django/oar/settings.py
```diff
@@ -334,6 +334,7 @@
 
 # Application settings
 MAX_UPLOADED_FILE_SIZE_IN_BYTES = 5242880
+TILE_CACHE_MAX_AGE_IN_SECONDS = 60 * 60 * 24
```
hectcastro: I'm OK with this number too. I may have started with something lower (5-60 minute range), but in the end we'll have to evaluate based on the cache hit rates we see on the distribution as real traffic hits the tile endpoint. Important part is that we have a non-TTL based way to bust the cache.

deployment/terraform/cdn.tf
rajadain: Ah, I thought that this was required to get the headers to the origin. We don't need to be invalidating on these. If it is the case that the origin gets all headers on a cache MISS, then I'll set this to `none`.

hectcastro: Might be good to test real quick with the `test/` branch just to make sure. I was trying to find something to confirm my expectation that it would still go through, and even though I'm pretty sure it will, I didn't find anything definitive.

jwalgran: >Do we want different objects to be cached based on referer and client API key? If there isn't anything user specific about the vector tiles, can different users see the same object?
No to both. Our tiles are not user specific.

These headers are used to make it difficult for code not served from our domain to make unauthenticated API requests. We would like to extend our protection to tiles to prevent other web maps hotlinking to our tile layers.

src/django/oar/settings.py
jwalgran: We can bust the cache by incrementing a "version" column in the database.

jwalgran:
>We should just rewrite one of these issues to account for the points here, or make a new 2-3 point issue which supersedes one of those. This work solves the issue even if the technical solution was different than what was originally imagined.

I agree that we should close #763 and #764, write up a new issue that covers the work in this PR and supersedes those 2 issues, and swap in some cards from the priority backlog.

deployment/terraform/cdn.tf
rajadain: > These headers are used to make it difficult for code not served from our domain to make unauthenticated API requests.

Ah, so in this case we _should_ have changes in those headers invalidate the cache, so that they can be sent to the back-end and be 400'd.

jwalgran: Those headers don't need to be involved in the caching process. We just need to make sure that they are always forwarded to the origin.

rajadain: If I'm reading @hectcastro's comments correctly, in case of a cache MISS, all headers are always forwarded to the origin, regardless of this setting. What this setting does is attach these values to the cached version. So a subsequent request with the same value for these two headers will get a cache HIT, but one with different ones will get a MISS and be forwarded to the origin. If we don't specify any headers, then all headers will be ignored during caching. This would allow a certain degree of hotlinking, as while uncached tiles would not be able to be requested, cached tiles could (because Cloudfront would not check the headers before serving the cached tile). To truly make hotlinking difficult, we _should_ include these headers here, so that if their value differs, it causes a fresh request to the origin, which can decide if it should be 200 or 400.

src/django/api/views.py
```diff
@@ -2287,6 +2288,7 @@ def get_parent_company_options(self, request):
 @api_view(['GET'])
 @permission_classes([AllowAny])
 @renderer_classes([MvtRenderer])
+@cache_page(settings.TILE_CACHE_MAX_AGE_IN_SECONDS)
```
jwalgran: Searching for `cache_page` in the Django docs you end up on https://docs.djangoproject.com/en/2.2/topics/cache/, which discusses the caching framework of which `cache_page` is a part. Based on that documentation an the comments in https://github.com/django/django/blob/ce97960e34baf7db865b5424d3c61274a9f88919/django/views/decorators/cache.py#L8 I think this decorator caches the object using Django machinery, which is not what we want.

The docs say:
>Local-memory caching
>
>This is the default cache if another is not specified in your settings file.

This leads me to believe that when we use the `cache_page` decorator we are be putting tiles into a local memory cache.

It looks like https://docs.djangoproject.com/en/2.2/topics/http/decorators/#django.views.decorators.cache.cache_control may be more appropriate for what we are trying to do. 


````

Comment (rajadain) [reply src/django/api/views.py]: Ah! That's a great catch. Yes, `cache_control` is more what we want. Will switch to it.

File: src/django/oar/settings.py (hunk 6 above)
Comment (rajadain) [reply src/django/oar/settings.py]: Since allow query forwarding in Cloudfront configuration, we can also bust the cache using query strings.

Comment (rajadain) [reply src/django/oar/settings.py]: (But yes in practice the Cache Key which is a function of database version number will likely be used to request a new URI for the same tile)

**Comment by rajadain** [comment]
Alright I've pushed up a version that addresses all the concerns so far. Ready for another look.

File: src/django/api/views.py (hunk 7 above)
Comment (hectcastro) [reply src/django/api/views.py]: Nice catch.

File: deployment/terraform/cdn.tf [hunk 8]
```diff
@@ -61,6 +61,28 @@ resource "aws_cloudfront_distribution" "cdn" {
...
 
+  ordered_cache_behavior {
+    path_pattern     = "tile/*"
+    allowed_methods  = ["GET", "HEAD", "OPTIONS"]
+    cached_methods   = ["GET", "HEAD", "OPTIONS"]
+    target_origin_id = "originAlb"
+
+    forwarded_values {
+      query_string = true
+      headers      = ["Referer"]
```
hectcastro: It is unclear why `Referer` is being kept in this list. If because we want to distinguish between staging and production, each will have its own CloudFront distribution.

rajadain: The reasoning is that if someone is hotlinking to this layer from another site, the Referer field will be `data-pirates.xyz`, which would not get the cached tile, but go straight to origin, which will 401 the request. Otherwise, they would get the cached tiles, so the hotlinking would partially work (only for the already cached tiles, won't be able to request new ones).

hectcastro: Ah, OK. 👍

rajadain: I will also add a comment to that effect so it's clearer going forward

rajadain:
Given the wording in the newly written #779, I think I should change the default cache duration from 1 day to 1 year. We seem to have pretty solid cache invalidation in place, and Cloudfront will kick-out least recently used items on its own, so there's no need to set as conservative a duration as a single day.

rajadain:
Changed to caching for a year in the latest version. Ready for another look now.

jwalgran:
Looking at this now.

jwalgran:
> Do we want to disable the cache control headers during development? [...] we can always use the  "Disable Cache" checkbox in developer tools to ignore the value, so may  not be required.

As suggested, I think it is easy enough to use that checkbox, or comment out the decorator if caching needs to be disabled for debugging purposes.

rajadain:
Thanks for the thorough review and discussion! Rebased on develop to resolve merge conflicts in the CHANGELOG. Will merge once green.

Issue #758, opened 2019-09-02
The OAR map is frequently returning an error message when scrolling through a set of results


katieashaw:
## Overview

When scrolling through a set of results on the left hand side of the site, the map is frequently returning an error message. 

### Expected Behavior

The map should remain steady / constant whilst scrolling through a list of search results.

### Actual Behavior

The map produces an error message - see screenshots. 

### Steps to Reproduce

* Run a search on the OAR
* Scroll through results

### Demo

![Screenshot 2019-09-02 at 12 59 37](https://user-images.githubusercontent.com/46344729/64128634-62ad3480-cdaf-11e9-8e50-1ac965dc7e43.png)

![Screenshot 2019-09-02 at 17 35 07](https://user-images.githubusercontent.com/46344729/64128643-693bac00-cdaf-11e9-93a0-9b1e41283f25.png)

![Screenshot 2019-09-02 at 17 40 19](https://user-images.githubusercontent.com/46344729/64128649-6fca2380-cdaf-11e9-8c70-8278c70168b7.png)

### Additional context

This appears to be a new bug since the deployment of version 2.10.0 of the tool. It happened several times whilst I was doing a big piece of moderation work (around 4.5 hours or so of work).

dboyer:
I also saw this issue on production when viewing the map in Firefox on a Macbook. I can't replicate a reliable set of steps that always causes it, however. It's some combination of scrolling, zooming, and clicking on assets.

kellyi:
I believe this is related either to react-infinite-any-height or to the new Leaflet map implementation. I was able to trigger it with the new vector tiles map on staging, too.

jwalgran:
Sidebar appears to continue working after the error. If we can find a solution quickly, catching and logging a warning to Rollbar may be a way to fix production.

kellyi:
Hypothesis: the MapAndSidebar component encompasses both the map and the infinite scroll component, and I believe has an ErrorBoundary above it. If the infinite scroll throws a non-fatal error, it could be that the ErrorBoundary is catching the infinite scroll error and setting the fallback for broken map states.

kellyi:
I was able to re-create this using the vector tile feature infinite scroll on staging. The specific error is:

https://reactjs.org/docs/error-decoder.html/?invariant=185

kellyi:
[...]

PR #777, opened 2019-09-06
Fix infinite scroll map bug

kellyi:
## Overview

The map and the sidebar components shared a common error boundary, which had special rules for rendering the fallback "Map can't be displayed!" component whenever any error got emitted within that tree. As a consequence, errors emitted from the sidebar -- including those from the infinite list component -- would cause the map component to get into a failure state, even if the error was not fatal to the sidebar.

To split those errors, we now wrap the sidebar in its own separate error boundary which will catch errors before they get propagated to the entire `MapAndSidebar` component. In some cases these may still be fatal to the filter sidebar -- and the app in general -- but they won't immediately crash the map. If these errors are fatal to the whole app, we can spend some time later figuring out how to recover from them.

Connects #758 

## Demo

Screen Shot 2019-09-06 at 5 15 53 PM

## Notes

I was able to reproduce the error locally by loading all local facilities, then just scrolling back and forth through the list rapidly. This can/will eventually cause the following error, which is not fatal the first time it is encountered, at least:

Screen Shot 2019-09-09 at 1 06 09 PM

Setting the `key` attribute for the relevant routes to be `this.state.hasError` will force rerender on the error happening the first time, since the value will change. If we want it to catch errors subsequent to the first time, we could also set this value to be an incrementing counter.

Rationale for just having it re-render once is that we probably do want to know if it's a repeated error.

## Testing Instructions

- make sure you've got the latest version of React devtools installed
- serve this branch with the vector tile feature on
- open the React devtools and click the `hasError` prop for the new `MapAndSidebarFilterErrorBoundary` to set it true. Note that the app still continues to work
- refresh page, then try scrolling rapidly back and forth through the facilities list until you see the error happen. Verify that the map no longer crashes

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Fix a bug which prevented the facility claims dashboard page header from loading [#778](https://github.com/open-apparel-registry/open-apparel-registry/pull/778)
 - Fix a bug which prevented the vector tile marker layer from rendering [#782](https://github.com/open-apparel-registry/open-apparel-registry/pull/782)
+- Capture infinite-scroll emitted bug in an ErrorBoundary in order not to crash the map component [#777](https://github.com/open-apparel-registry/open-apparel-registry/pull/777)
 
 ### Security
 - Fix several npm security vulnerabilities via GitHub dependabot [#792](https://github.com/open-apparel-registry/open-apparel-registry/pull/792)
--- a/src/app/src/components/MapAndSidebar.jsx
+++ b/src/app/src/components/MapAndSidebar.jsx
@@ -2,8 +2,7 @@ import React, { Component, Fragment } from 'react';
 import { Route, Switch } from 'react-router-dom';
 import Grid from '@material-ui/core/Grid';
 
-import FilterSidebar from './FilterSidebar';
-import FacilityDetailsSidebar from './FacilityDetailSidebar';
+import SidebarWithErrorBoundary from './SidebarWithErrorBoundary';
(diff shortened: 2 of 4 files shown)


src/app/src/components/FilterSidebarFacilitiesTab.jsx
```diff
@@ -273,10 +273,10 @@ function FilterSidebarFacilitiesTab({
                                         oar_id: oarID,
                                     },
                                 }) => (
-                                    <Fragment key={oarID}>
+                                    <Fragment key={`${oarID}-${address}-${name}`}>
```
jwalgran: > OAR IDs should already be unique, but this change also adds the name and address as parts of the React key for the rows shown in the infinite list.

Consider expanding this commit message to explain why we are adding values to the key if the OAR ID is already unique.

kellyi: I think I'll just drop this commit -- thinking about it again, it's not really doing anything here.

src/app/src/components/MapAndSidebarFilterErrorBoundary.jsx
```diff
@@ -0,0 +1,50 @@
...
+import {
+    facilityDetailsRoute,
+    facilitiesRoute,
+    mainRoute,
+} from '../util/constants';
+
+import FacilityDetailsSidebar from './FacilityDetailSidebar';
+import FilterSidebar from './FilterSidebar';
+
+export default class MapAndSidebarFilterErrorBoundary extends Component {
```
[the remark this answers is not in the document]
kellyi: That makes sense. I'll rename this to be `SidebarWithErrorBoundary.jsx` which conveys what it is and does in the name.

PR #782, opened 2019-09-09
Fix bug which prevented marker layer from loading

kellyi:
## Overview

Previous work to encode the querystring into the tile cachekey mixed up
some function arguments. In turn this prevents the vector tile
facilities layer from loading. Additionally: the function extracted to
create the tileURL was unconditionally requesting the facility grid
layer, which would cause the markers layer to load as a grid, too.

This change fixes the bug so that the marker layer will render
correctly.

Connects #762 

## Testing Instructions

- load the app on this branch and verify that the grid layer and marker layer both render as expected

## Checklist

- [ ] CI passes after rebase


kellyi:
Cool! Thanks for catching that and reviewing this so quickly!

Issue #735, opened 2019-08-16
Restrict permissions on `/tile` endpoint


kellyi:
## Overview

https://github.com/open-apparel-registry/open-apparel-registry/pull/730 adds a `/tile` endpoint that currently has `AllowAny` permissions. However, we want to restrict things such that only the OAR's React client can make requests to this endpoint.

### Describe the solution you'd like

https://github.com/open-apparel-registry/open-apparel-registry/pull/730 adds a `/tile` endpoint that currently has `AllowAny` permissions. However, we want to restrict things such that only the OAR's React client can make requests to this endpoint.

Outcome of this should be that:

- the React client can consume the endpoint successfully
- ~no other client can successfully use it~ it's difficult for other clients to use it; they cannot just drop the URL in as the map endpoint and have it work
- there is an automated test that demonstrates this working

jwalgran:
We should be able to reuse the "client key" and request header validation we implemented to protect unauthenticated requests to to the rest of the API

https://github.com/open-apparel-registry/open-apparel-registry/blob/3e4a64ba1cfba3400e050b7d2248afe7d011f90f/src/django/api/permissions.py#L27-L49

We may just need to factor out the "web client" bits into a new `BasePermission` subclass and then decorate our tile view with it.

rajadain:
Unfortunately, it seems non-trivial to configure `VectorGrid` to specify header values in its requests, which only supports adding query parameters by default: http://leaflet.github.io/Leaflet.VectorGrid/vectorgrid-api-docs.html. Should we invest the time to make this possible, or would it be enough to only have a Referer check for now?

kellyi:
@jwalgran should confirm but I think referrer check might be enough.

It appears that the protobuf class of the VectorGrid library -- http://leaflet.github.io/Leaflet.VectorGrid/vectorgrid-api-docs.html#vectorgrid-protobuf -- also does afford passing along a `token` or `key` property with the tile requests but it also seems like these values can be seen just as easily as the tile URL when using the developer tools, so spending time adding it there wouldn't really do anything except make it possible to switch out the key on a redeploy. That may be enough but I dunno. Additionally, putting the key in the URL like that would affect the tile caching.

PR #791, opened 2019-09-09
Restrict `/tile` Endpoint to Allowed Hosts Only

rajadain:
## Overview

To discourage tile hotlinking, this adds a check of the HTTP referer field to allow only those in `ALLOWED_HOSTS`.

This reuses previously created methods for checking the referer field. Leaflet VectorGrid extension doesn't support making requests with custom headers, which is why we're not also reusing the `X-OAR-Client-Key` check here.

Also adds some tests for these new permissions.

Connects #735 

## Demo

image

## Testing Instructions

* Check out this branch and visit [:6543/](http://localhost:6543/)
  - [ ] Ensure you can still see the tiles

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/tests.py
+++ b/src/django/api/tests.py
@@ -3378,3 +3378,24 @@ class LogDownloadTests(APITestCase):
         log = DownloadLog.objects.first()
         self.assertEqual(expected_path, log.path)
         self.assertEqual(expected_record_count, log.record_count)
+
+
+class TilePermissionsTest(APITestCase):
+    def setUp(self):
+        self.tile_path = reverse('tile', kwargs={
+            'layer': 'facilitygrid',
+            'cachekey': '1567700347-1-95f951f7',
+            'z': 6, 'x': 15, 'y': 29,
+            'ext': 'pbf',
+        })
+
+    @override_settings(ALLOWED_HOSTS=['testserver', '.allowed.org'])
+    @override_switch('vector_tile', active=True)
+    def test_allowed_hosts_can_fetch_tiles(self):
+        response = self.client.get(self.tile_path, {},
+                                   HTTP_REFERER='http://allowed.org/')
+        self.assertEqual(200, response.status_code)
+
+    def test_disallowed_hosts_cannot_fetch_tiles(self):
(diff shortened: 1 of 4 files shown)


jwalgran:
Looking at this now.

src/django/api/tests.py
```diff
@@ -3378,3 +3378,25 @@ def test_creates_record(self):
...
+    def setUp(self):
+        self.tile_path = reverse('tile', kwargs={
+            'layer': 'facilitygrid',
+            'cachekey': '1567700347-1-95f951f7',
+            'z': 6, 'x': 15, 'y': 29,
+            'ext': 'pbf',
+        })
+
+    @override_settings(ALLOWED_HOSTS=['testserver', '.allowed.org'])
+    def test_allowed_hosts_can_fetch_tiles(self):
```
jwalgran: Consider using the `@override_switch` decorator so this test can assert a 200 response.

```diff
diff --git a/src/django/api/tests.py b/src/django/api/tests.py
index 7b928ef..2a79c9a 100644
--- a/src/django/api/tests.py
+++ b/src/django/api/tests.py
@@ -3390,12 +3390,11 @@ class TilePermissionsTest(APITestCase):
         })
 
     @override_settings(ALLOWED_HOSTS=['testserver', '.allowed.org'])
+    @override_switch('vector_tile', active=True)
     def test_allowed_hosts_can_fetch_tiles(self):
         response = self.client.get(self.tile_path, {},
                                    HTTP_REFERER='http://allowed.org/')
-        # Expect a 404 instead of 200 because the test database does not
-        # have vector tiles loaded
-        self.assertEqual(404, response.status_code)
+        self.assertEqual(200, response.status_code)
 
     def test_disallowed_hosts_cannot_fetch_tiles(self):
         response = self.client.get(self.tile_path)

```

rajadain: Thanks, this is exactly what was missing!

rajadain:
Thanks for reviewing and for suggesting `override_switch`. It looks much cleaner with it. Will merge when green.

Issue #768, opened 2019-09-05
Incorportate grid layer symbology feedback


kellyi:
## Overview

Experiment with different methods of aggregating symbolizing facilities at low zoom levels.

- Show empty grid cells with a faint border. Clip them to land masses.
- Use circles instead of squares
- Use different color ramps
- Use a hex grid instead of a square grid

### Is your feature request related to a problem? Please describe.

Early feedback on the grid layer is that it is non-intuitive.

### Describe the solution you'd like

Experiment and implement an improved grid layer

kellyi:
- don't display the grid legend when we're past the maxzoom level for the grid -- and into the marker layer zoom levels

PR #755, opened 2019-08-29
Add facility grid layer

jwalgran:
## Overview

Add a layer to show facilities aggregated by a fishnet grid at low zoom levels.

Connects #741

## Notes

The first iteration of this used a generic "wrapping" approach where the exact query produced by the ORM was joined to a synthesized table of grid polygons. This is the safest option, but each tile query took multiple seconds on staging dues to the query planner falling back to full sequential table scans.

`JOIN`ing the `api_facility` table directly in the tile query provided a 5x-10x boost in query performance at the cost of making the tile query construction more brittle.

## Testing Instructions

* Verify that the grid tiles load quickly.
* Verify that clicking a grid cell zooms the map to the extent of that cell.
* Verify that the map switches to showing the marker layer at zoom levels greater than 11.

## Checklist

- [ ] `fixup!` commits have been squashed
- [ ] CI passes after rebase
- [ ] CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines

PR #797, opened 2019-09-10
Redesign facility grid layer

jwalgran:
## Overview

Incorporate design feedback into a revised facility grid layer that uses a hexagonal grid and circular symbology rather than a square grid.

Connects #768

## Notes

We implemented the first grid layer to aggregate facilities at low zoom levels in #755. The general consensus was that the grid was revealing interesting patters in the data, but had some problems.

- The sparseness of the grid made the layer look broken, as if it was not finished loading.
- It was not obvious that the grid squares were interactive.
- Distortion of the squares at very low zoom levels.

Some potential modifications were discussed.

  - Show empty grid cells with a faint border. Clip them to land masses.
  - Use circles instead of squares.
  - Use different color ramps.
  - Use a hex grid instead of a square grid.

I added the empty grid cells with a faint border since it was quick to do, but did not go to the trouble of loading land mass boundaries and modifying the query to clip the data. It looked very busy just over land and did not address the other shortcomings of the grid approach.

Implementing circular symbology was also relatively straightforward and produced much more pleasing results.

Screen Shot 2019-09-06 at 1 48 27 PM
 
In order to remove as much of the gap between the circles as possible I researched generating a hex grid rather than a square grid and found [an easy-to-use postgres function](https://gist.github.com/mjumbewu/1761802ea06fb78c596f9cf8c9b2e769) to build it. Using a function also eliminates an inefficiency in our grid approach, where we generated and sent a large SQL string to construct the grid for each tile. The tradeoff for using the hex grid was the need to calculate the bounds of each feature rather than use a ZXY triple, but the benefits far outweighed the cost.

## Testing Instructions

* Run `./scripts/update` to ensure the required migration is run
* Run `./scripts/manage dbshell` and verify the current version number. 
```
openapparelregistry=# SELECT version FROM api_version;
 version
---------
       8
```
* Run `./scripts/manage incrementtileversion` to ensure the cache is invalidate.
* Run `./scripts/manage dbshell` and verify the current version number has been incremented. 
```
openapparelregistry=# SELECT version FROM api_version;
 version
---------
       9
```
* Browse the and verify that grid circles display a color and size relative to the number of points they represent.
* Verify that clicking a circle zooms to the extent.
* Verify that the grid is filtered properly when searching. 

## Checklist

- [x] CI passes after rebase


jwalgran:
>Just as a placeholder we may want another card to redesign the legend. 

Good idea. I created #798 to cover a redesign and hiding the legend when the grid layer is not visible.

Issue #757, opened 2019-09-02
The zoom on the OAR map is behaving in an inconsistent manner


katieashaw:
## Overview

The zoom on the OAR map is behaving in an inconsistent manner, in that sometimes it zooms right into the detail of a facility profile a user is viewing, and other times it does not. 

### Expected Behavior

The map zoom should be more consistent and always zoom into the detailed view of an OAR facility a user is looking at. 

### Actual Behavior

Sometimes the map zooms right into the detail of a facility, other times it does not. 

### Steps to Reproduce

* Search multiple different facilities in the tool and note the inconsistencies in the way the zoom is performing. 

### Demo

Zoom working:

![Screenshot 2019-09-02 at 17 00 48](https://user-images.githubusercontent.com/46344729/64125862-7783cb00-cda3-11e9-8dbf-6b4521ef6e86.png)

Zoom not performing as it should:

![Screenshot 2019-09-02 at 17 00 43](https://user-images.githubusercontent.com/46344729/64125871-7fdc0600-cda3-11e9-9481-69a73026702c.png)

Zoom not performing as it should:

![Map not zooming bug](https://user-images.githubusercontent.com/46344729/64125885-8bc7c800-cda3-11e9-84c2-711de76fc7fc.png)

### Additional context

The first two screenshots are taken today, subsequent to changes being rolled out in version 2.10.0 of the OAR, whilst the third screenshot was taken from version 2.9.0 of the app.

katieashaw:
Zoom not performing as it should again: 

![Screenshot 2019-09-02 at 17 00 48](https://user-images.githubusercontent.com/46344729/64126249-fcbbaf80-cda4-11e9-9e74-e2607fd59d0b.png)

katieashaw:
Zoom performing as it should be:

![Screenshot 2019-09-02 at 17 14 24](https://user-images.githubusercontent.com/46344729/64126290-2543a980-cda5-11e9-98f2-41ebc6a53da5.png)

katieashaw:
Zoom not behaving correctly x2:

![Screenshot 2019-09-03 at 11 17 14](https://user-images.githubusercontent.com/46344729/64165197-7b176080-ce3c-11e9-99dc-7f719c9a82bb.png)

![Screenshot 2019-09-03 at 11 17 19](https://user-images.githubusercontent.com/46344729/64165215-85d1f580-ce3c-11e9-9a40-b0d27d66f68f.png)

[...]

PR #799, opened 2019-09-10
Regularize map zoom and pan behavior

kellyi:
## Overview

Extract a useUpdateLeafletMapImperatively hook, then use it to adjust
the map components to have the following behavior on selecting a new
facility:

For the non-vector-tile FacilitiesMap:

- [x] if the user clicks on a facility marker, do not pan & zoom to the
facility
- [x] if the user arrives at the map directly from a URL containing an OAR
ID, set the map view on the facility location, zoomed to level 15
- [ ] if the user clicks on a facility in the sidebar, do not pan & zoom to
the facility unless the selected facility is outside of the current map
viewport bbox when the details data returns

For the vector tile map:

- [x] if the user clicks on a facility marker, do not pan & zoom to the
facility
- [x] if the user arrives at the map directly from a URL containing an OAR
ID, set the map view on the facility location, zoomed to level 15
- [x] if the user clicks on a facility in the sidebar and the selected
facility is outside of the bbox OR the zoom level is
within the range in which the facilities grid layer is shown, pan to the
facility and set the zoom level at 15. This ensures that the facility
details marker will be visible even if the map was at grid-layer zoom
levels before

Connects #757 

## Testing Instructions

- serve this branch and use the app both in vector tile and non-vector tile modes to verify that it works according to the cases described above and that it addresses the spirit of #757 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Close Disambiguation Popup on Reset [#812](https://github.com/open-apparel-registry/open-apparel-registry/pull/812)
 - Change "Accept" to "Confirm" in About/Processing [#815](https://github.com/open-apparel-registry/open-apparel-registry/pull/815)
+- Regularize vector tile map zoom behavior [#799](https://github.com/open-apparel-registry/open-apparel-registry/pull/799)
 
 ### Security
 
--- a/src/app/src/components/FilterSidebarFacilitiesTab.jsx
+++ b/src/app/src/components/FilterSidebarFacilitiesTab.jsx
@@ -280,7 +280,12 @@ function FilterSidebarFacilitiesTab({
                                             style={facilitiesTabStyles.listItemStyles}
                                         >
                                             <Link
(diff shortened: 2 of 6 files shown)


jwalgran:
Looking at this now.

kellyi:
Oof, thanks for catching those discrepancies. I rebased this on develop and will work out the remaining issues.

jwalgran:
Looking at this now.

jwalgran:
With the vector grid off, when I select a facility from the sidebar the markers start blinking

jwalgran:
I can't reliably reproduce the blinking issue. I have not seen it in Chrome. In Firefox the following steps sometimes trigger the blinking. It does not happen every time.

- In Firefox...
- Open a new tab to http://localhost:6543/
- Click "Search"
- Click the first item in the sidebar.

kellyi:
We discussed this but:

I decided to drop the changes to the existing FacilitiesMap and leave it in its current state. However, I kept the custom hook to regularize the vector tile map's zoom behavior which should still work as it did in the last test.

jwalgran:
Looking at this now.

kellyi:
Cool! Thanks for your help with this.

PR #812, opened 2019-09-12
Close Disambiguation Popup on Reset

rajadain:
## Overview

Ensures the disambiguation popup closes when the search is reset for the Cluster (non-Vector Grid) map.

Connects #796 

## Testing Instructions

* Check out this branch and turn off the vector tile switch `manage waffle_switch vector_tile off`
* Go to [:6543/](http://localhost:6543/) and search for `Saitex`
* Zoom in to the results on the map and click on the markers until you see a disambiguation pop-up
* Go back to the "Search" tab and click `Reset`
  - [x] Ensure the disambiguation pop-up closes

## Checklist

- [ ] `fixup!` commits have been squashed
- [ ] CI passes after rebase
- [x] CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines

Issue #796, opened 2019-09-10
Open pop-ups on map don't clear with a new search


dboyer:
## Overview

A pop-up displayed on the map from a previous search does not clear when a new search is run and there is no way to get rid of the box. 

### Expected Behavior

The pop-up box should either disappear when a new search is run or the X to close the box should always be visible and operational.

### Actual Behavior

The pop-up remains when a new search is run and it is impossible to pan up on the map to close the box. 

### Steps to Reproduce

Tested in Firefox on a Mac but reproduced in other browsers.

1. Search Saitex in the facility name field and view results on the map
2. Click the bubble with a 7 in it
3. Map will zoom in
4. Click the bubble with a 6 in it to get a pop-up
5. Click Reset in the search filters
6. Map resets but popup remains
7. Search toy in the facility name field and select AGT from the contributor list
8. Map loads search results but pop-up remains
9. Attempting to pan up the map to use the X to close the search box just causes the map to bounce back to a view where the top of the box is not visible

### Demo

![Screen Shot 2019-09-10 at 10 00 58 AM](https://user-images.githubusercontent.com/516764/64620470-ea97cd80-d3b1-11e9-8da7-9a231fc027b9.png)

### Additional context

Optional. Add any other context about the problem here.

katieashaw:
I experienced this issue in Chrome on my Mac, FYI.

PR #825, opened 2019-09-20
Close Disambiguation Popup on New Search

rajadain:
## Overview

Similar to 02bd8f3 and #812, this closes the disambiguation popup when a new search is performed. We use `facilities.fetching` as a proxy for telling when a new search has been fired.

Connects #796 

## Demo

In the old map:

In the new map:

## Testing Instructions

* Note if `vector_tile` waffle switch is on or off
* Go to the home page and search for "saitex"
* Click through to the markers until you see a disambiguation popup
* Switch the sidebar to search, and search for something else, _without clicking Reset_
  - [x] Ensure the disambiguation popup closes
* Toggle `vector_tile` and run the above test again
  - [x] Ensure the disambiguation popup closes

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/components/FacilitiesMap.jsx
+++ b/src/app/src/components/FacilitiesMap.jsx
@@ -73,6 +73,7 @@ const unselectedMarkerIcon = createIcon(unselectedMarkerURL);
 const selectedMarkerIcon = createIcon(selectedMarkerURL);
 
 function FacilitiesMap({
+    fetching,
     data,
     navigateToFacilityDetails,
     facilityDetailsData,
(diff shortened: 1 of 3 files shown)


src/app/src/components/FacilitiesMap.jsx
```diff
@@ -226,6 +227,13 @@ function FacilitiesMap({
         setFacilitiesToDisambiguate,
     ]);
 
+    useEffect(() => {
+        // Close multiple facilities popup on fresh searches
+        if (fetching) {
+            setFacilitiesToDisambiguate(null);
+        }
+    }, [fetching]);
```
kellyi: This is probably fine but it's interesting that the linter didn't complain about `setFacilitiesToDisambiguate` not being included in the dependencies array. In practice that function should never change during the component lifecycle, so maybe the linting rule is robust enough to know that too.

rajadain: Reading up here: https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies, I see:

> The identity of the `setCount` function is guaranteed to be stable so it’s safe to omit.

Where `setCount` was from `useState()`:

```js
const [count, setCount] = useState();
```

So I think we should be good.

rajadain:
Thanks for reviewing!

Issue #798, opened 2019-09-10
Change grid layer legend to match layer symbology / hide legend when markers are shown 


jwalgran:
## Overview

Use circles instead of squares in the legend and only show the legend when the grid layer is visible.

### Is your feature request related to a problem? Please describe.

Now that we have switched to showing circles instead of squares the legend symbology no longer matches the map.

The legend is potentially confusing when we switch to showing markers instead of the grid layer.

### Describe the solution you'd like

- Use circles instead of squares in the legend 
- Only show the legend when the grid layer is visible.

PR #826, opened 2019-09-23
Restyle grid layer legend & display conditionally

kellyi:
## Overview

- restyle grid layer legend to use circles rather than boxes for the
legend items
- display the grid layer legend only for zoom levels at which the grid
layer is shown

Connects #798 

## Demo

Screen Shot 2019-09-23 at 12 34 16 PM

## Testing Instructions

- serve this branch with vector tiles switched on and the default facilities data loaded
- verify that the legend now displays only at zoom levels for which the grid layer is also shown and that it disappears when the marker layer is displayed

## Checklist

- [ ] CI passes after rebase


jwalgran:
Looking at this now.

kellyi:
> Consider also showing the circle size changes in the legend. Maybe something like:

Sounds good! I'll add this before merging.

Issue #822, opened 2019-09-19
App crashes when logging in after zooming in


rajadain:
## Overview

If the vector tile layer is turned on, and you zoom in all the way and pick a facility, then click "Log In" in the top right, the app crashes.

Unhandled Rejection (TypeError): layer is undefined
https://github.com/open-apparel-registry/open-apparel-registry/blob/8b262ae7619b40045c1e720d23d8e93aa52f3941/src/app/src/components/VectorTileFacilityGridLayer.jsx#L97

### Expected Behavior

The app does not crash, and allows logging in.

### Actual Behavior

The app crashes.

### Steps to Reproduce

* Have the vector tile layer turned on
* Zoom all the way in to the facility level and select one
* Click "Log In" in the top right
* Wait a few seconds
* See the app crash

### Demo

![image](https://user-images.githubusercontent.com/1430060/65261402-321b0980-dad6-11e9-83bd-bca0fa0e3c66.png)

![2019-09-19 12 09 00](https://user-images.githubusercontent.com/1430060/65261647-b9687d00-dad6-11e9-9da6-32a777b20cc7.gif)

kellyi:
This should be pretty straightforward to fix and we'll need to do that before releasing vector tiles to production.

PR #837, opened 2019-09-30
Fix vector tile map crashes on login/logout

kellyi:
## Overview

Fix a bug whereby the vector tile map would crash on logging in or
logging out caused by a missing check to ensure that the map object
existed before calling getZoom. Adjust it to use the stored zoom state
instead.

Connects #822 

## Notes

Fixing this uncovered another quirk whereby we are clearing the existing facilities data on logout, along with the filter settings. This is how the existing map setup also works, so I did not try to address it here and will make another card for fixing it.

## Testing Instructions

- get this branch then sign in and turn on the vector tile map
- try to re-create the bug described in #822 and verify that it has been fixed

## Checklist

- [ ] CI passes after rebase


jwalgran:
Looking at this now.

Issue #819, opened 2019-09-18
Add facility associations to history


jwalgran:
## Overview

Depends on #766 

Add history events to the existing history endpoint that detail when contributors are associated or disassociated with a facility.

### Describe the solution you'd like

- Use the `HistoricalFacilityMatch` table to generate history events.
  - When a match row is created, create an "ASSOCIATE" type event that includes the ID and name of the contributor and the name of the list, obtained by linking back through the `FacilityListItem` to `FacilityList` to `Contributor`
  - When `is_active` changes to `False`, generate a history record with a `DISASSOCIATE` type that includes the ID and name of the contributor and the name of the list, obtained by linking back through the `FacilityListItem` to `FacilityList` to `Contributor`

PR #841, opened 2019-10-01
Include match associations and dissociations in facility history response

kellyi:
## Overview

- add facility match associations to facility history record list
- add facility match dissociations to facility history record list

Connects #819

## Testing Instructions

- serve this branch and make sure you've run the data migration from the history endpoint PR
- turn on the facility_history feature
- run the tests to verify that they pass (and that the new tests are testing the correct thing)
- run `./scripts/resetdb` then try out the facility history endpoint for a facility to verify that you see the new association record there
- also try removing a list item and verify that you see it entered as `DISSOCIATE` in the facility's history endpoint

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/facility_history.py
+++ b/src/django/api/facility_history.py
@@ -4,6 +4,54 @@ from api.constants import ProcessingAction, FacilityHistoryActions
 from api.models import FacilityMatch
 
 
+def create_associate_match_string(facility_str, contributor_str, list_str):
+    return 'Associate facility {} with contributor {} via list {}'.format(
+        facility_str,
+        contributor_str,
+        list_str,
+    )
+
+
+def create_associate_match_change_reason(list_item, facility):
+    return create_associate_match_string(
+        facility.id,
+        list_item.facility_list.contributor.id,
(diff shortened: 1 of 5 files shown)


src/django/api/facility_history.py
```diff
@@ -4,6 +4,54 @@
...
 
+def create_associate_match_string(facility_str, contributor_str, list_str):
+    return 'Associate facility {} with contributor {} via list {}'.format(
+        facility_str,
+        contributor_str,
+        list_str,
+    )
+
+
+def create_associate_match_change_reason(list_item, facility):
```
kellyi: This is a separate function because it appears that the change reason field is VARCHAR(100), which makes it impossible to serialize every list name. In practice, the changes here don't depend on the change_reason existing at all, so I'm happy to dump writing to changeReason at all.

jwalgran:
Looking at this now.

kellyi:
I rebased this on the branch from https://github.com/open-apparel-registry/open-apparel-registry/pull/830

Issue #818, opened 2019-09-18
Add claim history to endpoint


jwalgran:
## Overview

Add claim-related events to the existing history endpoint.

### Describe the solution you'd like

- Use the `HistoricalFacilityClaim` model to add history events
  - When `FacilityClaim.status` changes to `APPROVED` add a `CLAIMED` event that includes the contributor ID and name.
  - When any of the `FacilityClaim` profile fields change, add a `CLAIM_UPDATE` event with the field change details. Use [`diff_against`](https://django-simple-history.readthedocs.io/en/latest/history_diffing.html) to get the field changes.
  - If a claim is revoked, add a `CLAIM_REVOKE` event.

PR #844, opened 2019-10-02
Add facility claim data to history response

kellyi:
## Overview

Add claim data to history response including:

- claim was approved
- claim was revoked
- claim had some public details changed (excluding non-public and
potentially non-public fields)

Connects #818 

## Notes

Per our discussion I opted to use a whitelist of public claim update field changes to ensure we don't accidentally serialize something non-public. If demand is there for including potentially non-public fields, we can reassess the approach.

## Testing Instructions

- run tests on this branch and verify that they pass
- serve this branch, turn on all the feature flag switches, then:
   - claim a facility
   - approve the claim
   - make some changes
   - revoke the claim

Then use the Swagger docs to request the facility history and verify that you see each action in the response.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/constants.py
+++ b/src/django/api/constants.py
@@ -47,6 +47,8 @@ class FacilityHistoryActions:
     MERGE = 'MERGE'
     SPLIT = 'SPLIT'
     OTHER = 'OTHER'
+    ASSOCIATE = 'ASSOCIATE'
+    DISSOCIATE = 'DISSOCIATE'
 
 
 class Affiliations:
--- a/src/django/api/constants.py
+++ b/src/django/api/constants.py
@@ -49,6 +49,9 @@ class FacilityHistoryActions:
     OTHER = 'OTHER'
     ASSOCIATE = 'ASSOCIATE'
     DISSOCIATE = 'DISSOCIATE'
+    CLAIMED = 'CLAIMED'
+    CLAIM_UPDATE = 'CLAIM_UPDATE'
+    CLAIM_REVOKE = 'CLAIM_REVOKE'
 
 
 class Affiliations:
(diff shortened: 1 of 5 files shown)


src/django/api/views.py
```diff
@@ -789,7 +789,7 @@ def count(self, request):
     @transaction.atomic
     def claim(self, request, pk=None):
         if not switch_is_active('claim_a_facility'):
-            return NotFound()
+            raise NotFound()
```
kellyi: I noticed this mistake while I was writing tests against the claim endpoint.

src/django/api/constants.py
```diff
@@ -49,6 +49,9 @@ class FacilityHistoryActions:
     OTHER = 'OTHER'
     ASSOCIATE = 'ASSOCIATE'
     DISSOCIATE = 'DISSOCIATE'
+    CLAIMED = 'CLAIMED'
```
jwalgran: Our pattern as been to use a present tense action label. Consider `CLAIM`.

kellyi:
Closing in favor of #852

src/django/api/facility_history.py
```diff
@@ -162,6 +162,60 @@ def processing_results_has_split_action_for_oar_id(list_item, facility_id):
...
+            'facility_parent_company',
+        ]
+
+        return {
+            'updated_at': str(claim.updated_at),
+            'action': FacilityHistoryActions.CLAIM_UPDATE,
+            'detail': 'Facility {} claim public data was updated'.format(
+                claim.facility.id,
+            ),
+            'changes': {
```
jwalgran: If a non-public field is updated, the history record looks like this:

Screen Shot 2019-10-03 at 10 37 29 AM

This looks a bit broken. Consider suppressing the entire history entry if the changes dict is empty.

Issue #817, opened 2019-09-18
Refactor models to include Source


jwalgran:
## Overview

Introduce a new `Source` model that allows for the submission of a single facility. 

Notes: https://docs.google.com/document/d/1WScneGlu3xD_QqT7nV8umjqnMStCcVIXPvdzk5-oUy4

### Is your feature request related to a problem? Please describe.

We need to support submitting single items directly via a new API endpoint.

### Describe the solution you'd like

To maintain backward compatibility, this will require 3 releases

#### Release 1
- Add a `Source` table
  - Add a nullable (for now) FK from `FacilityListItem` to `Source`
  - Add a nullable FK from `Source` to `FacilityList`
  - Add fields
    - `source_type`: "LIST" or "SINGLE"
    - `is_active`: bool default True
    - `is_publice`: bool default True
    - `create`: bool default true
    - `created_at`
    - `updated_at`
- Add a management command to idempotently populate `Source` from `FacilityList`
  - For each `FacilityList`: 
    - If there isn't a record in `Source` with type "LIST" and a FK matching the
      `FacilityList.id`:
      - Create a row in `Source`.
        - Set type to "LIST"
        - Set the `FacilityList` FK.
        - Copy `contributor`, `is_private`, and `is_active` values from `FacilityList` to `Source`.
    - If there is a record in `Source`:
      - Copy `contributor`, `is_private`, and `is_active` values from `FacilityList` to `Source`. 
    - Also copy `created_at` and `updated_at` using queryset update.
    - For all all `FacilityListItem`s associated with the list set the FK to   `Source`.
- Deploy and migrate

#### Release 2
- Update the CSV upload handling to populate the `Source` table and the `FacilityListItem` FK to `Source`
- Update references to `FacilityList.contributor`, `FacilityList.is_public`, and `FacilityList.is_active` to use `Source` fields instead.
- Run the management command
  - Minimizes the number of `FacilityList` rows without a related `Source`
- Deploy
- Run management command

#### Release 3
- Remove the management command
- Add a migration
  - Make the FK from `FacilityListItem` to `Source` not null
  - Remove FK from `FacilityListItem` to `FacilityList`
  - Remove columns from `FacilityList` that appear on `Source`, which are no longer referenced by any deployed code.
    - is_active
    - is_public
    - contributor
- Deploy and migrate

PR #856, opened 2019-10-04
Implement Source model - Step 1

jwalgran:
## Overview

The first step of a three step process to add a Source model between facility lists and list items, allowing for the submission of individual facilities.  

Each step in the process is intended to be released separately and maintain backward compatibility.

Connects #817 

## Testing Instructions

* Checkout `develop` and `./scripts/resetdb`
* Checkout this branch and `./scripts/migrate`
* Ensure that `./scripts/manage liststosources` runs without error
* Use `./scripts/manage shell_plus` to verify that a `Source` is created for each `FacilityList` and that a `Source` is assigned on each `FacilityListItem`
```
>>> list(FacilityList.objects.all().order_by('id').values_list('id', flat=True))
[2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
>>> list(Source.objects.all().order_by('facility_list').values_list('facility_list', flat=True))
[2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
>>> FacilityListItem.objects.filter(source=None).count()
0
```

## Checklist

- [x] CI passes after rebase


jwalgran:
This has a migration conflict. Fixing now.

jwalgran:
Fixed the migration conflict and rebased.

jwalgran:
Thank you for the review.

>I don't have a clear sense of the larger design change that this is serving

This is the highest level view of what this set of pulls it trying to achieve.

Screen Shot 2019-10-04 at 12 53 09 PM

In order to allow submission of individual facilities we are inserting `Source` in between `Contributor` and `FacilityListItem` rather than `FacilitiyList`. The `FacilityList` table becomes a place where we store additional metadata about a `Source` if the the `Source` is a list.

#857 is where all of the major work happens. #858 is "cleanup." I will be assigning those for review shortly.

jwalgran:
Thanks for the reviews.

PR #859, opened 2019-10-04
Enable Waffle switches when running resetdb

jwalgran:
## Overview

We want all of these features to be on by default when testing because we are
most often going to want to exercises all the available features.

## Testing Instructions

* Run `./scripts/resetdb`
* Browse http://localhost:8081/admin/waffle/switch/ and verify that all the switches are enabled.

## Checklist

- [ ] CI passes after rebase


jwalgran:
Thanks for the review.

Issue #843, opened 2019-10-01
Duplicate other locations are shown


jwalgran:
## Overview

The other locations section of the facility detail page shows duplicate entries.

### Expected Behavior

Each other location is shown once.

### Actual Behavior

Other locations can be repeated when multiple matches have the same location.

### Steps to Reproduce

As of writing this issue you can see this in production here: https://openapparel.org/facilities/BD2019083JKZJM7 (screenshot below)

### Demo

<img width="563" alt="Screen Shot 2019-10-01 at 12 32 41 PM" src="https://user-images.githubusercontent.com/17363/65994219-ad21df80-e447-11e9-9a26-38d50c08a3db.png">

### Proposed solution

- Group the other location records by coordinates
  - Show each contributor only once under the coordinate
  - Compensate for rounding errors by using turf to calculate the distance between points and combining the points if they are within a reasonably small minimum distance.

PR #860, opened 2019-10-07
Remove duplicate entries from other locations data

kellyi:
## Overview

Remove duplicate entries from other locations data using the following
logic:

- if the lat or lng are different, an entry is not a duplicate
- if the lat and lng are the same but the contributor names/ids are
present but not the same, entry is not a duplicate (because we want to
surface that more than one contributor suggested the same location)
- if the lat and lng are the same but one of the entries does not have
contributor name/id data, the entry is a duplicate
- if the lat and lng are the same otherwise, entry is a duplicate

Connects #843 

## Testing Instructions

- get branch, then run `./scripts/test` to verify that the new test passes
- read the new test and verify that it is checking the correct things
- run the app and verify that the alternate locations field now removes duplicates according to the logic described above

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/util/util.js
+++ b/src/app/src/util/util.js
@@ -25,6 +25,7 @@ import toInteger from 'lodash/toInteger';
 import keys from 'lodash/keys';
 import pickBy from 'lodash/pickBy';
 import every from 'lodash/every';
+import uniqWith from 'lodash/uniqWith';
 import { isEmail, isURL } from 'validator';
 import { featureCollection, bbox } from '@turf/turf';
 import { saveAs } from 'file-saver';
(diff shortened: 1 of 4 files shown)


jwalgran:
Looking at this now.

src/app/src/util/util.js
```diff
@@ -633,3 +634,35 @@ export const pluralizeResultsCount = (count) => {
...
+        }
+
+        if ((!id && otherID) ||
+            (!name && otherName) ||
+            (id && !otherID) ||
+            (name && !otherName)) {
+            return true;
+        }
+
+        if (name === otherName || id === otherID) {
```
jwalgran: Do we need to also be checking the contributor name if we are checking IDs? There should be no case where the IDs match and the names do not.

Issue #850, opened 2019-10-03
AttributeError: 'NoneType' object has no attribute 'y'


jwalgran:
View details in Rollbar: [https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/286/](https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/286/)

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/viewsets.py", line 114, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 497, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 457, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 468, in raise_uncaught_exception
    raise exc
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 494, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/src/api/views.py", line 617, in retrieve
    response_data = FacilityDetailsSerializer(queryset).data
  File "/usr/local/lib/python3.7/site-packages/rest_framework/serializers.py", line 559, in data
    ret = super().data
  File "/usr/local/lib/python3.7/site-packages/rest_framework/serializers.py", line 261, in data
    self._data = self.to_representation(self.instance)
  File "/usr/local/lib/python3.7/site-packages/rest_framework_gis/serializers.py", line 132, in to_representation
    feature["properties"] = self.get_properties(instance, fields)
  File "/usr/local/lib/python3.7/site-packages/rest_framework_gis/serializers.py", line 157, in get_properties
    representation = field.to_representation(value)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/fields.py", line 1873, in to_representation
    return method(value)
  File "/usr/local/src/api/serializers.py", line 452, in get_other_locations
    .filter(is_active=True)
[...]

PR #861, opened 2019-10-07
Check geocoded_point is not None when serializing other locations

jwalgran:
## Overview

Prevents an exception from being raised in the serializer when a list item failed geocoding but was able to be matched to another facility.

Connects #850

## Testing Instructions

* Verify that the unit tests pass

## Checklist

- [x] CI passes after rebase


jwalgran:
Thanks for the review.

Issue #875, opened 2019-10-14
Replace history endpoint switch with a flag


jwalgran:
## Overview

We want to allow individual contributors to have access to the history endpoint.

### Is your feature request related to a problem? Please describe.

A switch only allows us to turn a feature on and off for everyone.

### Describe the solution you'd like

- Create a `can_get_facility_history` `Group` https://docs.djangoproject.com/en/2.2/topics/auth/default/#groups
- Create a `can_get_facility_history` `Flag` https://waffle.readthedocs.io/en/stable/types/flag.html.
  - Assign the `can_get_facility_history` `Group` to the new `Flag`.
- Do not allow use of the facility history endpoint unless the flag is enabled.

PR #881, opened 2019-10-17
Change facility history API endpoint switch to a flag activated via group

kellyi:
## Overview

Change the facility history API endpoint feature flagging mechanism from a waffle-switch to a waffle-flag which restricts access based on whether the user is a member of a `can_get_facility_history` Group.

Update tests and views to replace switch with flag, adding tests to check that the endpoint works as expected for various levels of authorization.

Un-hide facility history API endpoint from Swagger docs, as it will now be public but permission-ed.

Add a data migration to delete the previous `facility_history` switch.

Update django-admin to display Group objects.

Connects #875 

## Notes

Due to an override we have made to the django-admin User object interactions, there is as yet no mechanism for assigning users to groups outside of the command line. I don't know what the preferred way to handle this is, but two options are:

- update the custom User interface we wrote for the django-admin to allow adding users to groups
- add a section to the React dashboard that allows adding users to groups

## Testing Instructions

- get branch, then `./scripts/resetdb`
- `./scripts/test` and verify that they pass
- sign in as c1@example then use the history endpoint and verify that you get a 200 response
- sign in as c10@example then use the history endpoint and verify that you get a 403 response

## Checklist

- [ ] CI passes after rebase


rajadain:
Rebased on `develop` to fix conflicts. Doing a final review, will merge after.

jwalgran:
Please hold off on merging until after the 2.16.0 release is complete to prevent a staging deploy.

kellyi:
🏄

Issue #873, opened 2019-10-14
DRAFT: Add a private submission flag


jwalgran:
# Draft

This is a draft issue, and it will need to clarify specs and formalize what
counts as "done" before work begins on it.

Check with the issue's creator before starting work on this issue.

## Overview

A clear description of the feature request.

Depends on #820 

### Describe the solution you'd like

- Create a `can_submit_private_data` `Group` https://docs.djangoproject.com/en/2.2/topics/auth/default/#groups
- Create a `can_submit_private_data` `Flag` https://waffle.readthedocs.io/en/stable/types/flag.html.
  - Assign the `can_submit_private_data` `Group` to the new `Flag`.
- Do not allow use of the `private` query string argument on the single item submission endpoint.

### Additional Context

Optional. Add any other context, screenshots, gifs, etc about the feature request here.

Issue #820, opened 2019-09-18
Add single facility submission API endpoint.


jwalgran:
## Overview

Add an API endpoint that allows authorized contributors to POST the details of single facilities, match those details to the current list of facilities, and optionally create and return a new facility or a new facility match.

### Is your feature request related to a problem? Please describe.

MSIs want to integrate facility submission into their workflow.

### Describe the solution you'd like

- Depends on the `Source` table #817 	
- Depends on support for single matches (exploration in #702)
- Feature flagged with access managed by group membership ~(make sure it does not appear in the Swagger docs if the flag is off)~
- POST JSON to /api/facilities/ (country, name, address)
- `create=false` Query string argument prevents creating match/facility. Just returns possible matches with confidence. True by default.
- `public=false` Query string argument sets the `is_public` field on the source. Return 403 if the user does not have the private submission feature flag enabled (#873) Implementing this is not blocked on the feature flag being implemented.
- Returns array of ~`FacilityMatch` rows~ matching facility details along with the item status and confidence score.

PR #896, opened 2019-10-25
Add single facility submission endpoint

jwalgran:
## Overview

Add single facility submission endpoint.

Connects #820
Connects #872 
Connects #873 

### Match response

```js
{
  "matches": [
    {
      "id": "CN2019303BQ3FZP",
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [
          120.596047,
          32.172013
        ]
      },
      "properties": {
        "name": "Nantong Jackbeanie Headwear Garment Co. Ltd.",
        "address": "No. 808, The Third Industry Park, Guoyuan Town, Rugao City Nantong",
        "country_code": "CN",
        "oar_id": "CN2019303BQ3FZP",
        "other_names": [],
        "other_addresses": [],
        "contributors": [
          {
            "id": 4,
            "name": "Researcher A (Summer 2019 Affiliate List)",
            "is_verified": false
          }
        ],
        "country_name": "China",
        "claim_info": null,
        "other_locations": []
      },
      "confidence": 0.8153
    }
  ],
  "item_id": 964,
  "geocoded_geometry": {
    "type": "Point",
    "coordinates": [
      120.596047,
      32.172013
    ]
  },
  "geocoded_address": "Guoyuanzhen, Rugao, Nantong, Jiangsu, China",
  "status": "MATCHED",
  "oar_id": "CN2019303BQ3FZP"
}
```

### Potential match response

```js
{
  "matches": [
    {
      "id": "CN2019303BQ3FZP",
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [
          120.596047,
          32.172013
        ]
      },
      "properties": {
        "name": "Nantong Jackbeanie Headwear Garment Co. Ltd.",
        "address": "No. 808, The Third Industry Park, Guoyuan Town, Rugao City Nantong",
        "country_code": "CN",
        "oar_id": "CN2019303BQ3FZP",
        "other_names": [],
        "other_addresses": [],
        "contributors": [
          {
            "id": 4,
            "name": "Researcher A (Summer 2019 Affiliate List)",
            "is_verified": false
          }
        ],
        "country_name": "China",
        "claim_info": null,
        "other_locations": []
      },
      "confidence": 0.7686
    }
  ],
  "item_id": 959,
  "geocoded_geometry": {
    "type": "Point",
    "coordinates": [
      120.596047,
      32.172013
    ]
  },
  "geocoded_address": "Guoyuanzhen, Rugao, Nantong, Jiangsu, China",
  "status": "POTENTIAL_MATCH"
}
```

### New facility response

```js
{
  "matches": [],
  "item_id": 954,
  "geocoded_geometry": {
    "type": "Point",
    "coordinates": [
      119.2221539,
      33.79772
    ]
  },
  "geocoded_address": "30, 32 Yanhuang Ave, Lianshui Xian, Huaian Shi, Jiangsu Sheng, China, 223402",
  "status": "NEW_FACILITY"
}
```

### No match and geocode returned no results response

```js
{
  "matches": [],
  "item_id": 965,
  "geocoded_geometry": null,
  "geocoded_address": null,
  "status": "ERROR_MATCHING"
}
```

## Notes

We are introducing 2 new groups/waffle flags

- `can_submit_facility` controls access to the single-item submission endpoint.
- `can_submit_private_facility` allows passing a `public=false` query string argument.

Implementing the previously unused `create` action on the facilities resource is a natural fit for submitting a single facility record.

The endpoint requires POSTing a JSON body with `name`, `address`, and `country` fields. We reuse the existing validation function from list processing to ensure that the submitted `country` can be converted to acountry code.

In order to support quickly matching individual list items we needed a way of holding a trained and indexed model in memory and keeping it updated as the facility data changed. The new `GazetteerCache` class implements this by holding a trained model in a class variable and reading from the `HistoricalFacility` table to determine whether items need to be added or removed from the index.

The `GazetteerCache` includes a threading lock to ensure that only one thread can ever be updating the cached gazetteer.

Although the management command-based matching uses the new `GazetteerCache` the actual behavior has not changed, since starting a process from scratch will result in training a new model, as we have always done.

As part of this refactor we have also addressed a problem with single item CSV uploads causing the match process run out of memory. We address this by changing the way models are trained. Instead of using the input file, we use 20% of the submitted list items as our "messy" data.

The flow of the `create` view is based on the batch processing pipeline, but all three steps are completed at once. Each stage is wrapped in an individual try block so we can report unexpected failures similar to the batch process.

The "parse" step is combined with the saving of the initial `Source` and `FacilityListItem` records, since the data fields are submitted "pre-parsed." We create `Source` and `FacilityListItem` objects even when the user passes the `create=false` option because we still want to track the submission of the data in the `Source` table and the performance of the matching in the `FacilityListItem.processing_results` field.

We were able to reuse the `save_match_details` function used by the batch processing. The only change we needed to make was to wrap the model save code with a check of the `Source.create` property.

We needed to adjust the facility delete method because submitting high-confidence matches with `create=false` was creating `FacilityListItem` rows with foreign key references to the `Facility`, but not a corresponding `FacilityMatch` row.

### Setup / Regression test batch processing

- Run `./scripts/manage migrate` and `./scripts/resetdb`. Verify that `resetdb` matches models without error and that the dev data is matched correctly.
- Log in as `c8@example.com`, browse http://localhost:6543/lists, and verify that there are some pending matches.
- Submit and process [other-location-test.csv.zip](https://github.com/open-apparel-registry/open-apparel-registry/files/3789808/other-location-test.csv.zip). Verify that the steps complete without error
- Browse http://localhost:6543/lists/16 and verify that there are matches

### Test single item submission

- Still logged in as c8@example.com, browse "My Profile" and create an API token"
- On the Vagrant VM, add the token to your shell environemnt: `export OAR_API_TOKEN={token}`
- Post to the endpoint and verify that it returns 401 / requires auth.
```
http POST http://localhost:8081/api/facilities/?create=false 
```
- Post to the endpoint with auth and verify that it returns 403
```
http POST http://localhost:8081/api/facilities/?create=false \
"Authorization: Token $OAR_API_TOKEN"
```
- Run `./scripts/manage shell_plus` and add the user to the group that allows single-item submission
```
from django.contrib import auth
g = auth.models.Group.objects.get(name='can_submit_facility')
u = User.objects.get(email='c8@example.com')
u.groups.add(g)
u.save()
```
- Post to the endpoint and verify that it returns 400 because a JSON body was not included
```
http POST http://localhost:8081/api/facilities/?create=false \
"Authorization: Token $OAR_API_TOKEN" 
```
- POST with a body and verify a successful response with a status of `NEW_FACILITY`.
```
http POST http://localhost:8081/api/facilities/?create=false \
"Authorization: Token $OAR_API_TOKEN" \
<<< '{
  "country": "US",
  "name": "Azavea",
  "address": "990 Spring Garden Street, Philadelphia"
}'
```
- POST a near match and verify a successful response with a status of `POTENTIAL_MATCH`
```
http POST http://localhost:8081/api/facilities/?create=false \
"Authorization: Token $OAR_API_TOKEN" \
<<< '{
	"country": "China",
	"name": "Nantong Jackbeanie Headwear & Garment Co. Ltd.",
	"address": "No.808,the third industry park,Guoyuan Town,Nantong 226500."
}'
```
- POST an exact match and verify a successful response with a status of `MATCHED`
```
http POST http://localhost:8081/api/facilities/?create=false \
"Authorization: Token $OAR_API_TOKEN" \
<<< '{
	"country": "China",
	"name": "Nantong Jackbeanie Headwear Garment Co. Ltd.",
	"address": "No. 808, The Third Industry Park, Guoyuan Town, Rugao City Nantong"
}'
```
- POST without the `create=false` argument, verify a successful response, and verify that the "Azavea" facility is now searchable http://localhost:6543/facilities?q=azavea
```
http POST http://localhost:8081/api/facilities/ \
"Authorization: Token $OAR_API_TOKEN" \
<<< '{
  "country": "US",
  "name": "Azavea",
  "address": "990 Spring Garden Street, Philadelphia"
}'
```
- POST a match without `create=false`, verify a successful response, and verify that the contributor with `"id": 8` is listed in the `contributors` array.
```
http POST http://localhost:8081/api/facilities/ \
"Authorization: Token $OAR_API_TOKEN" \
<<< '{
  "country": "Bangladesh",
  "name": "AKH Shirts Ltd.",
  "address": "Savar Dhaka 1340 Bangladesh"
}'
````
- Attempt to POST a facility with `public=false` and verify that a 403 is returned
```
http POST http://localhost:8081/api/facilities/?public=false \
"Authorization: Token $OAR_API_TOKEN" \
<<< '{
  "country": "Vietnam",
  "name": "28 Hung Phu",
  "address": "No. 168 Quang Trung Street Ward 10 Go Vap District, Hochi Minh"
}'
```
- Run `./scripts/manage shell_plus` and add the user to the group that allows private submission
```
from django.contrib import auth
g = auth.models.Group.objects.get(name='can_submit_private_facility')
u = User.objects.get(email='c8@example.com')
u.groups.add(g)
u.save()
```
- Repeat the POST with `public=false` and verify a successful response with a status of `NEW_FACILITY`
```
http POST http://localhost:8081/api/facilities/?public=false \
"Authorization: Token $OAR_API_TOKEN" \
<<< '{
  "country": "Vietnam",
  "name": "28 Hung Phu",
  "address": "No. 168 Quang Trung Street Ward 10 Go Vap District, Hochi Minh"
}'
```
- Verify that the facility is searchable http://localhost:6543/facilities?q=hung%20phu and that the details page for the facility does not mention the contributor.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Add single facility submission endpoint [#896](https://github.com/open-apparel-registry/open-apparel-registry/pull/896)
 
 ### Changed
 - Replace facility history feature switch with feature flag [#881](https://github.com/open-apparel-registry/open-apparel-registry/pull/881)
--- a/src/django/api/constants.py
+++ b/src/django/api/constants.py
@@ -40,6 +40,17 @@ class FacilityMergeQueryParams:
     MERGE = 'merge'
 
 
+class FacilityCreateQueryParams:
+    CREATE = 'create'
+    PUBLIC = 'public'
+
+
+class FeatureGroups:
+    CAN_GET_FACILITY_HISTORY = 'can_get_facility_history'
+    CAN_SUBMIT_FACILITY = 'can_submit_facility'
+    CAN_SUBMIT_PRIVATE_FACILITY = 'can_submit_private_facility'
+
+
 class FacilityHistoryActions:
     CREATE = 'CREATE'
     UPDATE = 'UPDATE'
--- a/src/django/api/management/commands/batch_process.py
+++ b/src/django/api/management/commands/batch_process.py
@@ -6,9 +6,9 @@ from django.db import transaction
 
 from api.constants import ProcessingAction
 from api.models import FacilityList, FacilityListItem
+from api.matching import match_facility_list_items
 from api.processing import (parse_facility_list_item,
                             geocode_facility_list_item,
-                            match_facility_list_items,
                             save_match_details)
 
 LINE_ITEM_ACTIONS = {
--- a/src/django/api/matching.py
+++ b/src/django/api/matching.py
@@ -1,0 +1,421 @@
+import dedupe
+import logging
+import os
+import re
+import threading
+
+from collections import defaultdict
+from datetime import datetime
+from django.conf import settings
+from django.db import transaction
+from django.db.models import Q, Max
+from unidecode import unidecode
+
+from api.models import (Facility,
+                        FacilityList,
+                        FacilityListItem,
+                        HistoricalFacility)
+
+logger = logging.getLogger(__name__)
+
+
+def clean(column):
+    """
+    Remove punctuation and excess whitespace from a value before using it to
+    find matches. This should be the same function used when developing the
+    training data read from training.json as part of train_gazetteer.
+    """
+    column = unidecode(column)
+    column = re.sub('\n', ' ', column)
+    column = re.sub('-', '', column)
+    column = re.sub('/', ' ', column)
+    column = re.sub("'", '', column)
+    column = re.sub(",", '', column)
+    column = re.sub(":", ' ', column)
+    column = re.sub(' +', ' ', column)
+    column = column.strip().strip('"').strip("'").lower().strip()
+    if not column:
+        column = None
+    return column
+
+
+def get_canonical_items():
+    """
+    Fetch all `Facility` items and create a dictionary suitable for use by a
+    Dedupe model.
+
+    Returns:
+    A dictionary. The key is the `Facility` OAR ID. The value is a dictionary
+    of clean field values keyed by field name (country, name, address). A
+    "clean" value is one which has been passed through the `clean` function.
+    """
+    facility_set = Facility.objects.all().extra(
+        select={'country': 'country_code'}).values(
+            'id', 'country', 'name', 'address')
+    return {str(i['id']): {k: clean(i[k]) for k in i if k != 'id'}
+            for i in facility_set}
+
+
+def get_messy_items_from_facility_list(facility_list):
+    """
+    Fetch all `FacilityListItem` objects that belong to the specified
+    `FacilityList` and create a dictionary suitable for use by a Dedupe model.
+
+    Arguments:
+    facility_list -- A `FacilityList`.
+
+    Returns:
+    A dictionary. The key is the `FacilityListItem` ID. The value is a
+    dictionary of clean field values keyed by field name (country, name,
+    address). A "clean" value is one which has been passed through the `clean`
+    function.
+    """
+    facility_list_item_set = facility_list.source.facilitylistitem_set.filter(
+        Q(status=FacilityListItem.GEOCODED)
+        | Q(status=FacilityListItem.GEOCODED_NO_RESULTS)).extra(
+            select={'country': 'country_code'}).values(
+                'id', 'country', 'name', 'address')
+    return {str(i['id']): {k: clean(i[k]) for k in i if k != 'id'}
+            for i in facility_list_item_set}
+
+
+def get_messy_items_for_training(mod_factor=5):
+    """
+    Fetch a subset of `FacilityListItem` objects that have been parsed and are
+    not in an error state.
+
+    Arguments:
+    mod_factor -- Used to partition a subset of `FacilityListItem` records. The
+                  larger the value, the fewer records will be contained in the
+                  subset.
+
+    Returns:
+    A dictionary. The key is the `FacilityListItem` ID. The value is a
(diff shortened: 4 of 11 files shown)

### CODE REVIEW

**Comment by rajadain** [comment]
Starting to read up this now.

**Comment by jwalgran** [comment]
I pushed an additional commit that adds some swagger documentation. There are some known shortcomings, including the "Try it out form" returning CSRF errors. I want to make resolving that a separate issue.

**Comment by jwalgran** [comment]
>we should send back a 201 instead of a 200, to better communicate that something was created.

That's a good idea. Should be a simple change.

**Comment by jwalgran** [comment]
>we should send back a 201 instead of a 200, to better communicate that something was created.

Implemented in fixup 8841b18

**Comment by hectcastro** [comment]
So, the Fargate pricing page has a pricing example that reads:

> Total memory charges = # of Tasks x memory in GB x price per GB x memory duration per day (seconds) x # of days
> 
> Total memory charges = 5 x 2 x 0.000001235 x 600 x 30 = $0.22

Following the math there for our current setup leads to:

```
Staging: 1 * .5 * 0.000001235 * 86400 * 30 = ~$1.60
Production: 2 * 2 * 0.000001235 * 86400 * 30 = ~$12.80

Total: ~$14.40
```

That roughly checks out via the Cost Explorer if I group by costs for usage type `EU-Fargate-GB-Hours (Hrs)`.

If we went up to 16GB for production and played it more conservatively (2GB) for staging, it leads to:

```
Staging: 1 * 2 * 0.000001235 * 86400 * 30 = ~$6.40
Production: 2 * 16 * 0.000001235 * 86400 * 30 = ~$102.40

Total: $108.80
```

This change appears as though it would increase our overall bill by about a third of its current cost.

jwalgran:
Thanks for dpcumenting those price calculations. Good to see that there is a basically linear relationship between resources and dollars.

jwalgran:
Thanks for the review.

PR #905, opened 2019-11-05
Train persistant Dedupe model on first app request

jwalgran:
## Overview

To support the single-item match API we need to hold a trained and indexed Dedupe model in memory. It can take a minute or two to train an index a model. Attempting to do this in a web request resulted in a gunicorn timeout. Gunicorn responds to timeouts by killing and restarting the worker process, which puts us in a failure loop.

In this PR we address the problem by:
- Kicking off model training on the first request of any type, not just a request to the single-item match endpoint. Health check requests start right away, so this effectively starts training as soon as the app is started.
- Spawning a separate thread for model training so that the request that triggered the training can return and avoid a timeout.
- Adding our own timeout logic at the Django level do that we can return 503 Service Unavailable in response to requests that come in before the training is complete.

Connects #820

## Notes

To load test this we needed to upload a large (≅ 20,000) item list, which was raising an unhandled exception. At a certain size, Django switches to providing our view with a `TemporaryUploadedFile` instance instead of an `InMemoryUploadedFile` instance. Correctly handling this was a small change.

## Testing Instructions

* Run `./scripts/resetdb`
* Run `./scripts/manage shell_plus` and update group memberships
```
u = User.objects.get(email='c2@example.com')
u.groups.set(Group.objects.all())
u.save()
```
* Log in as `c2@example.com`
* Upload the ≅ 20,000 item [oar-facilities-contribute-2019-11-04.csv.zip](https://github.com/open-apparel-registry/open-apparel-registry/files/3810375/oar-facilities-contribute-2019-11-04.csv.zip)
* Fully process the list. On my machine, running the steps took between 5 and 10 minutes.
* Start/restart `./scripts/server`
* Run `./scripts/manage showmigrations` and verify that model training information messages are NOT printed.
* On the Vagrant VM, add the token to your shell environemnt: `[export OAR_API_TOKEN=]{token}`.
* These next steps require quick timing:
  * Browse http://localhost:8081/health-check/. Verify a quick, successful response. Verify that `Initializing the gazetteer cache...'` is printed in the server console.
  * Immediately make the POST request below and verify that after 10 seconds a 503 is returned.
  * Verify that training completes and `Done initializing the gazetteer cache` is printed to the server console.
  * Repeat the request below and verify a successful response.

```
http POST http://localhost:8081/api/facilities/?create=false \
"Authorization: Token $OAR_API_TOKEN" \
<<< '{
  "country": "US",
  "name": "Azavea",
  "address": "990 Spring Garden Street, Philadelphia"
}'
```

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 ### Changed
+- Train persistant Dedupe model on first app request [#905](https://github.com/open-apparel-registry/open-apparel-registry/pull/905/)
 
 ### Deprecated
 
--- a/src/django/api/matching.py
+++ b/src/django/api/matching.py
@@ -357,6 +357,10 @@ def facility_to_dedupe_record(facility):
     }
 
 
+class GazetteerCacheTimeoutError(Exception):
+    pass
+
+
 class GazetteerCache:
     """
     A container for holding a single, trained and indexed Gazetteer in memory,
@@ -373,7 +377,11 @@ class GazetteerCache:
     @classmethod
     @transaction.atomic
     def get_latest(cls):
-        with cls._lock:
+        lock_aquired = cls._lock.acquire(timeout=10)
+        if not lock_aquired:
+            raise GazetteerCacheTimeoutError
+
+        try:
             db_version = HistoricalFacility.objects.aggregate(
                 max_id=Max('history_id')).get('max_id')
             if cls._gazetter is None:
@@ -417,5 +425,7 @@ class GazetteerCache:
                             # indexing.
                             pass
                 cls._version = db_version
+        finally:
+            cls._lock.release()
 
         return cls._gazetter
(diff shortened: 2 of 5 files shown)


rajadain:
I'm getting a "File not submitted properly" response when I upload the given file:

image

jwalgran:
>I'm getting a "File not submitted properly" response

Interesting. fb4363f is supposed to take care of that. I will see if I can reproduce the error.

rajadain:
Let me make sure I wasn't accidentally on `develop`

rajadain:
Yes I was 🤦‍♂️

Now I'm getting `User contributor cannot be None`

jwalgran:
I got that too. I have amended the test instructions to use c2@example.com and use a little `shell_plus` to make sure they have the permissions they need.

rajadain:
Ingesting the 20K+ rows took almost 10 minutes on my machine.

jwalgran:
Thanks for the review.

Issue #874, opened 2019-10-14
Add a full contributor details flag


jwalgran:
## Overview

Add a group that API requestor must be a member of to get access to full contributor details if the requestor is also in the `can_submit_private_data` group.

Depends on #871

### Describe the solution you'd like

- Create a `can_view_full_contrib_detail` `Group` https://docs.djangoproject.com/en/2.2/topics/auth/default/#groups
- Create a `can_view_full_contrib_detail` `Flag` https://waffle.readthedocs.io/en/stable/types/flag.html.
  - Assign the `can_view_full_contrib_detail` `Group` to the new `Flag`.
- In facility detail API responses show all contributor details as anonymous (e.g An MSI, 2 Brands) if the `can_submit_private_data` flag is active and this flag IS NOT.
- Show ASSOCIATE and DISSOCIATE events with anonymous details if the `can_submit_private_data` flag is active and this flag IS NOT.

PR #910, opened 2019-11-08
Add can view full contributor detail flag/group

jwalgran:
## Overview

This permission acts as an override to allow members of the "can submit private facilities" group to still see contributor details.

It replaces checking if the user has contributed any public data, allowing us to have more control over the visibility of contributors and to avoid a situation where a contributor submits thousands of private facilities and a single public one to get access to the full contributor details.

We also updated the history endpoint to follow the design of the recent changes to the facility details serializer where we conditionally show the contributor type instead of the full name unless the user making the request has the appropriate permissions.

Connects #874

## Demo

Screen Shot 2019-11-08 at 1 25 25 PM

## Notes

Also fixes a bug where items submitted individually would appear in the history as being from "list [Unknown]." We now exclude the list description if there is no list associated with the `Source.`

Also includes a new `user_groups` management command. The Django Admin does not have a default UI for managing group memberships. This command allows us to use `ecsmanage` to update group memberships in ECS deployments where do not have access to an interactive `shell_plus` session.

## Testing Instructions

* Run `./scripts/resetdb`
* Log in a `c2@example.com:password`
* Browse http://localhost:6543/?q=regina and select the facility in Shenzhen. Copy the OAR ID.
* Browse http://localhost:8081/api/docs/#!/facilities/facilities_get_facility_history. Submit the test form with the OAR ID. Verify that a 403 error is returned.
* Run management command to allow history access
```
./scripts/manage user_groups -e c2@example.com -a add -g can_get_facility_history
```
* Submit the API test form again and verify that history is returned and that the "detail" values contain full contributor details.
* Run management command to allow private submission access
```
./scripts/manage user_groups -e c2@example.com -a add -g can_submit_private_facility
```
* Submit the API test form again and verify that history is returned but that the contributor details are now anonymous.
* Browse http://localhost:6543/facilities/{the-oar-id} and verify that the contributor details are anonymous.
* Run management command to allow full contributor details
```
./scripts/manage user_groups -e c2@example.com -a add -g can_view_full_contrib_detail
```
* Submit the API test form again and verify that history is returned and that the "detail" values once again contain full contributor details.
* Browse http://localhost:6543/facilities/{the-oar-id} and verify that the full contributor details are displayed.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Train persistant Dedupe model on first app request [#905](https://github.com/open-apparel-registry/open-apparel-registry/pull/905/)
 - List inactive or private sources by contributor type [#907](https://github.com/open-apparel-registry/open-apparel-registry/pull/907)
+- Add can view full contributor detail flag/group [#910](https://github.com/open-apparel-registry/open-apparel-registry/pull/910)
 
 ### Deprecated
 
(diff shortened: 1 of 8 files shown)


jwalgran:
Thanks for the review.

Issue #845, opened 2019-10-02
EmptyResultSet


jwalgran:
View details in Rollbar: [https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/288/](https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/288/)

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 497, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 457, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 468, in raise_uncaught_exception
    raise exc
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 494, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/decorators.py", line 50, in handler
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/views/decorators/cache.py", line 31, in _cache_controlled
    response = viewfunc(request, *args, **kw)
  File "/usr/local/lib/python3.7/site-packages/waffle/decorators.py", line 50, in _wrapped_view
    return view(request, *args, **kwargs)
  File "/usr/local/src/api/views.py", line 2361, in get_tile
    request.query_params, layer, z, x, y)
  File "/usr/local/src/api/tiler.py", line 35, in get_facility_grid_vector_tile
    .values('location') \
  File "/usr/local/lib/python3.7/site-packages/django/db/models/sql/query.py", line 273, in sql_with_params
    return self.get_compiler(DEFAULT_DB_ALIAS).as_sql()
  File "/usr/local/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 489, in as_sql
[...]

PR #911, opened 2019-11-11
Handle EmptyResultSet error raised by tile requests

jwalgran:
## Overview

In difficult to recreate situations an `EmptyResultSet` exception is being raised by Django while generating the where clause for a grid vector tile. In this context it is logically correct to simply return 204 No Content with an empty body, which the Leaflet grid layer handles without any issues.

Connects #845

## Notes

I could not reliably recreate the error condition so there is an additional commit that should be dropped before merging the PR which deterministically raises the problematic exception for every tile request with an even numbered X and Y. 

## Testing Instructions

* (Re)start `./scripts/server`.
* Run `./scripts/manage incrementtileversion` to invalidate any in-browser cached tiles.
* Open the developer tools "Network" tab, browse http://localhost:6543/, and zoom in a few times. Verify that:
  * All the requests to `/tile/*` return 200 or 204.
  * No errors are logged to the server console.

## Checklist

- [x] DROP test commit
- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Removed
 
 ### Fixed
+- Handle EmptyResultSet error raised by tile requests [#911](https://github.com/open-apparel-registry/open-apparel-registry/pull/911)
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


jwalgran:
Thanks for the review.

Issue #921, opened 2019-11-27
Search for contributor overlap ("AND" search of multiple contributors)


jwalgran:
## Overview

Allow searching facilities by multiple contributors using "AND" instead of "OR."

### Is your feature request related to a problem? Please describe.

Finding the "overlap" between the facility lists from two or more contributors is a useful feature that has been requested by contributors. Combining searches with "AND" was also the default for the v1 application.

### Describe the solution you'd like

- When more than one contributor is selected, display a checkbox to enable searching for "shared facilities."
- Include the state of this checkbox in the search URL and restore it when loading a bookmarked search
- Add a `combine-contributors` parameter to the API. When set to "AND" (case insensitive) change the contributor filter to use "and" instead of "or."

### Additional Context

Winning hackathon project that implemented this idea using a static data set.

https://github.com/flache/hyper-oar/
http://hyper-oar.s3-website.eu-central-1.amazonaws.com/

jwalgran:
Scheduled a discussion with @designmatty to talk about the design.

designmatty:
Continuing the conversation from our call, I think the following UI should support this request without requiring drastic changes:

designmatty:
@jwalgran let me know if I missed anything.

jwalgran:
Thanks. This is great. We talked about including an information popup to include some inline documentation explaining in more detail what "shared facilities" means. How would you suggest implementing that help popup?

designmatty:
Oh good catch. I updated my previous comment and screenshot. There is now an "info" icon. On hover of the icon utilize the tooltip component, which I linked to in the above comment.

jwalgran:
@katieashaw and/or @nataliegrillon 

Matt and I discussed the overlapping contributor feature and came up with this design that includes a "i" help icon which gives us a place to show some popover text explaining what the checkbox does. 

![69988990-b3e4d580-1510-11ea-82bd-127a503b3133](https://user-images.githubusercontent.com/17363/70481054-9f6b8480-1a9e-11ea-858a-f7ad36e9f3a0.png)

Could you please post the short bit of help text you would like us to show?

katieashaw:
Oh, this is great - so simple, I love it!

Won't take long to write this. When do you need the text by @jwalgran ?

jwalgran:
Thanks. Missing texts is not blocking any development tasks, so there is no real rush. Would it work with your schedule to have it done by the end of the UK day on Wednesday?

[...]

Issue #406, opened 2019-03-26
Zoom to the extent of the search results when running a search or loading from a shared search or on demand


jwalgran:
## Overview

When running a search on page load, zoom the map to show the extent of all the results. Add a zoom-to-extent button to search results panel.

### Describe the solution you'd like

- When running a search on the map page, or on page load, zoom the map to show the extent of all the results. 
- Add a "Zoom to search" checkbox. Checked by default. If disabled, don't move the map when the search changes.

PR #925, opened 2019-12-11
Add filter option to search for contributor overlap

jwalgran:
## Overview

Adds a new option to the facility search API and and associated UI that allows optionally filtering facilities by contributor using an "AND" search rather than an "OR" search.

Connects #921

## Demo

NOTE [the help text was revised](https://github.com/open-apparel-registry/open-apparel-registry/pull/925#issuecomment-564654520) after this initial screencast was recorded.

## Notes

While adding support for this to the `filter_by_query_params` method we also replaced the previous filter by contributor code with a statement that computes the "IN" query on the server side. The large SQL statements generated by passing thousands of IDs in a literal "IN" statement were so large that they made analyzing query performance from SQL slow query logs difficult to impossible.

The tooltip text is a placeholder and will be replaced with a submission from Katie.

## Testing Instructions

* Run `./scripts/resetdb` and `./scripts/server`
* Search for "regina" and select the facility in China. Remember 2 of the three listed contributors.
* Reset the search, return to the "Search" tab, and enter the 2 contributors from the previous step in the "Filter by Contributor" box. Verify that the new checkbox only appears after the second item is added to the list.
* Run the search and verify that there are multiple facility matches.
* Return to the "Search" tab and check the checkbox. Run the search again and verify that only a single facility is now returned.
* Refresh the page, verify that the search is restored, return to the "Search" tab and verify that the checkbox is checked.
* Remove one of the facilities from the  "Filter by Contributor" box and verify that the checkbox is hidden and the `combine_contributors` querystring argument is no longer present in the URL.
* Browse http://localhost:8081/api/docs/#!/facilities/facilities_list and verify that the `combine_contributors` argument is documented.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/components/FilterSidebarSearchTab.jsx
+++ b/src/app/src/components/FilterSidebarSearchTab.jsx
@@ -4,17 +4,24 @@ import { connect } from 'react-redux';
 import InputLabel from '@material-ui/core/InputLabel';
 import TextField from '@material-ui/core/TextField';
 import Button from '@material-ui/core/Button';
+import Checkbox from '@material-ui/core/Checkbox';
 import CircularProgress from '@material-ui/core/CircularProgress';
+import FormControlLabel from '@material-ui/core/FormControlLabel';
+import IconButton from '@material-ui/core/IconButton';
+import InfoIcon from '@material-ui/icons/Info';
+import Tooltip from '@material-ui/core/Tooltip';
 import ReactSelect from 'react-select';
 import get from 'lodash/get';
 
 import FacilitySidebarSearchTabFacilitiesCount from './FacilitySidebarSearchTabFacilitiesCount';
+import ShowOnly from './ShowOnly';
 
 import {
     updateFacilityFreeTextQueryFilter,
     updateContributorFilter,
     updateContributorTypeFilter,
     updateCountryFilter,
+    updateCombineContributorsFilterOption,
     resetAllFilters,
 } from '../actions/filters';
 
@@ -70,6 +77,8 @@ function FilterSidebarSearchTab({
     updateContributorType,
     countries,
     updateCountry,
+    combineContributors,
+    updateCombineContributors,
     fetchingFacilities,
     searchForFacilities,
     facilities,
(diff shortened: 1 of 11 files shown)


jwalgran:
I force pushed an amended commit that fixes the JS unit tests that failed the CI build

jwalgran:
I pushed a fixup commit that adds the full help text I received from @katieashaw 

Screen Shot 2019-12-11 at 10 37 17 AM

katieashaw:
@jwalgran It looks from this screen shot that the text in bold at the top of the document I shared is missing. Could you include it please?

jwalgran:
Thank you for catching that, Katie. I added the title in another fixup commit.

Screen Shot 2019-12-11 at 11 58 24 AM

rajadain:
Taking a look at this now.

src/app/src/components/FilterSidebarSearchTab.jsx
```diff
@@ -107,6 +116,48 @@ function FilterSidebarSearchTab({
...
+                share? If so, tick this box.
+            </p>
+            <p>There are now two ways to filter a Contributor search on the OAR:</p>
+            <ol>
+                <li style={styles.tooltipLineItem}>
+                    You can search for all the facilities of multiple
+                    contributors. this means that the results would show all of
+                    the facilities contributed to the oar by, for example, brac
+                    university or clarks. some facilities might have been
+                    contributed by brac university but not by clarks, or
```
rajadain: This paragraph seems to be missing proper casing

the oar → the OAR
brac university → BRAC University
clarks → Clarks

jwalgran: Good catch. I pushed fixup commit b28bad8 to resolve this.

rajadain:
One interesting behavior I'm seeing is that when there's a single match, the sidebar jumps into the detail view, but the map remains in the general view:

image

This may be previous behavior unrelated to this card though, in which case we can ignore it.

jwalgran:
Loading the result when there is a single match is an existing behavior. You can see an example on staging by searching for contributor "Aldi [Public List]" and Country "Portugal"

src/app/src/components/FilterSidebarSearchTab.jsx
```diff
@@ -140,10 +140,10 @@ function FilterSidebarSearchTab({
...
                 <li style={styles.tooltipLineItem}>
                     You can search for all the facilities of multiple
-                    contributors. this means that the results would show all of
-                    the facilities contributed to the oar by, for example, brac
-                    university or clarks. some facilities might have been
-                    contributed by brac university but not by clarks, or
+                    contributors. This means that the results would show all of
+                    the facilities contributed to the OAR by, for example, BRAC
+                    University or Clarks. Some facilities might have been
+                    contributed by BRAC university but not by Clarks, or
```
rajadain: University should be capitalized here

jwalgran: Thanks. Pushed fixup 424bff5 to correct this.

src/django/api/models.py
```diff
@@ -1055,22 +1058,55 @@ def filter_by_query_params(self, params):
...
+                facility_ids = facility_ids.values_list(
+                    'facilitylistitem__facility', flat=True)
+
+                facilities_qs = facilities_qs.filter(id__in=facility_ids)
 
-            facilities_qs = facilities_qs \
-                .filter(id__in=name_match_facility_ids)
+            else:
+                facilities_qs = facilities_qs.filter(
+                    id__in=(Source.objects.filter(
```
rajadain: I notice that even in the previous case (of not having combine contributors active), we are now using `Source` instead of `FacilityMatch` for getting the values. Is this also deliberate?

jwalgran: It was meant to make the query used by the two conditions as similar as possible, and the more complicated "AND" query needs to start from the `Source` table. Also, it makes sense that, when we are filtering by contributor, we start our query from the table on which the `contributor` field exists.

katieashaw:
Chiming in here slightly belatedly having checked a point from this exchange with @nataliegrillon in reference to the point made by @rajadain (copied below). We would prefer for the map view to zoom in these instances. Is that possible @jwalgran ? I know we've had a similar exchange about the zoom function on specific facilities in the past [see issue "The zoom on the OAR map is behaving in an inconsistent manner #757"], so it might be worth revisiting as a group. 

> One interesting behavior I'm seeing is that when there's a single match, the sidebar jumps into the detail view, but the map remains in the general view:
> 
> image
> 
> This may be previous behavior unrelated to this card though, in which case we can ignore it.

jwalgran:
We discussed this on a weekly check-in call and agreed that the solution proposed #406 is a good one.

Issue #933, opened 2019-12-30
Allow group membership to be changed in the Django admin 


jwalgran:
## Overview

Allow group memberships to be changed in the Django admin.

### Is your feature request related to a problem? Please describe.

Superusers must currently contact the development team to grant API access to a contributor. 

### Describe the solution you'd like

Allow group memberships to be set via the User page in the Django admin.

### Additional Context

The Django admin supports setting group memberships but we mistakenly removed it when creating a custom user admin configuration.

https://github.com/open-apparel-registry/open-apparel-registry/blob/34def4ecd498c6ed25f17ab98c53fa9db7f3bcdc/src/django/api/admin.py#L45-L51

PR #934, opened 2019-12-30
Allow groups memberships to be changed in the Django admin

jwalgran:
## Overview

When we implemented a custom admin configuration for the users page we mistakenly omitted groups. This may have been because there was an error trying to save the page with an empty `username` field. Excluding `username` (we only use `email` in OAR) resolves the save error.

Connects #933

## Demo

Screen Shot 2019-12-30 at 9 13 08 AM

## Testing Instructions

These notes assume `./scripts/resetdb` has been run.

* Log in as `c2@example.com`
* Browse http://localhost:8081/api/docs/ and click the 'Authorize" button. Submit `Token {your token}`
* Browse http://localhost:8081/api/docs/#!/facilities/facilities_create (you will likely need to refresh)
  * Copy the "Sample Request Body" JSON and paste it into the "data" form field.
  * Set "create" to false.
  * Submit. Verify that a 403 error is returned.
* In a separate browser or private window, log in as `c1@example.com`
* Browse http://localhost:8081/admin/api/user/2/change/
* Move `can_submit_facility` from "Available groups" to "Chosen groups" and click "Save." Verify that the save is successful
* Return to the Swaggar API page and submit the request again. Verify that there is now a 200 response rather than a 403. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Allow groups memberships to be changed in the Django admin [#934](https://github.com/open-apparel-registry/open-apparel-registry/pull/934)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


rajadain:
Taking a look now.

jwalgran:
Thanks for the review.

Issue #937, opened 2020-01-06
Reports are not taking year into consideration


jwalgran:
## Overview

The admin reports exclusively filter by month, which is causing problems now that we are in a new year. 

### Expected Behavior

Reports filter and group by year and month in combination.

### Actual Behavior

Reports only filter and group by month.

### Steps to Reproduce

* Run any report that excludes data from the current month

PR #938, opened 2020-01-07
Update reports to properly handle multiple years

jwalgran:
## Overview

Previously the reports were filtering and grouping by month number. Now that we have data from different years we need to use a combination of year and month.

Connects #937

## Testing Instructions

I ran all of the revised reports on production and compared their output to the existing production versions edited to remove data filtering that filters out all the rows. All counts matched. 

The following steps assumes you have users created by running `./scripts/resetdb`

* Log in as `c1@example.com`
* Browse http://localhost:8081/admin/reports/
* Verify that all reports run without error. Many will be empty because the test data does not match the filter criteria of the report.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Removed
 
 ### Fixed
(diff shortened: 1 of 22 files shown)


rajadain:
Taking a look now

jwalgran:
Thanks for the review.

Issue #947, opened 2020-02-05
Research the impact of incorporating accept/reject decisions into model training 


jwalgran:
## Overview

Research the impact of incorporating accept/reject decisions into model training .

### Is your feature request related to a problem? Please describe.

Manual matches/rejects are not incorporated into the model. Submitting revised lists requires the same approvals/rejections for potential matches.

### Describe the solution you'd like

- Choose a test list that has potential matches when processed in a fresh development database.
- Confirm matches, submit the same list and verify (or not) that the same items appear as potential matches.
- Modify the model training code to incorporate approved/rejected matches as additional training examples
  - Match the JSON format of the existing training data.
- Submit the same list twice and verify that the second time matches are made automatically. If not, try to determine why.

PR #963, opened 2020-02-25
Use confirmed matches as training data

jwalgran:
## Overview

Attempts to make subsequent matching of replacements lists require less
intervention from the contributor by adding previously confirmed matches to the
training data.

This change does not appear to help. Adding it as a branch so it can be archived
and returned to if needed.

These are the testing steps:

  * Run ./scripts/resetdb
  * Log in as c8@example.com
  * Confirm matches on http://localhost:6543/lists/8
  * Upload src/django/api/management/commands/facility_lists/geocoded/8.csv and
  replace the previous list.
  * Fully process the uploaded list
  * Browse the results http://localhost:6543/lists/16

Adding the manual matches as training data via `markPairs` does not boost the
match scores. The match score for row index 11 is 0.69 both times.

Connects #947

### DIFF
--- a/src/django/api/matching.py
+++ b/src/django/api/matching.py
@@ -14,6 +14,7 @@ from unidecode import unidecode
 from api.models import (Facility,
                         FacilityList,
                         FacilityListItem,
+                        FacilityMatch,
                         HistoricalFacility)
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


jwalgran:
Closing because the purpose of this PR was just to create a historical record of a failed solution.

PR #964, opened 2020-02-25
Add confirmed matches to canonical facility list when matching

jwalgran:
## Overview

Prior to this commit, when a contributor manually confirmed a facility match, then uploaded a replacement list, the same potential match would be presented for repeat confirmation because there was no mechanism to feed the confirmation back into the matching process.

Our first attempt to resolve this was by augmenting the training data loading from a static JSON file with previously confirmed matches, but this did not boost the match scores enough.

This commit resolves the problem by augmenting the canonical list of facilities with items pulled from confirmed matches. We generate temporary, "extended" IDs for these items in the format `US2020052GKF19F_MATCH-23` and normalized them back to real Facility IDs when we create model objects from the output of the Dedupe matching process.

Connects #947 

## Testing Instructions

* Run ./scripts/resetdb
* Log in as c8@example.com
  * Reject the match for row index 10
  * CONFIRM the match for row index 11
  * CONFIRM the match for row index 18 (if it is suggested)
  * Reject both matches for row index 25
* Browse http://localhost:6543/contribute, upload src/django/api/management/commands/facility_lists/geocoded/8.csv, and **replace the previous list.**
* Fully process the uploaded list
* Browse the results http://localhost:6543/lists/16. Verify that there are no pending matches, particularly row index 11.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Add a report that groups facility counts by country [#979](https://github.com/open-apparel-registry/open-apparel-registry/pull/979)
+- Add confirmed matches to canonical facility list when matching [#964](https://github.com/open-apparel-registry/open-apparel-registry/pull/964)
 
 ### Changed
 - Zoom to search [#966](https://github.com/open-apparel-registry/open-apparel-registry/pull/966)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,26 +8,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
-
-- Add a report that groups facility counts by country [#979] (https://github.com/open-apparel-registry/open-apparel-registry/pull/979)
+- Add a report that groups facility counts by country [#979](https://github.com/open-apparel-registry/open-apparel-registry/pull/979)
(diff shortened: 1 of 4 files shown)


jwalgran:
Close by mistake.

rajadain:
Whoops, sorry, will take a look today.

rajadain:
After following the instructions, I see that row 11 is marked as a New Facility:

image

As is row 33:

image

Is this expected?

jwalgran:
I would expect everything to match but, in general, our results are not 100% repeatable since Dedupe uses a statistical approach and uses some intentional randomness in the training process. I am going to run through the test steps again to verify.

jwalgran:
I just looked more closely at your results and noticed the "Lion Brothers" facility at row index 11 where I would expect to see the "Jackbeanie" facility

"Lion Brothers" does not appear in the CSV mentioned in the test instructions
https://github.com/open-apparel-registry/open-apparel-registry/blob/0c502854baaf8f7efcfbd5c230bf1c44c14fb05a/src/django/api/management/commands/facility_lists/geocoded/8.csv#L13

rajadain:
Hmm taking another look.

rajadain:
Ah sorry I was using the non-geocoded list for the tests

https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/django/api/management/commands/facility_lists/8.csv#L13

Doing this over with the right one.

jwalgran:
Thanks for the review.

PR #966, opened 2020-02-28
Zoom to search

TaiWilkin:
## Overview

The features search API includes the extent (outer boundaries) of the
search results in the form [y1, x1, y2, x2], representing the lower left
and upper right corners of the boundaries of the results.

We want to zoom to the extent of the search results, so we have added
this to allow that functionality. If the data from a search includes an extent field
(an array of [y1, x1, y2, x2]) then the map will zoom to fit the boundaries of the extent. The
map will zoom again if the extent changes. 

The zoom to search functionality is turned on and off by a checkbox on
the map, which is selected by default. Users are able to disable zoom to
search by unchecking the checkbox.

The state of the checkbox is held in the UI reducer. This seemed like
a logical location as it primarily controls the way information is
presented, not the information itself.

Connects #406 

## Demo

Screen Shot 2020-02-28 at 11 28 50 AM

## Notes

- The checkbox to disable the filter is currently located on the map and mimics the style of the legend that already exists. This should be reviewed by a designer.

- Because there is no hook in the pagination classes to allow additional
data fields to the response, we added the extent directly to the
response data. We attempted an extension of the pagination class to
allow for the addition of the extent data directly, but the current
setup of the classes involved didn't allow for this extension.

## Testing Instructions

* Run `./scripts/update` in Vagrant ssh
* Search for facilities - searching with a specific region name will give the most noticeable results
* The map should zoom to the search results
* Search for a specific facility by name
* The map should zoom to the facility
* Disable 'zoom to search' with the checkbox on the map
* Search for facilities, then a specific facility
* The map should not zoom in response to the searches

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/util/hooks.js
+++ b/src/app/src/util/hooks.js
@@ -18,9 +18,33 @@ export default function useUpdateLeafletMapImperatively(
         data,
         shouldPanMapToFacilityDetails,
         isVectorTileMap = false,
+        extent,
     } = {},
 ) {
     const mapRef = useRef(null);
+
+    const [
+        currentExtent,
+        setCurrentExtent,
+    ] = useState(extent);
+    useEffect(() => {
+        if (zoomToSearch && extent != null && currentExtent !== extent) {
+            const leafletMap = get(mapRef, 'current.leafletElement', null);
+
+            if (leafletMap) {
+                leafletMap.fitBounds([
+                    [extent[3], extent[2]],
+                    [extent[1], extent[0]],
+                ], { maxZoom: detailsZoomLevel, padding: [20, 20] });
+            }
+
+            setCurrentExtent(extent);
+        }
+    }, [
+        extent,
+        currentExtent,
+        zoomToSearch,
+    ]);
 
     // Reset the map state when the reset button is clicked.
     const [
--- a/src/app/src/util/hooks.js
+++ b/src/app/src/util/hooks.js
@@ -19,6 +19,7 @@ export default function useUpdateLeafletMapImperatively(
         shouldPanMapToFacilityDetails,
         isVectorTileMap = false,
         extent,
+        zoomToSearch,
     } = {},
 ) {
     const mapRef = useRef(null);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
+- Zoom to search
+
 ### Deprecated
 
 ### Removed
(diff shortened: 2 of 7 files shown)


jwalgran:
@designmatty I added you as a reviewer on this commit PR Tai and I wanted to get your input on the design of the new "Zoom To Search" checkbox. Tai did a great job matching it to the style of the existing legend controb, but we would welcome a different design if you think a different approach would be more effective.

75566350-87f8c380-5a1d-11ea-81d1-87532f493a89

src/app/src/util/hooks.js
```diff
@@ -18,10 +18,35 @@ export default function useUpdateLeafletMapImperatively(
...
+    ] = useState(extent);
+    useEffect(() => {
+        if (zoomToSearch && extent != null && currentExtent !== extent) {
+            const leafletMap = get(mapRef, 'current.leafletElement', null);
+
+            if (leafletMap) {
+                leafletMap.fitBounds([
+                    [extent[3], extent[2]],
+                    [extent[1], extent[0]],
+                ], { maxZoom: detailsZoomLevel });
```
jwalgran: Ran into a literal edge case :-) When searching in Bangladesh there are some markers that are clipped at the edge of the map at my current monitor resolution. Consider adding a buffer here. I tried `[20, 20]`

TaiWilkin: This appears to be at least partially related to the dimensions of the viewport when zooming in the map. If I set my window to be tall and fairly narrow, the items at the left/right of the screen get clipped; similarly, if I set my window to be wide and short, the items at the top/bottom of the screen get clipped. Adding the padding prevents the clipping in most circumstances, but when the screen is at an extreme height/width ratio it still occurs. 

Do we want to accept the padding as 'fixes most cases', or should I go back to the drawing board to try to find a different method of handling this?

jwalgran: I think that padding is the the right solution for this problem. There are likely some magic numbers that prevent clipping 100% of the time, but I don't think it is worth spending more time trying to optimize it for every case. We can revisit this and adjust it if we get repeated feedback about it being a usability issue.

jwalgran:
@designmatty We merged this to not cause conflicts with our upcoming work on polygonal search. If you have some recommended changes we can implement them in a follow-up PR.

TaiWilkin:
Based on a conversation in Slack, adding a note on context related to this PR for historical reference:
"When adding "zoom to extent" as the default, we wanted to leave the original behavior as an option. Use case was that a researcher might zoom to a part of the world they are specifically interested in and want to play will the filters to see various search results in that area."

Issue #959, opened 2020-02-20
Language update: "Parent company" not always understood by facility owners / factory groups


katieashaw:
## Overview

Through the "Claim a Facility" mechanism, we request information on "Parent Company" which, if available on the OAR, is listed on claimed profiles. This needs to be updated to "Parent Company / Supplier Group".

### Is your feature request related to a problem? Please describe.

We heard during the recent OECD conference that many facility owners or factory groups do not understand use of the language "Parent Company" as it relates to company structure. 

### Describe the solution you'd like

This field in the OAR "Claim a Facility" process, and on the claimed profiles, should be updated to: "Parent Company / Supplier Group".

### Additional Context

Optional. Add any other context, screenshots, gifs, etc about the feature request here.

PR #971, opened 2020-03-02
Update parent company label

TaiWilkin:
## Overview

Through the "Claim a Facility" mechanism, we request information on "Parent Company" which, if available on the OAR, is listed on claimed profiles. We heard during the recent OECD conference that many facility owners or factory groups do not understand use of the language "Parent Company" as it relates to company structure. This label has been updated to "Parent Company / Supplier Group" where it appears throughout the application.

Connects #959 

## Demo

Screen Shot 2020-03-02 at 2 10 03 PM

## Testing Instructions

* Checkout this branch
* Claim a facility. When you get to Step 3 in the process, 'Parent Company' should read 'Parent Company / Supplier Group'
* Verify the facility claim through an admin account. 'Parent company' should be labeled 'Parent company / supplier group'
* As the user who claimed the facility, look at the facility details. 'Parent Company' should read 'Parent Company / Supplier Group'

## Checklist

- [ ] CI passes after rebase


src/django/api/models.py
```diff
@@ -873,8 +873,9 @@ class FacilityClaim(models.Model):
         null=True,
         default=None,
         on_delete=models.PROTECT,
-        verbose_name='parent company',
-        help_text='The parent company of this facility claim.')
+        verbose_name='parent company / supplier group',
+        help_text='The parent company / supplier group of this'
+        'facility claim.')
```
jwalgran: There is a missing space between `this` and `facility`. Unfortunately this means that, when it is corrected, you will need to recreate the migration. The easiest way to do this is to:
- Unapply the migration with `./scripts/manage migrate api 0042`
- Delete 0043_auto_20200303_1837.py
- Fix the help text
- Rerun `./scripts/manage makemigrations`

We like to rename the automatically generated migration files so that you can see at a glance what they do. After recreating the migration, consider renaming it to `0043_facility_claim_parent_company_verbose_name_change.py`

TaiWilkin: fixed in 5d0c66b0f2ef1941402f825469aaf174ca354f89

Issue #930, opened 2019-12-16
Display number of users per Contributor type in the Contributor dropdown menu


katieashaw:
## Overview

A user has contacted me to express interest in being able to see the number of Contributors to the OAR by Contributor type, e.g. 31 Brands / Retailers, 28 Civil Society Organisations etc. Could we explore making this information visible through the front end of the OAR? 

### Is your feature request related to a problem? Please describe.

These figures are currently only available to the OAR team through our monthly reports. Users would have to reach out to the OAR team to request this information 

### Describe the solution you'd like

Would it work to display the figure in brackets by each Contributor type in the Contributor dropdown menu and have this information update automatically when new Contributors upload to the tool?

dboyer:
@katieashaw Do you know anything more about the potential use case? Is a number sufficient or would they also want a list of the names of each brand, CSO, etc?

katieashaw:
I think at a minimum, displaying the numbers would be helpful for many users. Being able to navigate to then see the list of each brand, CSO etc is a nice to have, so would depend on how tricky that would be to build.

cc @nataliegrillon for info

katieashaw:
I'm doing a detailed review of feedback from our user survey and noting here that an additional user has requested this feature, also requesting the figures on a per-country basis. (The user also requests it at a regional level, which will be covered by polygonal search.)

PR #981, opened 2020-03-05
Display number of contributors per type

TaiWilkin:
## Overview

The endpoint to fetch all contributors now queries the number of
contributors per contributor type, the appends the count to the
display name for each contributor type. This allows the UI to
display the number of contributors for each type under the contributor
type dropdown in search.

This functionality was requested by several users, as currently this
information is only available through the admin reports.

Connects #930 

## Demo

Screen Shot 2020-03-05 at 11 52 44 AM

## Testing Instructions

* Checkout this branch
* Navigate to the search tab and dropdown the contributor type select
* The contributor types should each show a number of contributors.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 - Zoom to search [#966](https://github.com/open-apparel-registry/open-apparel-registry/pull/966)
+- Display number of contributors per type [#981](https://github.com/open-apparel-registry/open-apparel-registry/pull/981)
 
 ### Deprecated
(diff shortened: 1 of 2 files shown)


rajadain:
Sorry I missed this. My instinct is to send back a count from the API and do the visual formatting for the display in the front-end (as presentation is more the domain of the front-end), but not sure if that would have an impact on the performance.

jwalgran:
I don't think passing count as a separate data element would have any performance implications. That said, we do not do anything with the number other than display it within a select option so I think it is ok to handle the formatting "server side." It has the advantage of not changing the shape of our API response, but that caries little weight here since our web client is the only consumer of the API.

Issue #977, opened 2020-03-04
Open to search tab on initial page load


dboyer:
## Overview

When visiting the OAR map page, have the search tab visible instead of the facilities tab

### Is your feature request related to a problem? Please describe.

Currently, the page loads with the first tab (the facilities tab) visible and it is not immediately obvious what the user should do. Loading the search tab immediately showcases the search functionality.

### Describe the solution you'd like

The search tab should be visible when accessing the main openapparel.org page. Consult @designmatty in regards to whether the tabs should be reordered so Search is first or if just opening to the Search tab is sufficient.

### Additional Context

Currently, the number of total facilities/search results is at the bottom of the Search tab. Consult with @designmatty on where it could be placed higher up on the page so it is more visible.

designmatty:
Please switch the order of the tabs, Search first, then Facilities

TaiWilkin:
@designmatty Do you have thoughts on where we can place the number of total facilities/search results higher on the page so it is more visible?

designmatty:
I think we could do one of the following:

**Option A**
![image](https://user-images.githubusercontent.com/1928955/76109541-a4ab7300-5faa-11ea-9214-a1842be99bc7.png)

**Option B**
![image](https://user-images.githubusercontent.com/1928955/76109826-4632c480-5fab-11ea-9b5d-f9fb6bc4a149.png)

@dboyer any preference?

jwalgran:
We have discussed the possibility of removing that big blue bar, so I would vot for option A

![image](https://user-images.githubusercontent.com/1928955/76109541-a4ab7300-5faa-11ea-9214-a1842be99bc7.png)

TaiWilkin:
@dboyer @jwalgran Currently, the facilities count shown on the search tab is the count of total facilities (NOT the count of search results), but the count on the facilities tab is the count of search results. Which number do we want to show in the tab bar? And do we want to remove the total count (on search tab) and/or the search results count (on facilities tab)?

jwalgran:
I think we get a 2-for-1 here. It is important to engagement/marketing efforts that we prominantly show the total number of facilities when you arrive at https://openapparel.org/. I think we should connect the count in the tab bar to the current number of search results since that will also be the full count when arriving on the site with no filters. The separation between the full count and the filter count is a leftover from when we were capping search results at 500.

PR #985, opened 2020-03-10
Open to search tab

TaiWilkin:
## Overview

The page now opens to show the search tab on initial page load. This helps users see immediately the functionality that is available in terms of searching. 

In line with this change, the tabs' order has changed to search, facilities, guide.

The search results count has also been added to the tab bar.

Connects #977 

## Demo

Screen Shot 2020-03-10 at 10 13 05 AM
Screen Shot 2020-03-10 at 10 08 13 AM

## Notes

We are no longer using the total facilities count in the UI. If we don't think we will use that API in the future, we could remove the actions and reducer related to that functionality.

The search results count still appears in the facilities tab - is this clarifying for users, or needless repetition of information?

## Testing Instructions

* Checkout this branch
* Open the page; it should open to the search tab. The total facilities count should show in the tab bar.
* Complete a search. The facilities tab should open and the search results count in the tab bar should update.
* Refresh the page. The search tab should open, but the search results count in the tab bar should remain the same.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/actions/facilities.js
+++ b/src/app/src/actions/facilities.js
@@ -16,6 +16,8 @@ import {
     createQueryStringFromSearchFilters,
 } from '../util/util';
 
+import { makeSidebarFacilitiesTabActive } from './ui';
+
 export const startFetchFacilities = createAction('START_FETCH_FACILITIES');
 export const failFetchFacilities = createAction('FAIL_FETCH_FACILITIES');
 export const completeFetchFacilities = createAction('COMPLETE_FETCH_FACILITIES');
@@ -29,6 +31,7 @@ export const resetSingleFacility = createAction('RESET_SINGLE_FACILITY');
 export function fetchFacilities({
     pageSize = FACILITIES_REQUEST_PAGE_SIZE,
     pushNewRoute = noop,
+    activateFacilitiesTab = true,
 }) {
     return (dispatch, getState) => {
         dispatch(fetchCurrentTileCacheKey());
@@ -58,10 +61,14 @@ export function fetchFacilities({
 
                     pushNewRoute(makeFacilityDetailLink(facilityID));
                 }
-
                 return data;
             })
-            .then(data => dispatch(completeFetchFacilities(data)))
+            .then((data) => {
+                dispatch(completeFetchFacilities(data));
+                if (activateFacilitiesTab) {
+                    dispatch(makeSidebarFacilitiesTabActive());
+                }
+            })
             .catch(err => dispatch(logErrorAndDispatchFailure(
                 err,
                 'An error prevented fetching facilities',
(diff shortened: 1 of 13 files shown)


jwalgran:
Is there a situation where these two counts will be different? 

Screen Shot 2020-03-10 at 12 45 50 PM

If not, I think we can just leave the one count in the tab bar.

jwalgran:
@designmatty I am interested in your opinion on how can not loose the count at narrower screen widths

src/app/src/actions/facilities.js
```diff
@@ -29,6 +31,7 @@ export const resetSingleFacility = createAction('RESET_SINGLE_FACILITY');
 export function fetchFacilities({
     pageSize = FACILITIES_REQUEST_PAGE_SIZE,
     pushNewRoute = noop,
+    setTab = true,
```
jwalgran: Consider the more specific `activateFacilitiesTab`

src/app/src/actions/facilities.js
```diff
@@ -58,10 +61,14 @@ export function fetchFacilities({
...
                 }
-
                 return data;
             })
-            .then(data => dispatch(completeFetchFacilities(data)))
+            .then((data) => {
+                dispatch(completeFetchFacilities(data));
+                if (setTab) {
+                    dispatch(makeSidebarFacilitiesTabActive());
+                }
```
jwalgran: Moving this out of the reducer into an action makes a lot of sense. Nice update to the design 👍

jwalgran:
Filtering by polygon is working correctly. Sorry for the false alarm.

src/app/src/actions/facilities.js
TaiWilkin: fixed in 2589cf0

TaiWilkin:
Most of the issues have now been resolved.

However, the hiding of the facilities search result count on some screen sizes I think will only be corrected by adjusting the breakpoints for the Grid component. This, though, is a bit twitchy and I'm not sure is the effect we want. 

Mar-11-2020 14-03-00

designmatty:
Let me take a look today.

designmatty:
Okay, here's what I would do:

https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/components/MapAndSidebar.jsx#L41-L44

Change these lines to:
```
<Grid item sm={12} md={4} id="panel-container">
    <Route component={SidebarWithErrorBoundary} />
</Grid>
<Grid item sm={12} md={8} style={{ position: 'relative' }} className="map-container" >
```

In this file https://github.com/open-apparel-registry/open-apparel-registry/blob/develop/src/app/src/styles/css/Map.css add the following changes:

```
@media (min-width: 775px) {
  #panel-container {
    min-width: 450px;
    flex: none;
  }

  .map-container {
    flex: 1 !important;
    max-width: initial !important;
  }
}
```

With these changes, the sidebar should no longer cut off and the "mobile" layout will happen much sooner.

TaiWilkin:
Thanks @designmatty, that works great!

Issue #998, opened 2020-03-20
Contributors with one or more matched facilities and one error are excluded from the filter dropdown


jwalgran:
## Overview

Contributors with one or more matched facilities and one error are excluded from the filter dropdown

### Expected Behavior

If a contributor has 1 or more active matches or new facilities, they appear in the filter dropdown.

### Actual Behavior

If contributor has any list items in an error state they do not appear in the filter dropdown.

### Steps to Reproduce

* Register for a new account and confirm
* Upload [one-bad-one-good.csv.zip](https://github.com/open-apparel-registry/open-apparel-registry/files/4361791/one-bad-one-good.csv.zip) and fully process it
  * ./scripts/manage batch_process --list-id 16 --action parse
  * ./scripts/manage batch_process --list-id 16 --action geocode
  * ./scripts/manage batch_process --list-id 16 --action match
* Browse http://localhost:6543/lists/16 and verify that there is one item in an error state and one successful item
* Browse http://localhost:6543. The newly registered contributor does not appear in the contributors dropdown.

PR #999, opened 2020-03-20
Include contributors if they have any active facilities

jwalgran:
## Overview

We were previously excluding contributors from the filter list if any of their items were in an error state. This meant that if a contributor submitted a list with one successfully processed item and one error item, the contributor would be filtered out.

To address this we are now checking for any item in a `MATCHED` or `CONFIRMED_MATCH` status. This has an additional benefit of excluding contributors from the list while their list is still processing.

Connects #998

## Testing Instructions

* Check out the `develop` branch
* Follow the instructions in #998 to reproduce the failure
* Check out this branch
* Refresh localhost:6543 and verify that the newly registered contributor appears in the contributors dropdown. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Removed
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


jwalgran:
Thanks for the review.

Issue #1033, opened 2020-06-19
Mobile optimization: explainer text for Contributor Overlap not appearing 


katieashaw:
## Overview

The explainer text for Contributor Overlap is not appearing when you click the 'i' button having selected multiple Contributors. 

### Expected Behavior

Pop up text explaining Contributor Overlap should appear 

### Actual Behavior

Nothing happens when you click the 'i' button - the explanatory text does not appear. 

### Steps to Reproduce

* Visit staging.openapparel.org
* Select multiple Contributors
* Click the 'i' button

### Demo

![1E13CCF9-F443-41E7-A164-44B5504598B6](https://user-images.githubusercontent.com/46344729/85114230-99561d80-b210-11ea-93ff-dcb41c4542c2.png)

### Additional context

This is the text which should appear (taken from Production)

![Screenshot 2020-06-19 at 09 39 05](https://user-images.githubusercontent.com/46344729/85114324-c60a3500-b210-11ea-91d3-41c3c8fcda43.png)

jwalgran:
Resolved by #1059

PR #1036, opened 2020-06-30
fixes mobile overflow, mobile tooltip

designmatty:
## Overview

- Fixed an issue where the "Show only shared facilities" wasnt showing on mobile
    - Added max-height and overflow scrolling within the tooltip for mobile
- Fixed an issue when filtering by Contributors with long names caused the app to overflow on mobile.

Connects #1033

## Demo

**Tooltip fix**
Screen Shot 2020-06-30 at 1 06 33 PM

**Overflow issue before**
Screen Shot 2020-06-30 at 1 12 10 PM

**Overflow issue after**
Screen Shot 2020-06-30 at 1 11 52 PM

## Testing Instructions

* interact with the tooltip on mobile
* Filter by a contributor with a long name. The input should either truncate or go to a new line.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/App.css
+++ b/src/app/src/App.css
@@ -702,6 +702,21 @@ hr {
   .facility-detail_map {
     width: auto;
   }
+
+  #panel-container {
+    width: auto;
+    max-width: 100%;
+    min-width: auto;
+  }
+
+  .sharedFacilityTooltip {
+    z-index: 99999999 !important;
+  }
+
+  .sharedFacilityTooltip-inner {
+    max-height: 80vh;
+    overflow: auto;
+  }
 }
 
 @media (max-width: 410px) {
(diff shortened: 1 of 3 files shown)


jwalgran:

designmatty:
I dont see anyway of fixing this with the current approach. Material's tooltips are not optimized for mobile by any stretch of the imagination. The default behavior for mobile, is such that opening the tooltip starts a countdown. Once that countdown hits 0 the tooltip close. Not ideal when we have a bunch of content and people read at different speeds.

We can get around this by utilizing the `disableTouchListener` prop. This eliminates the countdown. But it also eliminates the ability to touch outside of the tooltip to close it. There is one exception to this, which is what I was inadvertently doing in my tests (and which you discovered as well). If you focus on a focusable element (button, input, label, etc) the tooltip will close.

The only proper way of moving forward that I see, would be to rewrite this component with https://v3.material-ui.com/utils/popover/ instead. This unfortunately removes hover capabilities in favor of a click interaction, but I think this negative doesn't outweigh the positives for mobile.

jwalgran:
Thanks for such a detailed description of the problem.

I agree that turning this into a `Popover` is the right thing to do.

jwalgran:
Closing this because we will be reimplementing this feature as a popover.

Issue #1030, opened 2020-06-11
Add PPE fields and allow them to be submitted with lists and through the API


jwalgran:
## Overview

Add PPE fields and allow them to be submitted with lists and through the API

### Describe the solution you'd like

- Add nullablefields to the `FacilityListItem` and `Facility` models
  - `ppe_product_types` (`ArrayField` of chars, like claimed facility product types)
  - `ppe_contact_email` (`EmailField`)
  - `ppe_contact_phone` (char)
  - `ppe_website` (URLField)
- Update the parse task to read these fields from CSV columns
  - `ppe_product_types` is pipe separated
- Add fields to the csv and xls templates
- When a facility is created, copy the ppe_ fields
- When a match is promoted, update the ppe_ fields on the `Facility`

PR #1037, opened 2020-07-01
Add PPE-related model fields and API support

jwalgran:
## Overview

Adds PPErelated model fields and API support.

Connects #1030

## Demo

Screen Shot 2020-07-01 at 1 43 47 PM

Screen Shot 2020-07-01 at 1 43 21 PM

## Notes

We add contact information and product type fields to both the `FacilityListItem` and `Facility` models so that this additional metadata can be submitted in CSV files and passed through to the facility master list. The field types are based on similar fields in the claimed facility model and are nullable in this migration for backward compatibility.

We don't want to confuse contributors into thinking that the fields are required so we add a separate pair of CSV and XLSX templates that have additional PPE columns.

We fixed some code that assumed `source.facility_list` would always be non-None. When a single facility is POSTed to the API, the `source` is not associated with a list.

## Testing Instructions

These instructions assume `resetdb` has been run.

- Log in as c1@example.com, browse http://localhost:8081/admin/api/user/7/change/ and make that user a member of all the available groups.
- Download and unzip [ppe.csv.zip](https://github.com/open-apparel-registry/open-apparel-registry/files/4860579/ppe.csv.zip)
- Log in as `c7@example.com`, browse http://localhost:6543/profile/7, and generate and save for later an API token
- Browse http://localhost:6543/contribute and submit `ppe.csv`
- Fully process the file by running commands on the vagrant VM
- Browse http://localhost:6543/lists/16 and get the OAR IDs for the newly created Azavea facility and the matched facility in Bangladesh.
- Browse http://localhost:8081/api/docs/#!/facilities/facilities_read and submit the OAR IDs. 
  - Verify that the "Azavea" JSON response includes PPE fields.
  - Verify that the JSON response for the Bangledesh facility does not yet include the PPE fields.
- Browse http://localhost:8081/api/docs/#!/facilities/facilities_create, click `Authorize` and submit `Token {token-value}`.
- Use the Swagger UI to POST the following JSON. Verify that a match is returned.
```js
{
    "country": "Bangladesh",
    "name": "AKH Apparels Ltd JSON",
    "address": "28 Hamayetpur PS-Savar Dhaka-134",
    "ppe_product_types": ["Gowns", "Surgical Masks"],
    "ppe_contact_phone": "987-654-3210",
    "ppe_contact_email": "ppefromjson@example.com",
    "ppe_website": "https://ppefromjson.example.com"
}
```
- Log in as c1@example.com and browse http://localhost:6543/dashboard/adjustfacilitymatches
- Enter the OAR ID of the Bangladesh facility, promote one of the matches, then use http://localhost:8081/api/docs/#!/facilities/facilities_read to verify that the facility record now has PPE data.
  - Repeat with other match.

## Checklist

- [x] Issue created to follow up with a migration that makes nullable fields non-null.
- [x] CI passes after rebase


jwalgran:
Thanks for the review, and for mentioning that you were drafting comments about currently missing but expected behavior. Making a not to myself that when we break a feature into a "chain" of cards it is a good idea to clearly summarize the sub-features included in the current PR.

jwalgran:
I created #1038 to clean up nullable fields after this migration is deployed.

Issue #1043, opened 2020-07-08
Add a PPE data facility search parameter and a UI switch 


jwalgran:
## Overview

Add a `ppe` query parameter to the facility query API that filters the list of matching facilities to include only those that have a value in any of the `ppe_` fields. Add a checkbox to the filter panel connected to this query parameter.

### Describe the solution you'd like

- Add a `ppe` query parameter to the facility search API. Follow the semantics of other "flag" fields in the API
  - Update swagger docs appropriately
- Add a "Show only PPE producers" checkbox to the search page, connected to a "ppe" query parameter in the URL.

### Additional Context

~Replaces~ Augments #1031. After discussing the feature we realized that we needed this search in addition to a free text search.

PR #1044, opened 2020-07-09
Add PPE filter checkbox to the facility search page

jwalgran:
## Overview

Adds a PPE checkbox to the facility search page and connects it to the search API.

Connects #1043

## Demo

Screen Shot 2020-07-08 at 4 26 03 PM

## Testing Instructions

These instructions assume `resetdb` has been run.

- Download and unzip [ppe.csv.zip](https://github.com/open-apparel-registry/open-apparel-registry/files/4893793/ppe.csv.zip)
- Log in as `c7@example.com`, browse http://localhost:6543/contribute and submit `ppe.csv`
- Fully process the file by running commands on the vagrant VM
- Browse http://localhost:6543/facilities?ppe=true and verify that all 5 facilities from the CSV are returned.
- Clear the search and run a text search for "all" and verify that there are 11 results. Then check the PPE checkbox and search again. Verify that there is a single result.
- Browse http://localhost:8081/api/docs/#!/facilities/facilities_list and verify that the "ppe" field is documented.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/models.py
+++ b/src/django/api/models.py
@@ -20,6 +20,15 @@ from api.constants import Affiliations, Certifications, FacilitiesQueryParams
 from api.helpers import prefix_a_an
 
 
+class ArrayLength(models.Func):
+    """
+    A Func subclass that can be used in a QuerySet.annotate() call to invoke
+    the Postgres cardinality function on an array field, which returns the
+    length of the array.
+    """
+    function = 'CARDINALITY'
+
+
 class Version(models.Model):
     """
     A table storing feature version numbers.
@@ -1069,6 +1078,15 @@ class FacilityManager(models.Manager):
             FacilitiesQueryParams.BOUNDARY, None
         )
 
+        # The `ppe` query argument is defined as an optional boolean at the
+        # swagger level which is the built in field type option that most
+        # closely matches our desired behavior. Our intended use of the
+        # argument is conditionally "switch on" a special section of filter
+        # logic. To support that behavior we consider a missing value or any
+        # value than the string "true" to be `False`.
+        ppe = (True if params.get(FacilitiesQueryParams.PPE, '') == 'true'
+               else False)
+
         facilities_qs = Facility.objects.all()
 
         if free_text_query is not None:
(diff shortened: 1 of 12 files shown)


src/django/api/models.py
```diff
@@ -20,6 +20,15 @@
 from api.helpers import prefix_a_an
 
 
+class ArrayLength(models.Func):
+    """
+    A Func subclass that can be used in a QuerySet.annotate() call to invoke
+    the Postgres cardinality function on an array field, which returns the
+    length of the array.
+    """
+    function = 'CARDINALITY'
```
mmcfarland, line 29: Nice.

src/django/api/models.py
```diff
@@ -1069,6 +1078,9 @@ def filter_by_query_params(self, params):
             FacilitiesQueryParams.BOUNDARY, None
         )
 
+        ppe = (True if params.get(FacilitiesQueryParams.PPE, '') == 'true'
```
mmcfarland, line 1081: Confirming it's desired for `ppe` to only be `true` or `false` even though it is not required (in the API) and could be `None`. There may be no use case for `ppe == False` to return Facilities that don't have PPE and the value absent indicating that both PPE and non-PPE. I got a different understanding from the UI checkbox (binary) and the Swagger value selector (ternary), but that False and None give the same results.

jwalgran: I did discuss this with the team and we came to the conclusion that a search that excludes PPE facilities was not likely to be helpful given that the PPE feature set is designed to help people find PPE producers.

I will attempt to address this ambiguity in states with an inline comment and some additional clarification in the swagger help text.

jwalgran: Added in fixup 0bf9260

jwalgran:
While fixing up the Swagger docs I realized that we should be conditionally hiding the `ppe` query argument documentation based on the feature flag. I made this simple change in `4bc25b3` and tested it.

jwalgran:
Thanks for the review.

Issue #1050, opened 2020-07-15
Copy PPE fields to facility record when matching if the fields are empty


jwalgran:
## Overview

Currently we only copy PPE fields to the `Facility` record when creating a new facility or when promoting a facility match. After discussing the situation with stakeholders we decided that we should copy PPE fields from a facility match if the corresponding PPE field on the `Facility` is blank.

### Is your feature request related to a problem? Please describe.

We do not want to require moderation actions (promotion) in order to searchable PPE fields.

### Describe the solution you'd like

* Update the matching process to copy PPE field values that have not already been set on the `Facility`
* When splitting a facility match, remove PPE field values from the `Facility` that were copied from the list item that is being separated.

Issue #1048, opened 2020-07-09
Revise PPE matching process


jwalgran:
## Overview

Discuss how PPE fields are handled and modify how they are copied from `FacilityListItem` to `Facility` if needed.

jwalgran:
Superseded by #1050

PR #1055, opened 2020-07-17
Update PPE fields when matching, splitting, and merging

jwalgran:
## Overview

When creating matches to existing facilities, splitting a match, or merging two facilities, update the PPE fields as needed

~Connects #1048~
Connects #1050

## Notes

After discussing how PPE fields should be handled in the case of multiple matches we decided that it is OK to break with the previous behavior of all the properties of `Facility` being copied from a single `FacilityListItem`. We want to be able to add PPE information to facilities by uploading a list and matching items to existing facilities, but we don't want to require promoting matches since it will:
  1. Require extra work
  2. May overwrite a better name or address in the process

When this PR is merged, matching a list item to a facility wither by automatic match or a manual merge, will copy PPE fields from the item if they have not already been set on the `Facility`. Implementing this behavior required us to also modify the split action to restore/clear PPE field values by copying them back from the list item that originally created the `Facility`. This isn't a 100% correct solution in all cases since the PPE field values could be from multiple matches, but we are not expecting that to be a situation that occurs in practice.

## Testing Instructions

* Run ./scripts/resetdb
* Open two browser sessions, and admin session logged in as c1@example.com and a normal session logged in as c2@example.com
* In the c2@example.com session browse http://localhost:6543/contribute and upload 
[clarksville-hosiery-ppe.csv.zip](https://github.com/open-apparel-registry/open-apparel-registry/files/4940223/clarksville-hosiery-ppe.csv.zip)
* Process the list
* Browse http://localhost:6543/lists/16, verify that the row `MATCHED`, and click the link to navigate to the detail page. Verify that the PPE fields are populated. Copy the OAR ID.
* In the c1@example.com session browse http://localhost:6543/dashboard/adjustfacilitymatches and submit the copied OAR ID.
* Split the match and verify that it succeeds. Browse the new facility detail page and verify that the PPE data is listed. Browse the old facility detail page and verify that the PPE data has been removed. Copy the new OAR ID.
* Browse http://localhost:6543/dashboard/mergefacilities, enter the old OAR ID first, then the new OAR ID created when the match was split (The "clarksville-hosiery-ppe" item should be on the right)
* Merge the facilities then browse the detail page of the old OAR ID. Verify that the PPE details have returned.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Show PPE fields on facility detail and include in CSV downloads [#1041](https://github.com/open-apparel-registry/open-apparel-registry/pull/1041)
 - Add PPE filter checkbox to the facility search page [#1044](https://github.com/open-apparel-registry/open-apparel-registry/pull/1044)
 - Include PPE product types in free text query [#1045](https://github.com/open-apparel-registry/open-apparel-registry/pull/1045)
+- Update PPE filter checkbox label / allow viewing RequestLog in the Django admin / log matching traceback [#1049](https://github.com/open-apparel-registry/open-apparel-registry/pull/1049)
+- Update PPE fields when matching, splitting, and merging [#1055](https://github.com/open-apparel-registry/open-apparel-registry/pull/1055)
 
 ### Changed
-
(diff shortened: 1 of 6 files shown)


jwalgran:
Thanks for the review.

>Should any of this functionality be included within the existing feature flag for the PPE Feature?

A good question. I have been making decisions on what to wrap in feature flag checks and what not to based on user visibility. The data storage processing related to PPE is not gated, but representations of the data returned to clients is.

Issue #1061, opened 2020-07-29
The parse job crashes if one of the values of the ppe_product_types array field exceeds the max length of 50 characters


jwalgran:
## Overview

The parse job crashes if one of the values of a `ppe_product_types` array field exceeds the max length of 50 characters.

### Expected Behavior

When processing a list if one of the values of a `ppe_product_types` array field exceeds the max length of 50 characters the processing job completes successfully and the line is moved to the`ERROR_PARSING` status and shows an informative validation error message.

### Actual Behavior

When processing a list if one of the values of a `ppe_product_types` array field exceeds the max length of 50 characters the batch_process job crashes with an unhanded exception and the line is stuck in the `UPLOADED` status.

### Steps to Reproduce

These steps assume `./scripts/resetdb` has been run.

* Log in as c2@example.com
* Upload [single-item-ppe-product-type-too-long.csv.zip](https://github.com/open-apparel-registry/open-apparel-registry/files/4997568/single-item-ppe-product-type-too-long.csv.zip)
* Run the parse job
  * ./scripts/manage batch_process --list-id 16 --action parse

### Demo

```
vagrant@vagrant:/vagrant$ ./scripts/manage batch_process --list-id 16 --action parse
Starting vagrant_database_1 ... done
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.InvalidTextRepresentation: malformed array literal: ""
LINE 1: ... "api_facilitylistitem" SET "ppe_product_types" = ''::varcha...
                                                             ^
DETAIL:  Array value must start with "{" or dimension information.


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "manage.py", line 15, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 364, in execute
[...]

PR #1062, opened 2020-07-29
Prevent crashing the parse job if ppe_product_types value too long

jwalgran:
## Overview

If there is a validation error on the `ppe_product_types` array field, `full_clean` appears to set it to an empty string which then causes `save` to raise an exception. We work around this by setting the field to an empty list if it is not None and not a list.

Connects #1061 

## Demo

Screen Shot 2020-07-29 at 2 39 13 PM

## Testing Instructions

These steps assume `./scripts/resetdb` has been run.

* Log in as c2@example.com
* Upload [single-item-ppe-product-type-too-long.csv.zip](https://github.com/open-apparel-registry/open-apparel-registry/files/4997568/single-item-ppe-product-type-too-long.csv.zip)
* Run the parse job
* Verify the job reports a failed line but exits successfuly.
* Browse http://localhost:6543/lists/16, click the row, and verify that an informative validation error is shown.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Prevent crashing the parse job if ppe_product_types value too long [#1062](https://github.com/open-apparel-registry/open-apparel-registry/pull/1062)
+
 ### Security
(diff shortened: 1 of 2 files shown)


jwalgran:
Thanks for the review.

PR #1067, opened 2020-07-30
Expand ppe_product_types and ppe_contact_phone

jwalgran:
## Overview

After processing real world data we are seeing product types longer than 50 characters and multiple phone numbers combined into the same field. The phone number data is not consistently delimited, so we are opting to just expand the field rather than enforce delimiting rules that make processing of the source data challenging.

Connects #1061 

## Demo

Screen Shot 2020-07-30 at 9 14 12 AM

## Testing Instructions

* Run `./scripts/resetdb`, which will apply the migration.
* Log in as c2@example.com
* [Browse http://localhost:654]3/contribute and upload 
[expanded-field-demo.csv.zip](https://github.com/open-apparel-registry/open-apparel-registry/files/5002457/expanded-field-demo.csv.zip)
* Verify that you can fully process the list
* Browse http://localhost:6543/lists/16 and click the link to the facility. Verify that the PPE data is correctly set on the facility.

## Checklist

- [x] CI passes after rebase


jwalgran:
Thanks for the review.

Issue #1065, opened 2020-07-30
Confirmed/rejected matches do not update PPE fields


jwalgran:
## Overview

When a pending match is confirmed or rejected, the PPE data is not copied to the existing or newly created `Facility`.

### Expected Behavior

Confirming or rejecting a match behaves the same as an automatic match. PPE data is copied to the `Facility`.

### Actual Behavior

When a pending match is confirmed, or a new facility is created as the result of rejecting a match, the PPE data is not copied to the `Facility`.

### Steps to Reproduce

* Upload a list that includes pending matches and includes PPE data
* Confirm a pending match
* Reject a pending match

jwalgran:
@katieashaw I discovered a problem with confirmed matches not updating the PPE fields when I was taking a look at the processing results of the "PPE: OAR community submissions & desk research" list.

Issue #1058, opened 2020-07-22
Update PPE fields when matches are deactivated by either list replacement or the list item remove button.


jwalgran:
## Overview

If a facility stops producing PPE, we need to remove the PPE details from the facility record. Adapt the behavior we implemented when splitting matches and reuse it when deactivating matches.

### Describe the solution you'd like

* Update the replacement list processing and list item removal code to revert PPE fields on the `Facility`.

PR #1069, opened 2020-07-30
Update PPE fields when deactivating sources and matches or confirming/rejecting matches

jwalgran:
## Overview

Ensures that PPE fields on the `Facility` are updated appropriately when the following actions are taken:

- Confirming a potential match.
- Rejecting a potential match.
- Removing (deactivating) a single item from a list.
- Replacing a list.

Connects #1058 
Connects #1065

## Notes

When a line item is not longer associated with a facility, we revert the PPE field values on the `Facility`, following the existing workflow of splitting a match.

We choose to trigger this logic in overridden `save` methods on `Source` and `FacilityMatch` so that we will apply the same reversion logic any time a block of code changes `is_active` from `True` to `False`. The `is_active` flag is sometimes changed by moderators via the Django admin so we need a solution that will work even when the change in active status is triggered by code that we do not control.

To reduce duplication we moved the set of `Q` statements that we use to filter only facilities or items that have one or more PPE fields filled in to a constant and we added a method to `Facility` handles the conditional reverting of fields.

## Testing Instructions

* Download and unzip [lists-for-testing-1069.zip](https://github.com/open-apparel-registry/open-apparel-registry/files/5008269/lists-for-testing-1069.zip)
* Run `./scripts/resetdb`
* Log in as c2@example.com
* Browse http://localhost:6543/contribute and upload `list-8-with-ppe.csv`. 
* Fully process the list
* Browse http://localhost:6543/lists/16 and confirm the potential match at row index 10. Click the row and then click the facility link. Confirm that the PPE data is populated.
* Browse http://localhost:6543/lists/16 and reject the potential match at row index 11. Click the facility link and confirm that the newly created facility has populated PPE data.
* Browse http://localhost:6543/contribute and replace the previous list by uploading `list-8.csv` and choosing `list-8-with-ppe` in the "Select list to replace" dropdown. 
* Fully process the list.
  * ./scripts/manage batch_process --list-id 17 --action parse
  * ./scripts/manage batch_process --list-id 17 --action geocode
  * ./scripts/manage batch_process --list-id 17 --action match
* Browse http://localhost:6543/lists/17, click a MATCHED row and follow the link. Verify that the PPE data is no longer preset.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
+- Update PPE fields when deactivating sources and matches or confirming/rejecting matches [#1069](https://github.com/open-apparel-registry/open-apparel-registry/pull/1069)
+
 ### Deprecated
 
 ### Removed
--- a/src/app/src/components/FacilityDetailSidebarPPE.jsx
+++ b/src/app/src/components/FacilityDetailSidebarPPE.jsx
@@ -8,8 +8,9 @@ import { PPE_FIELD_NAMES } from '../util/constants';
 
 const FacilityDetailSidebarPPE = ({ properties }) => {
     const ppeFields = pickBy(pick(properties, PPE_FIELD_NAMES), identity);
+    const hasPPEValues = !Object.values(ppeFields).every(isEmpty);
 
-    return !isEmpty(ppeFields)
+    return hasPPEValues
         ? (
             <div className="control-panel__group">
                 <h1 className="control-panel__heading">
--- a/src/django/api/models.py
+++ b/src/django/api/models.py
@@ -311,6 +311,24 @@ class Source(models.Model):
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
+    def __init__(self, *args, **kwargs):
+        super(Source, self).__init__(*args, **kwargs)
+        self.__original_is_active = self.is_active
+
+    def save(self, force_insert=False, force_update=False, *args, **kwargs):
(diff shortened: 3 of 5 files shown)


mmcfarland:
I'm not getting the results I expected after running through the instructions, but I'm not sure if my expectations are correct or if this is not functioning as intended.

Observations after "with PPE" list uploaded:
* MATCHed facilities do show PPE data ("Mask{rownum}")
* CONFIRMed facility does show PPE data
* REJECTED facility: potential match does not show PPE after rejection, new facility does

This all seems correct to me.

Observations after "without PPE" list uploaded, replacing "with PPE":
* MATCHed facility still showing PPE, even though contributed from "without PPE" list
   image

My understanding is that the previously matched facilities would no longer have PPE values on them. Is this working as intended? I ran through the workflow twice, with a `resetdb` in between.

jwalgran:
Short version: this is working as intended, but I now want to make a small change tot he implementation based on your feedback.

The screenshot shows the "Masks12" item. Following the test instructions, that is the one facility where we reject the potential match, creating a new facility.

The way things are currently written, when removing an association. the PPE values are rolled back to the values that are set when the facility was first created. For brand new facilities, this is a noop. The PPE values stay.

This does make it impossible for a contributor to clear the PPE data for a facility, so I am going to modify the code to clear out the PPE fields if the item from which the facility was created is being deactivated.

jwalgran:
I implemented and pushed up an additional to handle removing PPE fields from the facility when the item that created the facility is deactivated. This is ready for another look.

jwalgran:
Thanks for catching that stray header. I pushed 4d86329 with a fix for it. `[]` is truthy, so the previous `isEmpty` check was not working the way we wanted it to.

jwalgran:
Thanks for the review.

PR #1086, opened 2020-08-19
Move AWS specific Gunicorn settings to ECS task definition

### LINKED ISSUE #1082: Temporarily increase ECS service count in staging
Aim to increase ECS service count so that it aligns with the production environment. In addition, consider applying the following patch to increase the Gunicorn worker count:

See: https://github.com/azavea/fieldscope/pull/489

rbreslow:
## Overview

Specify the Gunicorn configuration settings tuned for AWS ECS directly in the task definition for the Django application. These values have been ported over from the Dockerfile, except `--workers`, which is now derived by the following formula:

```
  (2 x vCPUs) + 1 = (2 x 1) + 1 = 3
```

See: https://docs.gunicorn.org/en/stable/design.html#how-many-workers

Resolves #1082 

## Testing Instructions

I scaled `app_fargate_cpu` to `1024`. See that Terraform worker count is at `3`.

image

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/deployment/terraform/container_service.tf
+++ b/deployment/terraform/container_service.tf
@@ -100,15 +100,16 @@ data "template_file" "app" {
   template = "${file("task-definitions/app.json")}"
 
   vars = {
-    cpu    = "${var.app_fargate_cpu}"
     image  = "${local.app_image}"
-    memory = "${var.app_fargate_memory}"
 
     postgres_host     = "${aws_route53_record.database.name}"
     postgres_port     = "${module.database_enc.port}"
     postgres_user     = "${var.rds_database_username}"
     postgres_password = "${var.rds_database_password}"
     postgres_db       = "${var.rds_database_name}"
+
+    # See: https://docs.gunicorn.org/en/stable/design.html#how-many-workers
+    gunicorn_workers = "${ceil((2 * (__builtin_StringToFloat(var.app_fargate_cpu) / 1024)) + 1)}"
 
     google_server_side_api_key = "${var.google_server_side_api_key}"
     google_client_side_api_key = "${var.google_client_side_api_key}"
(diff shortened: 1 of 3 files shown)


deployment/terraform/task-definitions/app.json
```diff
@@ -2,8 +2,18 @@
   {
     "name": "django",
     "image": "${image}",
-    "cpu": ${cpu},
-    "memory": ${memory},
```
rbreslow, line 6: Specifying these limitations within the task definition template is redundant with Fargate.

deployment/terraform/task-definitions/app_cli.json
```diff
@@ -2,9 +2,7 @@
   {
     "name": "django",
     "image": "${image}",
-    "cpu": ${cpu},
-    "memory": ${memory},
-    "essential": true,
```
rbreslow, line 7: This is the only container in the task, so it doesn't matter if we mark it as `essential` or not.

deployment/terraform/container_service.tf
```diff
@@ -100,16 +100,17 @@ data "template_file" "app" {
...
 
     postgres_host     = "${aws_route53_record.database.name}"
     postgres_port     = "${module.database_enc.port}"
     postgres_user     = "${var.rds_database_username}"
     postgres_password = "${var.rds_database_password}"
     postgres_db       = "${var.rds_database_name}"
 
+    # See: https://docs.gunicorn.org/en/stable/design.html#how-many-workers
+    gunicorn_workers = "${ceil((2 * (__builtin_StringToFloat(var.app_fargate_cpu) / 1024)) + 1)}"
+
```
rbreslow, line 113: UwU https://github.com/hashicorp/terraform/issues/14967#issuecomment-305277001

ColeKettler: 👀

Issue #1101, opened 2020-09-09
Two Contributors are appearing on Production with no facilities attached to their account


katieashaw:
## Overview

Two Contributors are appearing in the Contributor dropdown menu on Production with no facilities attached to their account. 

### Expected Behavior

Contributors which have not uploaded facilities to the OAR should not appear in the Contributor dropdown. 

### Actual Behavior

Two Contributor accounts which have not uploaded facility lists to the OAR are appearing in the Contributor dropdown menu. 

### Demo

![Screenshot 2020-09-03 at 16 01 23](https://user-images.githubusercontent.com/46344729/92624575-93c41d00-f2bf-11ea-85db-988a759d8c74.png)

![Screenshot 2020-09-03 at 15 51 24](https://user-images.githubusercontent.com/46344729/92624594-9888d100-f2bf-11ea-9336-c8724d20f2ce.png)

### Additional context

Could you please confirm with the Rushan Dexin Garment Co entry that processing of a list has not stalled / got blocked whilst in progress.

PR #1114, opened 2020-09-23
Exclude contributors if all sources are create=False

jwalgran:
## Overview

If a contributor only submits data via the API and has submitted all facilities with the `create` option set to false then they will not have any publicly visible data and should therefore be excluded from the contributors list.

Connects #1101

## Testing Instructions

* Checkout `develop`
* Run `./scripts/resetdb`
* Register a new account. If the confirmation email does not appear you can manually activate the account with
```
vagrant@vagrant:/vagrant$ echo "UPDATE api_user SET is_active = 't' WHERE is_active = 'f';" | ./scripts/manage dbshell
vagrant@vagrant:/vagrant$ echo "UPDATE account_emailaddress SET verified = 't' WHERE verified = 'f';" | ./scripts/manage dbshell
```
* Log in as the **admin** user c1@example.com and browse http://localhost:8081/admin/api/user/101/change/
* Add the newly register user to all the groups and save.
* Log in as the new user, browse http://localhost:6543/profile/101 and create an API token
* Browse http://localhost:8081/api/docs/#!/facilities/facilities_create and "Authorize" with the token, making sure that you enter `Token abc123...`
* Set `create` to `false` and POST the following
```
{
    "country": "US",
    "address": "990 Spring Garden, Philadelphia",
    "name": "Azavea"
}
```
* Browse http://localhost:6543/ and confirm the incorrect behavior. The new user will be listed in the Contributors dropdown but clicking "Search" after selecting that contributor returns no results.
* Check out this branch and refresh http://localhost:6543/. Verify that the contributors list is still populated but the new user is excluded.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - Return distinct source names from the facility API [#1113](https://github.com/open-apparel-registry/open-apparel-registry/pull/1113)
+- Exclude contributors if all sources are create=False [#1114](https://github.com/open-apparel-registry/open-apparel-registry/pull/1114)
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


jwalgran:
Pushed a rebase with a lint fixup

rajadain (approved):
+1 tested, working well!

jwalgran:
Thanks for the review.

PR #1123, opened 2020-10-07
Increase app_ecs_grace_period_seconds

jwalgran:
## Overview

We were seeing instances not become healthy within 120 seconds, which put us into a loop. We assume that we did not catch this because we were using an experimental value of 6 during the testing of recent changes to the service startup process and then returned it to value of 2.

To choose a new value we modified the grace period through the console and found 150 still resulted in tasks not reaching a healthy status, but 180 allowed sufficient time.

Connects open-apparel-registry/open-apparel-registry-clients#11

## Testing Instructions

* View the staging cluster in the console after this branch is deployed and verify that the instances stabilize.

## Checklist

- [x] DROP commits are dropped
- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
-Directly connect health check to GazetteerCache and remove the use of threading in model training [#1117](https://github.com/open-apparel-registry/open-apparel-registry/pull/1117)
+Directly connect health check to GazetteerCache and remove the use of threading in model training [#1117](https://github.com/open-apparel-registry/open-apparel-registry/pull/1117) [#1123](https://github.com/open-apparel-registry/open-apparel-registry/pull/1123)
 
 ### Deprecated
 
--- a/deployment/terraform/variables.tf
+++ b/deployment/terraform/variables.tf
@@ -173,7 +173,7 @@ variable "app_ecs_deployment_max_percent" {
 }
 
 variable "app_ecs_grace_period_seconds" {
-  default = "120"
+  default = "180"
 }
 
 variable "app_fargate_cpu" {


Jenkinsfile
```diff
@@ -28,7 +28,7 @@ node {
 
 		env.OAR_SETTINGS_BUCKET = 'openapparelregistry-staging-config-eu-west-1'
 
-		if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/')) {
+		if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME == 'PR-1123') {
```
jwalgran, line 31: Reminder to drop this change before deployment

deployment/terraform/dns.tf
```diff
@@ -20,7 +20,7 @@ resource "aws_route53_record" "database" {
   name    = "database.service.${var.r53_private_hosted_zone}"
   type    = "CNAME"
   ttl     = "10"
-  records = ["${module.database_enc.hostname}"]
+  records = ["openapparelregistry-prd-clone.ctdp4povekhv.eu-west-1.rds.amazonaws.com"]
```
jwalgran, line 23: Reminder to drop this change before merging

hectcastro (approved):
> We were seeing instances not become healthy within 120 seconds, which put us into a loop. We assume that we did not catch this because we were using an experimental value of 6 during the testing of recent changes to the service startup process and then returned it to value of 2.

Just wanted to clarify for future us that the 6 and 2 here refer to the `unhealth_threshold` setting of the service target group health checks. Presumably having that set to 6 was mimicking a longer grace period than 120s.

👍

Issue #1111, opened 2020-09-22
Make the replica of production the permanent staging database


rbreslow:
Working through #1084, we set up a replica of the production database with sanitized email addresses. Folks at OAR were happy to have fresher data in staging and already went through the process of recreating the accounts they use there.

Making this new database the permanent staging database makes sense. The new database is a replica of production, so it's `db.t3.small` and multi-AZ vs. the smaller `db.t3.micro` and single-AZ. It also was upgraded to PostgreSQL 12.

The following questions are left unanswered:
- Do we want to leave it as-is, or scale it back?
- The upgrade process was fairly straight forward, but I didn't document anything as I went through. Do we want to keep the old database around so that we can walk through a PostgreSQL 12 upgrade with more-accurate info on the length of downtime?

The following things need to be done:
- Manipulate the Terraform state to remove the old database / parameter group and slot in the new set of resources.

rbreslow:
I removed the old RDS instance from the Terraform state and then renamed it in the AWS console with the suffix `-old`. I renamed the new RDS instance to the old RDS instance's original name and imported it back into the Terraform state. Then, I cycled Terraform using the changes in #1125 to create and apply a new parameter group that is codified in Terraform. 

These AWS docs served as a useful sanity check: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_RenameInstance.html#USER_RenameInstance.RR

Everything looks clean now. I left the old database instance in a stopped state just in case and set up a 2 week Slack reminder to delete it in its entirety.

rbreslow:
I've moved this to `in review`. @hectcastro If you could take a look at #1125, the description of events in this issue, and the state of things in the AWS console, I think we could then close this out.

hectcastro:
Because https://github.com/open-apparel-registry/open-apparel-registry/pull/1125 makes the current staging database default to PostgreSQL 12, I think that now we need to accelerate the effort to get development and production on PostgreSQL 12 as well. There is a draft issue https://github.com/open-apparel-registry/open-apparel-registry/issues/948 to upgrade to PostgreSQL 12, but I think we may need to split that into at least two separate issues (one for upgrading to PostgreSQL 12 in development, and another for production). After they exist, we should aim to slot them into next sprint.

rbreslow:
These are up and slotted in.

hectcastro:
Can the `openapparelregistry-enc-stg-old` RDS instance be taken offline now?

[...]

PR #1125, opened 2020-10-08
Make it possible to modify the RDS DB parameter group

rbreslow:
## Overview

Add support for [name_prefix](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_parameter_group#name_prefix) and the [create_before_destroy](https://www.terraform.io/docs/configuration/resources.html#create_before_destroy) lifecycle customization so that a new parameter group is created and applied before disposing of the old one when making modifications that require a new resource.

Connects #1111 

## Testing Instructions

See that there are no Terraform changes to apply on staging:

```bash
$ docker-compose -f docker-compose.ci.yml run --rm terraform
bash-4.4# GIT_COMMIT=d39830b ./scripts/infra plan
. . .
No changes. Infrastructure is up-to-date.
```

And the new parameter group is in place:

image

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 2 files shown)


hectcastro (approved):
Test played out as expected. 👍

Issue #1137, opened 2020-10-16
Adjust reports to not include create=false


dboyer:
Records added with create=false should not be included in any report other than API calls. Including them does not give an accurate representation of the geocode results and other metrics associated with data added to OAR.

PR #1142, opened 2020-10-20
Update reports to consider sources and source.create

jwalgran:
## Overview

We introduced the `Source` model to unify list-based contributions and one-at-a-time contributions from the API. We had not updated our reports to incorporate this change, and we are now starting to get real data from API contributions.

In addition to focusing more on `Source` we also need to consider the `Source.create` property. We allow API users to set `create=false` on their POST requests which prevents the creation of `Facility` and `FacilityMatch` records. These "test" requests were inflating the numbers in some of the reports.

Connects #1137

## Notes

Here is the complete review of all the existing reports. Reports marked with ✅ were added or updated in this PR, but the output may not have changed.

- Api Request By User
  - No change desired
- Api Requests
  - No change desired
- Api Tokens
  - No change required
-  ✅ Average List Count Per Contributor
  - Results unchanged
- Average Matches Per Facility
  - Unchanged
  - Matches are not created when `create=false`
- Average Matches Per Facility By Type
  - Unchanged
  - Matches are not created when `create=false`
- ✅ Confirmed Matches
  - Results unchanged
  - Matches are not created when `create=false`
- ✅ Contributor First Appearance 
  - Discovered a “bug”
  - This report was list-centric
- Contributor List Count 
  - No change desired
  - This report is list focused
- ✅ Contributor Source Count
  - New report based on Contributor List Count but using sources with create=true
- ✅ Contributors With Active Facilities
  - Results unchanged
- Contributors
  - No change desired.
  - This is just a “dump” of the contributors table.
- Country First Appearance
  - No change desired.
- Facilities Created Not Matched
  - Unchanged. 
  - `create=false` does not create facility records.
- Facilities Matched
  - Unchanged. 
  - `create=false` does not create match records.
- Facility Country Count 
  - No change desired.
- Facility Counts By Country
  - No change desired.
- ✅ Facility List Items Uploaded
  - Updated to use `Source`
- Facility Lists Uploaded
  - Unchanged
  - This report is list-centric and lists are always `create=true`
- ✅ Geocode Errors
  - Updated to exclude `create=false` by joining `Source`
- ✅ Geocode No Results
  - Updated to exclude `create=false` by joining `Source`
Mailing List Recipients
 - No change desired.
- ✅ Non Public Contributors
  - Discovered a “bug”
  - This report was list-centric
  - The new version includes a source type column and picks up contributions via the API with create=true
- ✅ Public Contributors
  - No real change to the numbers, but we updated the report to match the non-public contributors report and not be list-centric in case we end up using the API to add public data in the future
- Rejected Matches
  - No change desired.
  - `create=false` does not create match records
- ✅ Unique Contributor Count By Type
  - Discovered a “bug”
  - This report was list-centric and now includes API only contributors with create=true
- Unique Public List Count
  - No change desired. 
  - This is a list-centric report.
- Updated Lists
  - No change desired. 
  - This is a list-centric report.

## Testing Instructions

NOTE: If you are using the development data created from ./scripts/resetdb then some of these reports are expected to be empty.

* Do a visual audit of the SQL statements and verify them for correctness
* Log in as c1@example.com
* Browse http://localhost:8081/admin/reports/ and verify that all the reports run without error

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 ### Changed
+
 - Update `django-ecsmanage` to version 2.0.0 [1129](https://github.com/open-apparel-registry/open-apparel-registry/pull/1129)
+- Update reports to consider sources and source.create [#1142](https://github.com/open-apparel-registry/open-apparel-registry/pull/1142)
 
 ### Deprecated
 
--- a/src/django/api/reports/average_list_count_per_contributor.sql
+++ b/src/django/api/reports/average_list_count_per_contributor.sql
@@ -12,6 +12,7 @@ FROM (
   WHERE to_char(l.created_at, 'YYYY-MM') != to_char(now(), 'YYYY-MM')
   AND s.is_public = 't'
   AND s.is_active = 't'
+  AND s.create = 't'
   GROUP BY to_char(l.created_at, 'YYYY-MM'), c.id
 ) s
 GROUP BY month
--- a/src/django/api/reports/confirmed_matches.sql
+++ b/src/django/api/reports/confirmed_matches.sql
@@ -1,8 +1,10 @@
 SELECT
   to_char(li.created_at, 'YYYY-MM') AS month,
   COUNT(*) AS list_item_count
-from api_facilitylistitem li
+FROM api_facilitylistitem li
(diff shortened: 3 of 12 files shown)


TaiWilkin (approved):
This is working well for all the reports for which I have data, and the SQL statements generated for other reports looks correct when I read them over.

jwalgran:
Thanks for the review.

PR #1152, opened 2020-10-26
improved UX on facility list tables

designmatty:
## Overview

- Better highlighting/visualization of certain status'
- Expanded row content will now render in a new table row instead of as extra divs within columns. This will fix alignment issues and overflow issues.
- Fixed table layouts. Several columns now have `colspans` defined which prevent shrinking/expanding based on content.
- Removed truncation and tooltips from longer addresses and names. These are longer required.

Connects #XXX

## Demo

**Highlight areas which require specific actions:**
Screen Shot 2020-10-23 at 4 07 06 PM

**Highlight rows which have expandable rows**
Screen Shot 2020-10-23 at 4 07 19 PM
Screen Shot 2020-10-23 at 4 07 23 PM

**Improved highlighting of rows with errors**
Screen Shot 2020-10-23 at 4 07 30 PM
Screen Shot 2020-10-23 at 4 07 35 PM

**Deleted rows will now be crossed-out**
Screen Shot 2020-10-23 at 4 08 07 PM

**Deleted rows that are expandable will still be highlighted**
Screen Shot 2020-10-23 at 4 08 11 PM
Screen Shot 2020-10-23 at 4 08 15 PM

**Long names/addresses will no longer truncate and tooltip.**
Screen Shot 2020-10-23 at 4 09 47 PM

## Notes

I was originally going to try and refactor the code-base for these layouts. This proved to be a time-consuming and confusing process. So instead of spending a ton of time trying to rewrite, I instead focused on usability. 

There is a ton of duplicate code and functionality, but I felt improving the UX was a better way of spending time.

One UX issues to note: Some rows can be "expanded" but they don't have any extra content to reveal. I never figured out how to resolve this. 

## Testing Instructions

* Open up facility list tables
* Things to check:
    * Long names and address' (should not truncate)
    * Deleted rows (should be crossed-out)
    * Errors (should be highlighted in red)
    * Potential Match (should be highlighted blue)
    * Expandable rows (should render new row when open. Alignment issues should be resolved. Titles and content will be lined up correctly)

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/App.css
+++ b/src/app/src/App.css
@@ -725,6 +725,14 @@ hr {
   border: none;
 }
 
+.STATUS_POTENTIAL_MATCH td {
+  border: none !important;
+}
+
+.STATUS_POTENTIAL_MATCH + .STATUS_POTENTIAL_MATCH--SUB-ROW td {
+  border-bottom: 1px solid #e0e0e0 !important;
+}
+
 .facility-list-row__expanded + .facility-list-row__expanded td {
   border-bottom: 1px solid #e0e0e0
 }
@@ -747,4 +755,8 @@ hr {
 
 .STATUS_PARSED:not(:hover) {
   background: #fff !important;
+}
+
+.STATUS_POTENTIAL_MATCH--ACTIONS + .STATUS_POTENTIAL_MATCH--ACTIONS {
+  display: none !important;
 }
--- a/src/app/src/App.css
+++ b/src/app/src/App.css
@@ -720,3 +720,31 @@ hr {
   font-size: 12px !important;
   margin: 0 !important;
 }
+
+.facility-list-row__expanded td {
+  border: none;
+}
+
+.facility-list-row__expanded + .facility-list-row__expanded td {
+  border-bottom: 1px solid #e0e0e0
+}
+
+.STATUS_MATCHED_COLLAPSED {
+  background: rgb(243, 243, 243);
+}
+
+.STATUS_MATCHED_COLLAPSED.STATUS_ERROR:hover {
+  background: #ffc7c7 !important;
+}
+
+.STATUS_ERROR {
+  background: #ffe2e2;
+}
+
+.STATUS_ERROR td {
+  border-color: #ffe2e2;
+}
+
+.STATUS_PARSED:not(:hover) {
+  background: #fff !important;
+}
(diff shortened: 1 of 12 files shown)


jwalgran:
I saw content in the address column breaking out into the status column

Screen Shot 2020-10-28 at 9 06 59 PM

I reproduced this by:
* Running ./scripts/resetdb
* Logging in as c8@example.com with password "password"
* Browse http://localhost:6543/lists/8 and scroll to list index 5

src/app/src/App.css
```diff
@@ -719,4 +719,32 @@ hr {
...
+
+.facility-list-row__expanded td {
+  border: none;
+}
+
+.facility-list-row__expanded + .facility-list-row__expanded td {
+  border-bottom: 1px solid #e0e0e0
+}
+
+.STATUS_MATCHED_COLLAPSED {
```
jwalgran, line 732: Consider updating the comment in api/models.py and explain that if a new status is added we should update this stylesheet as well https://github.com/open-apparel-registry/open-apparel-registry/blob/783e782349255d1ea10241c56983c2dd07f753b0/src/django/api/models.py#L489-L490

src/app/src/components/FacilityListItemsTable.jsx
```diff
@@ -481,6 +487,7 @@ class FacilityListItemsTable extends Component {
                         handleSelectRow={handleSelectRow}
                         hover
                         isRemoved={anyListItemMatchesAreInactive(item)}
+                        className={`STATUS_${item.status}`}
```
jwalgran, line 490: Nice, compact way of doing conditional styling based on status.

jwalgran (approved):
Great looking revisions. I showed the screenshots to the OAR moderation team and they were looking forward to the workflow improvements.

designmatty:
- I added a word-break to the address columns which will prevent the line from overflowing:
image

_Note: I think a better approach might be to utilize this regex `\,[^ ]` to find commas without spaces after and then inject a space with react. But I'm unsure if this would be a technically correct approach. [See demo](https://regex101.com/r/iunjX2/1)_

- I noticed in situations where both the address and facility name changed, we rendered two sets of confirm/reject actions. I took the simple approach and hid the duplicate actions with css.
**Before:**
image
**After:**
image

jwalgran:
>Note: I think a better approach might be to utilize this regex \,[^ ] to find commas without spaces after and then inject a space with react. But I'm unsure if this would be a technically correct approach. 

I like the use of `wordBreak` in CSS. Our approach so far has been that we never alter the data that is submitted. Since this is a display issue, I think it makes sense that we resolve it with styling.

jwalgran:
Is there anything blocking us from merging these changes?

designmatty:
I don't believe so.

Issue #1139, opened 2020-10-16
Create a management command to count API requests, create/update ApiBlock records, and send notifications 


jwalgran:
## Overview

Create a management command to:
- Count API requests
- Create/update `ApiBlock` records
- Send notification emails about changes to `ApiBlock` records

### Describe the solution you'd like

Use unit tests to verify the multiple logic branches

The OAR moderation team will create the text for the email notifications.

**NOTE**: Make sure to use transactions where appropriate so that running this command multiple times concurrently will not result in an invalid state.

- Add `API_FREE_REQUEST_LIMIT` to `settings.py` with a value of `50`.
- Add `NOTIFICATION_EMAIL_TO` to `settings.py`. Read the value from the `NOTIFICATION_EMAIL_TO` environment variable.
- Add a placeholder value of `notification@example.com` for the `NOTIFICATION_EMAIL_TO` environment variable in `docker-compose.yml`
- Create a `check_api_limits` management command with no arguments
- Put the logic of the command in a testable `check_api_limits(at_datetime)` function
  - Management command calls this function passing UTC now as `at_datetime`
- Add a `ContributorNotifications` model
  - Non-null OneToOneField to Contributor
  - `api_limit_warning_sent_on` - nullable datetime
  - `api_limit_exceeded_sent_on` - nullable datetime
  - `api_grace_limit_exceeded_sent_on` - nullable datetime
- `check_api_limits` logic:
  - Count the `RequestLog` records with a status code between 200 and 299 created during the month `at_datetime` is within grouped by `Contributor`
  - For each `Contributor` that made requests`
    - Get or create a `ContributorNotifications` instance.
    - Get the request limit by checking if there is an `ApiLimit` record for the `Contributor`
      - If there is no limit, use the `API_FREE_REQUEST_LIMIT` from settings
    - The the limit is zero, do nothing. Zero represents unlimited usage.
    - If the count is over 90% of the limit but less than the limit, and `ContributorNotifications.api_limit_warning_sent_on` is null or in a previous month.
      - Send a user-facing notification email to the `Contributor.admin`
        - If the email is sent successfully, update  `ContributorNotifications.api_limit_warning_sent_on` to the current time 
      - Send an admin-facing notification email to the `NOTIFICATION_EMAIL_TO` address
    - If the count is over the limit
[...]

PR #1153, opened 2020-10-28
Add API limits management command

TaiWilkin:
## Overview

Creates a management command to:

* Count API requests
* Create/update ApiBlock records
* Send notification emails about changes to ApiBlock records

Connects #1139 

## Demo

Screen Shot 2020-10-28 at 9 59 00 AM
Screen Shot 2020-10-28 at 9 58 49 AM

## Testing Instructions

* Checkout this branch
* Run `./scripts/update`
* Set a contributor's ApiLimit.monthly_limit to 0 using the admin dashboard.
* Run `./scripts/manage check_api_limits`. Nothing should happen.
* Set the contributor's ApiLimit.monthly_limit to 10.
* Use `./scripts/manage shell_plus` to create a RequestLog for the contributor, setting the date within the current month the response code in the 200s.
* Run `./scripts/manage check_api_limits`. Nothing should happen.
* Add eight more RequestLogs for the contributor (as above), for a total of nine.
* Run `./scripts/manage check_api_limits`. You should see an email printed to the console. Check the ContributorNotifications for the user; the api_limit_warning_sent_on should equal the current date.
* Run `./scripts/manage check_api_limits` again. Nothing should happen.
* Add one more RequestLog for the contributor. 
* Run `./scripts/manage check_api_limits`. You should see an email printed to the console. Check the ContributorNotifications for the user; the api_limit_exceeded_sent_on should equal the current date. 
* There should be an active ApiBlock set for the contributor with an end date set to the end of the current month. 
* Run `./scripts/manage check_api_limits`. Nothing should happen.
* Set the grace limit for the ApiBlock to 12 and set active to False. 
* Run `./scripts/manage check_api_limits`. Nothing should happen.
* Create two more RequestLogs for the contributor.
* Run `./scripts/manage check_api_limits`. Nothing should happen. You should see an email printed to the console. Check the ContributorNotifications for the user; the api_grace_limit_exceeded_sent_on should equal the current date. The ApiBlock should be active again.
* Run `./scripts/manage check_api_limits`. Nothing should happen.
* Run `./scripts/test`. All tests should pass.

## Checklist

- [ ] CI passes after rebase


jwalgran:
Looking at this now.

jwalgran:
Curious about this co-authorship with `vagrant`

Screen Shot 2020-10-28 at 11 15 06 AM

Did this happen, perhaps, by running git commands inside the VM?

src/django/api/templates/mail/api_limit_admin_subject.txt
```diff
@@ -0,0 +1 @@
+Open Apparel Registry contributor API limit notice
```
jwalgran, line 1: Consider changing this admin-facing email title to include the contributor name so that mail clients don't "helpfully" stack up notifications for different contributors.

One possibility
```
OAR API limit exceeded - {{contributor_name}}
```

TaiWilkin:
> Curious about this co-authorship with `vagrant`
> Did this happen, perhaps, by running git commands inside the VM?

Yes, I think so. I should be able to correct it by editing the commit when I rebase. (Learned something new - I didn't know that signing as `vagrant` was possible!)

src/django/api/templates/mail/api_limit_body.html
```diff
@@ -0,0 +1,26 @@
...
+        <p>
+            Hi,
+        </p>
+        <p>
+            You're receiving this email because you have used {{ count }} of
+            your {{ limit }} requests for this month.
+        </p>
+        {% if grace_limit %}
+        <p>
+            The grace limit previously set was {{ grace_limit }}.
```
jwalgran, line 18: While testing I realized that we are being more detailed about our limit calculations than we absolutely have to be.

Consider this simplified version that excludes the current count
```
        {% if grace_limit %}
        <p>
            You're receiving this email because you have exceeded
            your limit of {{ grace_limit }} OAR API requests for this month.
        </p>
        {% else %}
        <p>
            You're receiving this email because you have exceeded
            your limit of {{ limit }} OAR API requests for this month.
        </p>
        {% endif %}
```

src/django/api/templates/mail/api_limit_body.txt
```diff
@@ -0,0 +1,14 @@
+{% block content %}
+Hi,
+
+You're receiving this email because you have used
+{{ count }} of your {{ limit }} requests for this month.
+
+{% if grace_limit %}
+    The grace limit previously set was {{ grace_limit }}.
+{% endif %}
```
jwalgran, line 9: While testing I realized that we are being more detailed about our limit calculations than we absolutely have to be. 

Consider this simplified version that excludes the current count

```
{% if grace_limit %}
You're receiving this email because you have exceeded your 
limit of {{ grace_limit }} OAR API requests for this month.
{% else %}
You're receiving this email because you have exceeded your 
limit of {{ limit }} OAR API requests for this month.
{% endif %}
```

jwalgran (commented):
Nice, clear implementation. I was able to successfully exercise this using real API calls from the Swagger page and see both notifications being sent when they should.

I added a question and made some relatively small suggestions

TaiWilkin:
@jwalgran Sincere apologies for way too many fixups! I kept thinking I was done and realizing there were additional changes I should have made. 
The functionality inside the loop has been moved to a separate, `@transaction.atomic`-decorated function. The email templates have been updated and simplified. The ApiBlock is being set as initially active (good catch!). Initially I was creating the at_datetime with a naive datetime object, and adding the timezone afterwards - once I had fixed that, I no longer needed to do the extra work of updating it ex post facto, but I missed removing that instance. It's been cleaned up now.

jwalgran:
Looking at this now.

jwalgran:
I made an incomplete suggestion when I proposed simplifying the emails. The current implementation uses the same email templates for the 90% warning and the limit exceeded notifications. As a result, in my manual testing, when my test user reached 90%, the email sent said "you have exceeded your limit"

The options for handling this that I can think of are:
- Restore `count` to the context and add logic to the templates.
- Add separate templates for "warning" and "limit exceeded" emails.

I lean slightly toward using different templates. In general, "dumb" templates are easier to maintain.

jwalgran (commented):
The changes look good and I successfully tested the creation of ApiBlocks and limit and grace_limit emails. 

The only remaining issue that I can see is the one that I commented on previously. We need to distinguish between limit warning and limit exceeded emails.

TaiWilkin:
@jwalgran I updated to a second set of templates. The text in them is, "You're receiving this email because you have used 90% of your limit of {{ limit }} OAR API requests for this month." and the subject says "Open Apparel Registry API limit warning." Does this text seem clear for users?

jwalgran (approved):
Confirmed the new templates. We will likely get updated text from the OAR team so the existing text is a good starting point. Thanks.

Issue #1149, opened 2020-10-21
Allow setting the contributor on a new ApiLimit via the Django admin


jwalgran:
## Overview

Allow setting the contributor on a new ApiLimit via the Django admin

### Is your feature request related to a problem? Please describe.

We can't create new `ApiLimit`s via the admin because the contributor field is read only. We want `contributor` to be read only for editing, but not on creation.

### Describe the solution you'd like

- Instead of declaring a static list of readonly fields, override a method that does it conditionally

```python
def get_readonly_fields(self, request, obj=None):
    if obj:
        return ["contributor",]
    else:
        return []
```

PR #1159, opened 2020-10-30
Allow setting the contributor on a new ApiLimit

TaiWilkin:
## Overview

We couldn't create new ApiLimits via the admin because the contributor
field was read only. The contributor is now read only for editing, and
not for creation.

Connects #1149 

## Demo

Screen Shot 2020-10-30 at 8 34 06 AM
Screen Shot 2020-10-30 at 8 34 14 AM

## Testing Instructions

* Run `./scripts/server`
* Go to [:8081/admin/api/apilimit/](http://localhost:8081/admin/api/apilimit/)
* Create a new ApiLimit. You should be able to select a contributor. Save.
* Go to the ApiLimit you created. You should not be able to edit the contributor. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
-- Add a dissociate API [#1156](https://github.com/open-apparel-registry/open-apparel-registry/pull/1156)
+-   Add a dissociate API [#1156](https://github.com/open-apparel-registry/open-apparel-registry/pull/1156)
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


jwalgran:
Looking at this now.

jwalgran (approved):
Confirmed.

As I was testing I realized that it will be easier for admins to manage existing limits if we show the name of the contributor. Consider this change to `ApiLimit.__str__`:

```python
    def __str__(self):
        return 'ApiLimit {} - {} ({}) - limit {}'.format(
            self.id, self.contributor.name, self.contributor.id,
            self.monthly_limit)
```


Screen Shot 2020-10-30 at 8 58 54 AM

Issue #1143, opened 2020-10-20
Add an admin-only ApiBlock REST API 


jwalgran:
## Overview

Add an admin-only ApiBlock REST API 

### Describe the solution you'd like

- Add an admin-only API for `ApiBlock` records (does not appear in the swagger docs at /api/docs/) by adding an `ApiBlockViewSet`. 
  - We need list, retrieve, and update actions.
- In the update handler
  - Return 400 if the caller attempts to update `grace_limit` to a value that is not greater than `limit` and `actual`.
  - If `grace_limit` is being set or increased, set `active` to `false`.

PR #1161, opened 2020-11-02
Add an admin-only ApiBlock REST API

TaiWilkin:
## Overview

Adds an admin-only API for ApiBlock records, offering list, retrieve,
and update actions.

In the update handler, we return a 400 if the caller attempts to update
the `grace_limit` to a value that is not greater than `limit` and
`actual`.

We set `active` to false when the `grace_limit` is updated.

Connects #1143

## Demo

Screen Shot 2020-11-02 at 9 33 46 AM

Screen Shot 2020-11-02 at 9 34 06 AM

Screen Shot 2020-11-02 at 9 34 14 AM

## Testing Instructions

* Run `./scripts/server`
* Add an ApiLimit to a contributor, then make enough requests that an ApiBlock is created. 
* Login as 'c1@example.com'. Navigate to [:8081/api/api-blocks/](http://localhost:8081/api/api-blocks/).
* You should see a list of ApiBlocks. 
* Navigate to the ApiBlock you created at [:8081/api/api-blocks/:id](http://localhost:8081/api/api-blocks/1) (place address with appropriate id).
* You should see the ApiBlock details and an editing form.
* Use the form to edit the ApiBlock to increase the grace_limit above the limit and actual counts. 
* Your ApiBlock should be updated. The ApiBlock should now be inactive. 
* Update the ApiBlock to have a grace_limit that is lower than or equal to the limit and actual. 
* You should see a 400 error. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 -   Add a dissociate API [#1156](https://github.com/open-apparel-registry/open-apparel-registry/pull/1156)
+-   Add an admin-only ApiBlock API [#1161](https://github.com/open-apparel-registry/open-apparel-registry/pull/1161)
 
 ### Changed
 
--- a/src/django/api/limits.py
+++ b/src/django/api/limits.py
@@ -61,6 +61,7 @@ def check_contributor_api_limit(at_datetime, c):
             grace_limit = apiBlock.grace_limit
             if request_count > grace_limit:
                 with transaction.atomic():
+                    apiBlock.actual = request_count
                     apiBlock.active = True
                     apiBlock.save()
                 send_api_notice(contributor, limit, grace_limit)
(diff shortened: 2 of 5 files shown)


jwalgran:
Looking at this now.

jwalgran (approved):
Confirmed. Nice use of DRF.

The functionality covered by this PR appears to be correct, but while I was testing I spotted something that I missed in a previous PR.

I think it would be helpful, and appear more correct when reviewing the `ApiBlock`s, if when a contributor exceeds their `grace_limit` we also update `actual` to the new amount.

```diff
diff --git a/src/django/api/limits.py b/src/django/api/limits.py
index 1720c77..bcbdf67 100644
--- a/src/django/api/limits.py
+++ b/src/django/api/limits.py
@@ -61,6 +61,7 @@ def check_contributor_api_limit(at_datetime, c):
             grace_limit = apiBlock.grace_limit
             if request_count > grace_limit:
                 with transaction.atomic():
+                    apiBlock.actual = request_count
                     apiBlock.active = True
                     apiBlock.save()
                 send_api_notice(contributor, limit, grace_limit)

```

Issue #1140, opened 2020-10-16
Add RequestMeter middleware


jwalgran:
## Overview

Add `RequestMeter` middleware that returns 402 if a request was made with a token and there is an active, unexpired `ApiBlock` record associated with the contributor.  

### Describe the solution you'd like

- Create a `RequestMeter` middleware class in src/django/api/middleware.py
- Add the `RequestMeter` middleware to `settings.py`
  - Ensure that it is listed after `RequestLog`, so that any 402 is written to the request log
- In the `RequestMeter.__call__` method
  - Get the contributor for the authenticated user
  - Check if there is an existing, ApiBlock for the Contributor where at_datetime is less than or equal to ApiBlock.until. There may be multiple blocks so sort by until descending and look at the first block. (Potentially create a [custom manager](https://docs.djangoproject.com/en/2.2/topics/db/managers/#custom-managers) and reuse the lookup built for #1139)
    - If there is no active `ApiBlock` just return self.get_response(request)
    - If there is an active block, return 402 with a JSON object 
```json
{
  "error": true,
  "message": "You have exceeded the API request limit for your contributor account" 
}
```

PR #1166, opened 2020-11-03
Add RequestMeter middleware

TaiWilkin:
## Overview

Adds RequestMeter middleware that returns 402 if a request was made
with a token and there is an active, unexpired ApiBlock record
associated with the contributor.

Explicitly excludes the '/api/docs/?format=openapi' path. When authenticating
the user, this route is called to fetch the Swagger UI, but was falling under the 
category of token-authenticated route. Excluding it allows users to authenticate 
and to then access the Swagger UI even if they are over their request limit. 

Connects #1140 

## Demo

Screen Shot 2020-11-03 at 12 22 48 PM

## Testing Instructions

* Create an ApiBlock for c2@example.com which is active until a future date.
* Log in as c2@example.com.
* Browse [:8081/api/docs/](http://localhost:8081/api/docs/) and Authenticate with Token 1d18b962d6f976b0b7e8fcf9fcc39b56cf278051
* Submit an authenticated request through the Swagger UI. You should be blocked with a 402 status.
* Navigate to [:6543/](http://localhost:6543/) and view the facilities list. The facilities should load without issue. 
* Change the block to be inactive or in the past. 
* Submit another authenticated request. You should not be blocked. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 -   Add a dissociate API [#1156](https://github.com/open-apparel-registry/open-apparel-registry/pull/1156)
 -   Add an admin-only ApiBlock API [#1161](https://github.com/open-apparel-registry/open-apparel-registry/pull/1161)
+-   Add RequestMeter middleware [#1166](https://github.com/open-apparel-registry/open-apparel-registry/pull/1166)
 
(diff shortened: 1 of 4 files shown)


jwalgran:
Looking at this now.

jwalgran (approved):
Clear, effective implementation. I successfully tested many block scenarios and verified that using the web app when logged in as a blocked user still works.

Issue #1151, opened 2020-10-26
Temporary pop-up when users visit OAR encouraging them to complete our user survey


katieashaw:
## Overview

We'll be publishing our annual survey on Thursday 29th October. Could we add a pop-up when users land on the site encouraging them to complete the survey. 

We can re-enable the temporary survey dialog created in #869 for this new survey.

### Is your feature request related to a problem? Please describe.

We want to drive traffic to our annual survey. 

### Describe the solution you'd like

A pop-up encouraging users to complete our survey. Show the popup until December 13th.

### Additional Context

Pop up text to read:

<title> Help us help you

<blurb> We prioritize developments and new features on the OAR based on user feedback.

Got something you'd like to see on the tool? Share your feedback through our survey: <Katie to share URL>

katieashaw:
I'll share the URL to direct users to once the survey is published.

katieashaw:
The survey URL is: https://openapparelregistry.typeform.com/to/ue3mV6OD

katieashaw:
@jwalgran Typeform has various embedding options available to us with the account package we have. Let me know if you'd like to take a look & I can share log-in details.

PR #1167, opened 2020-11-03
Enable temporary survey

TaiWilkin:
## Overview

Re-enables the temporary survey popup until December 13th, 2020.

Connects #1151 

## Demo

Screen Shot 2020-11-03 at 2 43 56 PM

## Testing Instructions

* Run `./scripts/server`
* Navigate to [:6543/](http://localhost:6543/). You should see the survey popup with the updated link. Close the popup. 
* Refresh the page. You should no longer see the popup. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 -   Update the development environment to use PostgreSQL 12.4 [1146](https://github.com/open-apparel-registry/open-apparel-registry/pull/1146)
 -   Allow setting the contributor on a new ApiLimit [1159](https://github.com/open-apparel-registry/open-apparel-registry/pull/1159)
+-   Reenable survey popup [#1159](https://github.com/open-apparel-registry/open-apparel-registry/pull/1167)
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


jwalgran (commented):
When writing up the issue I didn't realize that everyone who dismissed the previous survey dialog would not see the new survey dialog because they already had the key set in their local storage.

I can think of two ways to resolve this:
1. Change the name of the constant/string to `SURVEY_2_DIALOG_HAS_BEEN_DISPLAYED` or similar
2. Add `surveyStartDate`, change the local storage key to `SURVEY_DIALOG_DISMISSED_DATE`, and show the dialog if `SURVEY_DIALOG_DISMISSED_DATE` either does not exist or is a date less than `surveyStartDate`. Also remove the `SURVEY_DIALOG_HAS_BEEN_DISPLAYED` key if it exists to avoid confusion.

Since there is a good chance we will be reusing this feature again, I lean toward the more flexible option 2.

TaiWilkin:
I've added a section, outside of the component function, which looks for and removes the `SURVEY_DIALOG_HAS_BEEN_DISPLAYED` key from local storage. We will most likely want to remove this code in the future, once it's had a chance to serve its purpose. (Possibly when we next update the survey URL and start/end dates).

jwalgran (approved):
Nice fixups. I exercised the new local storage interactions and they look good.

Issue #1144, opened 2020-10-20
Add dashboard page to list ApiBlock records and allow deactivating by increasing the grace_limit and entering notes 


jwalgran:
## Overview

Add dashboard page to list ApiBlock records and allow deactivating by increasing the grace_limit and entering notes 

### Describe the solution you'd like

- Add a /dashboard/apiblocks route
- Add a `DashboardApiBlocks` component loosely modeled after `DashboardLists` in that it shows a tabular view of the `ApiBlock` records
  - Display active, not expired blocks in one table at the top, and the rest of the blocks in another table
  - Display the `ApiBlock` records in reverse chronological order by `changed_at`
  - Show active, until, limit, actual, grace_limit, and grace_reason columns
- Only the active, non expired table has clickable rows
  - When a row is clicked render an edit view where the grace_limit and grace_reason can be edited and PUT to the API.
    - Include a back button to return to the table view

PR #1170, opened 2020-11-09
Add ApiBlock Dashboard

TaiWilkin:
## Overview

Adds the ApiBlock list view and the ApiBlock update view. Users can
view and update the ApiBlocks' grace_limit and grace_reason.

The list view displays active, not expired blocks in one table at the
top, and the rest of the blocks in another table. The full list of
blocks is filterable by contributor. This filter is client-side only.

Connects #1144 

## Demo

Screen Shot 2020-11-09 at 3 17 11 PM
Screen Shot 2020-11-09 at 3 17 46 PM
Screen Shot 2020-11-09 at 3 18 36 PM

## Testing Instructions

* Run `./scripts/server`
* Ensure you have at least one active and one inactive Api Block in your database. 
* Navigate to [:6543/dashboard](http://localhost:6543/dashboard).
* Click on 'View API Blocks'
* You should be navigated to the API Block list. The top list should include only active, unexpired blocks. 
* The bottom list should include all the blocks you have in your database. 
* You should be able to filter the bottom list by contributor. 
* Click on an active block in the top list. You should be taken to the update view. 
* Enter a grace limit above the actual count and request limit. Enter a grace reason. Click submit.
* The ApiBlock should be updated. 
* Return to the list by clicking the 'back to Api Blocks' button. You should see the ApiBlock was updated here as well. 
* Go to an active block's update page. Attempt to enter a grace limit below the actual count or request limit. You should see an error. 

## Checklist

- [x] CI passes after rebase


jwalgran:

src/django/api/views.py
```diff
@@ -3222,7 +3222,8 @@ def validate_request(self, request):
 
     def list(self, request):
         self.validate_request(request)
-        response_data = ApiBlockSerializer(self.queryset, many=True).data
+        response_data = ApiBlockSerializer(
+            ApiBlock.objects.all(), many=True).data
```
jwalgran, line 3226: I think this is ok since we documented it with a comment message but it surprised me that we needed to do this. I explored plain Django queryset interactions in `shell_plus` and verified my expectations that each evaluation of a queryset would fetch new results.

```
>>> q = ApiBlock.objects.all()
>>> q
<QuerySet [<ApiBlock: ApiBlock 1 - Contributor 2 until 2020-11-30 00:00:00+00:00>]>
>>> b = ApiBlock.objects.get(id=1)
>>> b.until = '2020-11-29'
>>> b.save()
>>> q
<QuerySet [<ApiBlock: ApiBlock 1 - Contributor 2 until 2020-11-29 00:00:00+00:00>]>
```

I don't know how DRF is altering this behavior. Is there some documentation that pointed you to this solution?

jwalgran (approved):
Excellent implementation. I exercised this using real API calls and `check_api_limits` and was able to successfully go through the entire workflow.

src/app/src/components/DashboardApiBlocksTable.jsx
```diff
@@ -0,0 +1,112 @@
...
+                        <TableCell padding="dense">Actual Count</TableCell>
+                        <TableCell padding="dense">Grace Limit</TableCell>
+                        <TableCell>Grace Reason</TableCell>
+                        <TableCell padding="dense">Active</TableCell>
+                    </TableRow>
+                </TableHead>
+                <TableBody>
+                    {apiBlocks.map(block => (
+                        <TableRow
+                            key={`${block.contributor}-${moment(block.until).format('L')}`}
```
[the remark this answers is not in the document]
TaiWilkin: Yes, I can update this. At the time I initially wrote this line, we weren't sending the id from the server, but that's since been updated.

src/django/api/views.py
TaiWilkin: Only a few StackOverflow posts. [This one](https://stackoverflow.com/a/42764200) suggested that the class's queryset is set once on initialization, and then isn't refreshed. "...anything happening outside a function (ie at a module's top-level, in a class statement's body etc) is executed only once (per process) at module's first import". Switching to fetching the data directly in the `list` request then fixed the issue, so it seems to be accurate.

Issue #1185, opened 2020-12-09
Confirm / Reject: Multiple Potential Matches Linked as 1


OAR-Community:
## Overview

During confirm/reject analysis of a list, the following bug was noticed. When an entry has multiple potential matches, names and address of the multiple facilities are not broken by line items. Instead, the facilities are linked via 1 hyperlink formatted as "https://openapparel.org/facilities/[OAR ID 1],[OAR ID 2],[OAR ID 3], etc.". Additionally, the confirm and reject button shows only for the top-most potential match. If confirmed, the other matches are ignored as expected. If rejected, the button moves to the second entry. 

### Expected Behavior

When facilities have multiple potential matches, the matches should show in individual line items with a Confirm and Reject option for each potential match. 

### Actual Behavior

Multiple potential matches show as a singular entry, with hyperlink combining all OAR IDs into one Error-producing link. 

### Steps to Reproduce

* Using Google Chrome Browser. 
* Dashboard > View Contributor Lists > Any List with Potential Matches
* Example List: Better Cotton Initiative 
* Line 595

![Screen Shot 2020-12-09 at 9 25 28 AM](https://user-images.githubusercontent.com/68333623/101642878-83782700-3a01-11eb-9bbd-4f67a55bdab5.png)

### Demo

Hyperlink creation from click the above example: 
![Screen Shot 2020-12-09 at 9 33 15 AM](https://user-images.githubusercontent.com/68333623/101642943-9a1e7e00-3a01-11eb-8f9b-274bc3f6d344.png)

### Additional context

Optional. Add any other context about the problem here.

jwalgran:
This issue exists in development list 8 item 25. "Hanes Choloma"

<img width="1554" alt="Screen Shot 2020-12-11 at 9 00 05 AM" src="https://user-images.githubusercontent.com/17363/101925708-683b2200-3b8f-11eb-9366-75f113c55a88.png">

jwalgran:
Likely a regression caused by restyling the table in https://github.com/open-apparel-registry/open-apparel-registry/pull/1152

TaiWilkin:
Reopening to pursue a more permanent fix through refactoring.

PR #1189, opened 2020-12-11
Fix pending matches bug

TaiWilkin:
## Overview

During confirm/reject analysis of a list, the following bug was noticed.
When an entry had multiple potential matches, names and address of the
multiple facilities were not broken by line items. The links were
linked to an incorrect address. Additionally, the confirm and reject
button showed only for the top-most potential match.

When facilities have multiple potential matches, the matches show in
individual line items with a Confirm and Reject option for each
potential match.

There were a minor propType warning and key warning which have also
been corrected.

Connects #1185

## Demo

Screen Shot 2020-12-11 at 1 28 30 PM

## Notes

The line items aren't currently perfectly aligned. This PR fixes the functionality issue, but additional styling might be desired in the future. 

Currently, the matches are all stored in a single row, with each cell containing the section header and multiple matches' data for that value. Due to the varying height of the cells, and the cell component being used in multiple places, my attempts to fix this with styling were inefficient. I also worried that a set height to the cell content would cause overflow issues. 

My suggested long-term fix for the styling is to refactor this section to render the headers and each facility match as separate rows in the table, which would keep the line-items aligned despite the variable height. 

## Testing Instructions

* Checkout this branch.
* Navigate to [a list item with two pending matches](http://localhost:6543/lists/8?search=Hanes%20Choloma&page=1&rowsPerPage=20)
* The matches should be rendered as distinct items.
* The Facility Match Name link should lead to the correct facilities. 
* Selecting 'confirm' or 'reject' should confirm or reject the correct facility. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/App.css
+++ b/src/app/src/App.css
@@ -693,7 +693,7 @@ hr {
   .form__action {
     margin-bottom: 5px;
   }
-  
+
   .filter-sidebar-tabgroup {
     position: sticky !important;
     top: -11px;
@@ -755,8 +755,4 @@ hr {
 
 .STATUS_PARSED:not(:hover) {
   background: #fff !important;
-}
-
-.STATUS_POTENTIAL_MATCH--ACTIONS + .STATUS_POTENTIAL_MATCH--ACTIONS {
-  display: none !important;
 }
--- a/src/app/src/components/FacilityListItemsConfirmationTableRow.jsx
+++ b/src/app/src/components/FacilityListItemsConfirmationTableRow.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import { bool, func } from 'prop-types';
 import { connect } from 'react-redux';
-import { Link } from 'react-router-dom';
 import TableRow from '@material-ui/core/TableRow';
 import TableCell from '@material-ui/core/TableCell';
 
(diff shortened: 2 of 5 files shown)


src/app/src/App.css
```diff
@@ -756,7 +756,3 @@ hr {
 .STATUS_PARSED:not(:hover) {
   background: #fff !important;
 }
-
-.STATUS_POTENTIAL_MATCH--ACTIONS + .STATUS_POTENTIAL_MATCH--ACTIONS {
-  display: none !important;
```
TaiWilkin, line 761: This line explicitly hid any confirm/reject buttons after the first.

src/app/src/components/FacilityListItemsDetailedTableRowCell.jsx
```diff
@@ -71,7 +71,7 @@ export default function FacilityListItemsDetailedTableRowCell({
                 </div>
                 {
                     data.map((item, index) => (
-                        <Fragment key={hasActions ? item.id : item}>
+                        <Fragment key={item.id ? item.id : item}>
```
TaiWilkin, line 74: This line was causing the main facility to receive an object as a key, instead of an id, causing a key warning in the console.

src/app/src/components/FacilityListItemsConfirmationTableRow.jsx
```diff
@@ -154,42 +153,35 @@ function FacilityListItemsConfirmationTableRow({
...
+                    />
                 </TableCell>
                 <TableCell
                     padding="default"
                     style={listTableCellStyles.statusCellStyles}
                 >
+                    <b>Actions</b>
                     <FacilityListItemsDetailedTableRowCell
-                        title
+                        title={' '}
```
TaiWilkin, line 184: A string title is required by the FacilityListItemsDetailedTableRowCell. Not passing one causes a prop-type warning.

src/app/src/components/FacilityListItemsConfirmationTableRow.jsx
```diff
@@ -154,42 +153,35 @@ function FacilityListItemsConfirmationTableRow({
...
+                    />
                 </TableCell>
                 <TableCell
                     padding="default"
                     style={listTableCellStyles.statusCellStyles}
                 >
+                    <b>Actions</b>
                     <FacilityListItemsDetailedTableRowCell
-                        title
+                        title=" "
```
TaiWilkin, line 184: A string title is required by the FacilityListItemsDetailedTableRowCell. Not passing one causes a prop-type warning.

jwalgran:
CI caught some lint

```
/usr/local/src/src/components/FacilityListItemsConfirmationTableRow.jsx
  158:31  error  Curly braces are unnecessary here  react/jsx-curly-brace-presence
  172:31  error  Curly braces are unnecessary here  react/jsx-curly-brace-presence
  184:31  error  Curly braces are unnecessary here  react/jsx-curly-brace-presence
```

jwalgran:
Looking at this now.

jwalgran (approved):
Thanks for digging into one of the trickiest bits of code in the project and finding a relatively simple solution. I confirmed that this is working. Nice explanatory comments.

@designmatty I decided to get this partial fix to #1185 sooner rather than later because a significant feature is broken on production. I would like to follow up with a more comprehensive refactor that aligns each potential match.

@TaiWilkin could you please add a note explaining the challenge you encountered when trying to align the matches?

Issue #1186, opened 2020-12-09
Change metered pricing to count yearly rather than monthly


jwalgran:
## Overview

Change metered pricing to count yearly rather than monthly

### Is your feature request related to a problem? Please describe.

In real world usage it will be common for contributors to make a high volume of requests only a few time a year. We will likely be charging a yearly rate.

### Describe the solution you'd like

- Modify the limit checking to use year-size windows rather than month-size windows

Issue #1175, opened 2020-11-12
Add a switch to the check_api_limits the API block middleware logic


jwalgran:
## Overview

Add a waffle switch that disables the functionality in `check_api_limits` and `RequestMeterMiddleware`.

### Is your feature request related to a problem? Please describe.

We want to deploy our scheduling of `check_api_limits` but don't want it active in production until our pricing structure is complete.

### Describe the solution you'd like

* Add a `request_metering` waffle switch, following the convention of previous feature flags.
* Wrap the `check_api_limits` and `RequestMeterMiddleware` logic with a check of the `request_metering` switch.

jwalgran:
We went live with this feature in production so we do not need a feature flag.

PR #1193, opened 2020-12-28
Scheduled check_api_limits invocation with Step Functions

### LINKED ISSUE #1165: Set up scheduled invocation of check_api_limits
In #1153, we created a Django management command named `check_api_limits` that checks current API usage and blacklists folks who go over their quota. The next step is to set up a system for invoking the command once an hour. 

As part of #1145, I met with @jwalgran and @TaiWilkin. We decided to use a Step Functions state machine, triggered by a CloudWatch Events Rule, to start and monitor an ECS task. As an organization, we have had success with this approach in a few projects (e.g. [in PWD Stormdrain Marking](https://github.com/azavea/nome-state-of-the-data/pull/68)).

- [ ] Create supporting IAM resources (Step Functions service role w/ policy and CloudWatch Events service role).
- [ ] Create a Step Functions state machine that will run the App CLI ECS task.
  - [ ] Wait for it to complete and timeout after a defined interval.
  - [ ] Catch failures and trigger a Slack notification. 
- [ ] Create a CloudWatch Events Rule that triggers the state machine on a schedule.

The Slack notification is the only part of the workflow that we haven't implemented before. We could have the state machine invoke a Lambda function that triggers a Slack web hook. We could also have the state machine publish a message to an SNS topic and [add an SNS topic trigger](https://aws.amazon.com/premiumsupport/knowledge-center/sns-lambda-webhooks-chime-slack-teams/) to the Lambda function. We have used CloudWatch Alarms in the past with state machines, and could also trigger a Lambda function based on an alarm state, however we don't know how to retrieve the cause of the state machine's failure in that scenario.

External resources:

- https://cloudonaut.io/resilient-task-scheduling-with-ecs-fargate-cron-scheduled-task/
- https://docs.aws.amazon.com/AmazonECS/latest/developerguide/scheduled_tasks.html
- https://docs.aws.amazon.com/step-functions/latest/dg/concepts-error-handling.html
- https://docs.aws.amazon.com/step-functions/latest/dg/tutorial-creating-lambda-state-machine.html

rbreslow:
## Overview

The following changes were made to support the scheduled invocation of the `check_api_limits` management command:

- Create supporting IAM resources (Step Functions service role w/ policy and CloudWatch Events service role).
- Create a Step Functions state machine that will run the App CLI ECS task.
- Catch and surface failures to Rollbar via a Lambda function.
- Create a CloudWatch Events Rule that triggers the state machine with a configurable schedule.

Resolves #1165 

## Testing Instructions

- Select **Step Functions** → **State machines** → **stateMachineStagingAppCLI** .
- See that the CloudWatch Events Rule is invoking the task every hour:
   image
- Select **Start Execution**.
  - **Input**: 
  ```json
  {
    "commands": [
      "check_api_limits_asdfghjkl"
    ]
  }
  ```
- Wait for a Rollbar message to appear in Slack.
   image

In order to view specific log output, you'll need to find the task in the ECS console or find the CloudWatch log stream. Although, if `check_api_limits` is successful, sometimes [there isn't any](https://github.com/open-apparel-registry/open-apparel-registry/pull/1153#issue-511568953) log output.

## Checklist

- [x] CI passes after rebase


deployment/terraform/cdn.tf
```diff
@@ -80,7 +80,7 @@ resource "aws_cloudfront_distribution" "cdn" {
     viewer_protocol_policy = "redirect-to-https"
     min_ttl                = 0
     default_ttl            = 3600
-    max_ttl                = 31536000 # 1 year. Same as TILE_CACHE_MAX_AGE_IN_SECONDS in src/django/oar/settings.py
+    max_ttl                = 31536000            # 1 year. Same as TILE_CACHE_MAX_AGE_IN_SECONDS in src/django/oar/settings.py
```
rbreslow, line 83: `terraform fmt`.

deployment/terraform/container_service.tf
```diff
@@ -59,7 +59,7 @@ resource "aws_lb" "app" {
   # This is a "hidden" dependency that Terraform cannot automatically infer, so
   # it must be declared explicitly.
   depends_on = [
-    "aws_s3_bucket_policy.alb_access_logging"
+    "aws_s3_bucket_policy.alb_access_logging",
```
rbreslow, line 62: `terraform fmt`.

deployment/terraform/iam.tf
```diff
@@ -179,9 +179,9 @@ resource "aws_iam_role_policy_attachment" "spot_fleet_policy" {
 }
 
 #
-# Lambda resources
+# Lambda IAM resources
 #
-data "aws_iam_policy_document" "alert_batch_failures_assume_role" {
+data "aws_iam_policy_document" "lambda_assume_role" {
```
rbreslow, line 184: Generalize the name of this policy document so that it can be used by the IAM role for the second Lambda function.

deployment/terraform/config.tf
```diff
@@ -1,6 +1,6 @@
 provider "aws" {
   region  = "${var.aws_region}"
-  version = "~> 1.56.0"
+  version = "~> 2.70.0"
```
rbreslow, line 3: This resolves a race condition I encountered with the creation of resources for Step Functions: https://github.com/hashicorp/terraform-provider-aws/issues/7893.

deployment/terraform/sfn.tf
```diff
@@ -0,0 +1,47 @@
...
+    lambda_function_arn     = "${aws_lambda_function.alert_sfn_failures.arn}"
+  }
+}
+
+resource "aws_sfn_state_machine" "app_cli" {
+  name     = "stateMachine${var.environment}AppCLI"
+  role_arn = "${aws_iam_role.step_functions_service_role.arn}"
+
+  definition = "${data.template_file.app_cli_state_machine.rendered}"
+}
```
rbreslow, line 23: I designed this to be something that we could reuse, independent of the rule that invokes `check_api_limits`. We can invoke this State Machine with whatever command we want. This means we could easily add more scheduled management commands.

deployment/terraform/step-functions/app_cli.json
```diff
@@ -0,0 +1,45 @@
...
+                    "Next": "Notify scheduled task failure"
+                }
+            ],
+            "End": true
+        },
+        "Notify scheduled task failure": {
+            "Type": "Task",
+            "Resource": "${lambda_function_arn}",
+            "End": true
+        }
```
rbreslow, line 43: The [AWS docs](https://docs.aws.amazon.com/step-functions/latest/dg/connect-lambda.html) have some reference at the bottom on what I'm doing here.

> To invoke a Lambda function, you can also call the resource Amazon Resource Name (ARN) directly in the "Resource" string.

> The state input is passed to Lambda as the payload.

deployment/terraform/variables.tf
```diff
@@ -245,6 +245,14 @@ variable "batch_default_ce_instance_types" {
   ]
 }
 
+variable "app_cli_state_machine_timeout_seconds" {
+  default = "120"
+}
```
rbreslow, line 250: Arbitrary 🤷.

docker-compose.ci.yml
```diff
@@ -4,7 +4,7 @@ services:
     image: "openapparelregistry:${GIT_COMMIT:-latest}"
 
   terraform:
-    image: quay.io/azavea/terraform:0.11.11
+    image: quay.io/azavea/terraform:0.11.14
```
rbreslow, line 7: When I was upgrading the AWS provider version, I thought I might as well do this too since it's a patch bump. We'll be in a good place to do a Terraform 0.12 upgrade now.

scripts/infra
```diff
@@ -41,6 +41,7 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
 
                 # Build Lambda function archives
                 make -sC lambda-functions/alert_batch_failures
+                make -sC lambda-functions/alert_sfn_failures
```
rbreslow, line 44: I see a world where we combine or further abstract these functions, [like we do in Raster Foundry](https://github.com/azavea/raster-foundry-deployment/tree/develop/core/terraform/lambda-functions/alert_failures). I decided not to pursue this since there was already a challenge in parsing the ECS output that's specific to Step Functions. 

Imagine if we had a Python module that's job was just to handle AWS errors and surface them to Slack. I think that could be useful.

deployment/terraform/iam.tf
```diff
@@ -196,10 +196,157 @@ data "aws_iam_policy_document" "alert_batch_failures_assume_role" {
...
+
+    actions = [
+      "events:PutTargets",
+      "events:PutRule",
+      "events:DescribeRule",
+    ]
+
+    resources = [
+      "arn:aws:events:${var.aws_region}:${data.aws_caller_identity.current.account_id}:rule/StepFunctionsGetEventsForECSTaskRule",
+    ]
```
rbreslow, line 292: I'm still not entirely sure what this is for but it is referenced in the docs from the above comment and things will not work without it.

deployment/terraform/lambda-functions/alert_sfn_failures/src/alert_sfn_failures.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Extract "Task failed ELB health checks" from "Task failed ELB health checks in (...)"
+    if re.match("^(.+) in \((.+)$", reason):
+        reason = reason.partition(" in")[0]
+    # Extract "OutOfMemoryError" from "OutOfMemoryError: ..."
+    elif re.match("^(.+): (.+)$", reason):
+        reason = reason.partition(":")[0]
+    # Extract "Error" from "Error (...)":
+    elif re.match("^(.+) \((.+)\)$", reason):
+        reason = reason.partition(" (")[0]
```
rbreslow, line 38: These were copied from my work [here](https://github.com/azavea/raster-foundry-deployment/blob/e5b2420ae469429f2ec96ec669bd2cf27f8bb16b/core/terraform/lambda-functions/alert_failures/src/alert_failures.py#L90-L98).

deployment/terraform/config.tf
ColeKettler: Good catch, glad to see an upgrade!

deployment/terraform/iam.tf
```diff
@@ -196,10 +196,157 @@ data "aws_iam_policy_document" "alert_batch_failures_assume_role" {
...
+  }
+
+  statement {
+    effect = "Allow"
+
+    actions = ["ecs:RunTask"]
+
+    # Allow ecs:RunTask for all current and future versions of the task
+    # definition.
+    resources = ["${replace(aws_ecs_task_definition.app_cli.arn, "/:\\d+$/", ":*")}"]
```
ColeKettler, line 257: Good approach!

deployment/terraform/iam.tf
ColeKettler: Big shrug from me too. Based on my shallow experience with Step Functions, it touches on a pretty wide swath of AWS products and privileges, so I'm inclined to just go with whatever AWS suggests.

deployment/terraform/sfn.tf
```diff
@@ -0,0 +1,47 @@
+#
+# Step Function Resources
+#
+data "template_file" "app_cli_state_machine" {
+  template = "${file("step-functions/app_cli.json")}"
+
+  vars = {
+    ecs_cluster_arn         = "${aws_ecs_cluster.app.arn}"
+    security_group_id       = "${aws_security_group.app.id}"
+    subnet_ids              = "${jsonencode(module.vpc.private_subnet_ids)}"
```
ColeKettler, line 10: Ooooo nice approach.

scripts/infra
ColeKettler: Agreed, that sounds super useful. I'm kinda surprised that AWS doesn't have some SNS-adjacent feature or product to clean up and send valid notifications out to common chat platforms already. Looks like Lambda is still the way to go!

deployment/terraform/sfn.tf
ColeKettler: Good idea! This feels like the right abstraction.

scripts/infra
ColeKettler: Also, lemme just say, I love Make.

deployment/terraform/lambda-functions/alert_sfn_failures/src/alert_sfn_failures.py
```diff
@@ -0,0 +1,53 @@
...
+
+ECS_TASK_URL = "https://console.aws.amazon.com/ecs/home?region={region}#/clusters/{cluster}/tasks/{task}/details"
+
+environment = os.getenv('ENVIRONMENT', 'Staging')
+rollbar_token = os.getenv('ROLLBAR_SERVER_SIDE_ACCESS_TOKEN')
+rollbar.init(rollbar_token, environment)
+
+@rollbar.lambda_function
+def handler(event, context):
+    event = json.loads(event["Cause"])
```
ColeKettler, line 14: It looks like there some cases (step function run [here](https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/executions/details/arn:aws:states:eu-west-1:249322298638:execution:stateMachineStagingAppCLI:a99e8ea3-008c-df2b-ea30-f58ba1d283c8_a5839bc8-b698-7163-2b24-721714e0dcec)) where `event["Cause"]` is an empty string, like when `"Error": "States.Timeout"`. This has been propagating to Rollbar, but as an uncaught exception (item [here](https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/630/)) rather than an intentionally submitted item.

My thoughts are:

1. The `check_api_limits` command seems to be timing out, not sure if that's something that just happens sometimes or if we should increase the timeout to see if that improves the success rate.
2. We should handle cases where the cause is an empty string, possibly by submitting the `Error` instead.

What are your thoughts here?

ColeKettler (commented):
Looking good so far! Coming back to you with comments!

jwalgran:
Thanks for your work on this.

Regarding timeouts, single-digit minute execution times are not unexpected. The `check_api_limits` command loops over a list of contributors and runs a reporting query for each one. Bumping our timeout to 600 seconds should be sufficient.

Regarding the merging of this change when it is complete, at least one of these two issues needs to be completed before we start running the scheduled task
- Add a switch to the check_api_limits the API block middleware logic #1175
- Change metered pricing to count yearly rather than monthly #1186

deployment/terraform/lambda-functions/alert_sfn_failures/src/alert_sfn_failures.py
rbreslow: Nice catch, I missed that. I added something that catches the `JSONDecodeError` and surfaces a generic message.

image

I wanted to add more color here, like what task timed out, but as you can see here there really isn't much to go off of besides the name of the Step Functions error and the Lambda function itself.

https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/635/occurrences/147678787027/

Any ideas on how we can push more detail to the Lambda function? Maybe it's not necessary? I adjusted the timeout to 600 seconds as @jwalgran suggested.

deployment/terraform/iam.tf
hectcastro: This is a piece of CloudWatch Events rule [glue](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#rules:name=StepFunctionsGetEventsForECSTaskRule) that helps SFN react to ECS task state changes. 

```json
{
  "source": [
    "aws.ecs"
  ],
  "detail-type": [
    "ECS Task State Change"
  ],
  "detail": {
    "lastStatus": [
      "STOPPED"
    ],
    "desiredStatus": [
      "STOPPED"
    ],
    "startedBy": [
      "AWS Step Functions"
    ]
  }
}
```

hectcastro (approved):
Changes look good and include some nice incremental improvements. 👍🏼 

As a side note, it is a shame that so much IAM resource wrangling is necessary to make this work.

deployment/terraform/lambda-functions/alert_sfn_failures/src/alert_sfn_failures.py
ColeKettler: Some way to correlate the reported error back to the step function execution would be nice, but I like that Lambda isn't aware of the context that it's run in.

I can see an approach where we pass an envelope of arbitrary data to the Lambda reporting function to pass along to Rollbar, but I don't think that's a high priority at this time. Correlating recent failures and glancing at timestamps should be enough given the frequency that this is run at.

ColeKettler (approved):
Recent round of changes look good. Excellent and thorough work on this!

PR #1197, opened 2021-01-05
Add ApiBlock.yearly_limit to prepare for replacing monthly_limit

jwalgran:
## Overview

Most contributors will use the API in bursts during over the course of the year so we have opted to switch to counting API calls per year.

Connects #1186

## Notes

To rename the non-null `monthly_limit` to `yearly_limit` we will do a three-stage deployment. The first commit adds a nullable `yearly_limit` field and will be deployed to production by itself as a hotfix.

The bulk of the change is in the second command and the second migration makes the old `monthly_limit` field nullable in preparation for its removal. To maintain backward compatibility we will only remove the field after this PR has been deployed code that references it has been removed from all deployments.

## Testing Instructions

* Verify that the tests pass and visually verify the code changes.

## Checklist

- [ ] CI passes after rebase


TaiWilkin (approved):
Looks good! All tests are passing and the code reads well.

jwalgran:
Thanks for the review.

PR #1198, opened 2021-01-05
Remove ApiLimit monthly_limit and make yearly_limit non-null 

jwalgran:
## Overview

This is the last of a three-step process to rename the non-null monthly_limit field to yearly_limit.

We don't need to worry about existing values with null `yearly_limit` fields because no records have been created in production.

Connects #1186

## Notes

This should be deployed only after #1197 has been deployed.

## Testing Instructions

* Verify the migration can be applied and the tests pass.

## Checklist

- [x] CI passes after rebase


TaiWilkin (approved):
Looks good! The migrations ran and the tests passed.

jwalgran:
Thanks for the review.

Issue #1155, opened 2020-10-28
It's becoming difficult to navigate in Django with so many single facility submissions


katieashaw:
## Overview

As users connect with our push pull API and submit multiple single facility queries / contributions, the volume of entries in the Django admin dashboard is growing exponentially. This makes it difficult to quickly search and make changes following user requests.

### Expected Behavior

A more efficient way to search through different sections of Django. 

### Actual Behavior

There are numerous pages in each section of Django, e.g. 54,773 pages of Sources, 537 facility lists (6 pages)

### Additional context

The most frequent user requests we receive which are becoming more cumbersome as the database grows and users make use of our API are:

- Deactivating lists, when users forget to select "replace a list" when uploading an updated list
- Moving facility lists from one Contributor account to another, e.g. when moving data uploaded by the OAR team over to a user, or when one person leaves an organisation and a colleague takes over managing data in the OAR

katieashaw:
An example of this in practice: I've just spent over 20 mins searching for a Public List in the Sources menu in Django and still not been able to find it, due to the volume of single item sources that we now have.

Is there a way we could separate out list sources from single item sources that have been contributed via API?

jwalgran:
Enabling `list_filter` on the admin paged may get us what we need quickly and easily.

https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_filter

PR #1207, opened 2021-01-08
Admin updates to allow filtering the source admin list by type and contributor and search users

jwalgran:
## Overview

Moving ownership of a public list to a new contributor account after they have registered is a common administrative activity. Now that the application is receiving thousands of API submissions, it is impractical to browse the Source list to find a specific list. Adding filters allows for more quickly finding a specific record.

Attempting to search for a user was raising an exception because by default it was attempting to search for `first_name` and `last_name` fields that we do not have on our User model. Setting `search_fields` resolves this.

We don't use `username` on our User model so setting `list_display` suppresses this always empty field in the list view.

Connects #935
Connects #1155

## Demo

Screen Shot 2021-01-08 at 2 33 25 PM

Screen Shot 2021-01-08 at 2 33 00 PM

## Testing Instructions

* Log in as `c1@example.com`
* Browse http://localhost:8081/admin/api/source and verify that you can filter by type and contributor
* Browse http://localhost:8081/admin/api/user and verify that the search field at the top works

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -666,7 +666,9 @@ Directly connect health check to GazetteerCache and remove the use of threading 
 
 -   Initial release.
 
-[unreleased]: https://github.com/open-apparel-registry/open-apparel-registry/compare/2.37.1...HEAD
+[unreleased]: https://github.com/open-apparel-registry/open-apparel-registry/compare/2.38.1...HEAD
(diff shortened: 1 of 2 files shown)


TaiWilkin (approved):
This looks good and is working well!

jwalgran:
Thanks for the review.

Issue #1191, opened 2020-12-15
Add page to manage cookie preferences


dboyer:
## Overview

The user should have the option to access a cookies preferences page where they can choose to change the opt in or out preference that they set when first visiting OAR.

### Describe the solution you'd like

* If `GA_TRACKING` is in local storage add a section to the https://openapparel.org/profile/{id} page that shows the same cookie message that is shown in the initial popup

<img width="624" alt="Screen Shot 2021-01-19 at 8 41 40 PM" src="https://user-images.githubusercontent.com/17363/105124213-ea740b80-5a96-11eb-9e42-51fba6b085d6.png">

* Show either an accept or reject button based on the value of GA_TRACKING (reference `app/src/util/util.ga.js`)
* If a button is clicked, call the same helper functions as the popup to change the preference. Ensure that the profile page rerenders to show the updated state.

katieashaw:
Could we prioritize this when we're back in January please.

PR #1213, opened 2021-01-21
Add page to manage cookie preferences

TaiWilkin:
## Overview

Gives the user the option to access a cookies preferences page where
they can choose to change the opt in or out preference that they set
when first visiting OAR.

Connects #1191 

## Demo

Screen Shot 2021-01-21 at 9 05 07 AM

Screen Shot 2021-01-21 at 9 21 46 AM

## Testing Instructions

* Checkout this branch and run `./scripts/server` 
* Login and navigate to the user profile
* Open the local storage and view the current settings. Click the approve/reject button.
* The settings in local storage should be updated, and the button should toggle approve/reject. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/components/UserCookiePreferences.jsx
+++ b/src/app/src/components/UserCookiePreferences.jsx
@@ -1,0 +1,83 @@
+import React, { Component } from 'react';
+
+import Divider from '@material-ui/core/Divider';
+
+import COLOURS from '../util/COLOURS';
+import Button from './Button';
+import CookiePreferencesText from './CookiePreferencesText';
+
+import {
+    userHasAcceptedGATracking,
+    acceptGATrackingAndStartTracking,
+    rejectGATracking,
+    clearGATrackingDecision,
+} from '../util/util.ga';
+
+const componentStyles = Object.freeze({
+    header: Object.freeze({
+        padding: '0 1rem',
+    }),
+    content: Object.freeze({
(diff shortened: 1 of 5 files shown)


jwalgran:
Looking at this now.

src/app/src/components/UserCookiePreferences.jsx
```diff
@@ -0,0 +1,94 @@
...
+                        agree to allow us to place cookies and share information with
+                        Google Analytics. For more information, please visit our{' '}
+                        <a
+                            href="https://info.openapparel.org/tos/"
+                            target="_blank"
+                            rel="noopener noreferrer"
+                        >
+                            Terms and Conditions of Use and Privacy Policy.
+                        </a>
+                    </div>
```
jwalgran, line 72: Consider extracting this tex into a component that can be used both here and in `GDPRNotification.jsx`. If we need to update this language in the future it will eliminate the possibility of accidentally updating it in one place and not another.

jwalgran (approved):
Straightforward, working implementation. I suggested DRYing it up a bit.

src/app/src/components/UserCookiePreferences.jsx
TaiWilkin: Good call! Thank you for the review!

Issue #1205, opened 2021-01-07
Script the generation of 100,000 and 200,000 facility data sets for load testing


jwalgran:
## Overview

Our 5 year goal is to reach 200,000 facilities. We need to load test at that capacity to see what is too slow or breaks completely. The first step is generating a synthetic data set of the appropriate size.

### Describe the solution you'd like

- Generate large test fixtures using the existing tooling and see if they are loadable in development.

PR #1214, opened 2021-01-26
Add a management command to make CSVs from a CSV download

jwalgran:
## Overview

Add a tool that allows us to generate a collection of CSV files suitable for uploading
by processing a CSV downloaded from the running application with the option of generating more output rows than exist in the input. Parameterize `makefixtures` and `processfixtures` so that they can be run with larger input sets.

Connects #1205

## Testing Instructions

* Run `./scripts/server`
* Download and unzip `{secure file storage}/Open\ Apparel\ Registry/data-dumps-for-testing/facilities-2021-01-25.csv.zip` into `src/django/facilities-2021-01-25.csv`
* `mkdir src/django/out`
* Run makecsvs 
```
vagrant@vagrant:/vagrant$ ./scripts/manage makecsvs -f 100 -r 100000 -o out facilities-2021-01-25.csv
```
* Copy the new csvs into place
```
vagrant@vagrant:/vagrant$ cp src/django/out/*.csv src/django/api/management/commands/facility_lists/geocoded/
```
* Verify `src/django/out` contains 100 files. Inspect one and make sure the structure looks good
* Run makefixtures 
```
vagrant@vagrant:/vagrant$ ./scripts/makefixtures -u 1000 -c 999 -l 100 -s 100
```
* Run `./scripts/resetdb` and `./scripts/manage processfixtures --startid 16 --endid 101`. The `processfixtures` process will take a while because it needs to run 84 addition files through the matcher. (the first 16 are handled by `resetdb`)
* Verify that it runs without failure (parse and match errors that do not crash the process are fine)
* Browse http://localhost:6543/ and verify there are about 90k facilities (There are some parse errors the number of matches created will be different each time, so there is not an exact target number of facilities)
* After testing clean up your working copy
```
rm src/django/api/management/commands/facility_lists/geocoded/*.csv
git checkout src/django/api/management/commands/facility_lists/geocoded
```  
## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
+- Add a management command to make CSVs from a CSV download [#1214](https://github.com/open-apparel-registry/open-apparel-registry/pull/1214)
+
 ### Changed
 
 ### Deprecated
--- a/scripts/makefixtures
+++ b/scripts/makefixtures
@@ -17,6 +17,6 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     if [ "${1:-}" = "--help" ]; then
         usage
     else
-        docker-compose exec django /bin/bash -c "pip install -q -r requirements.dev.txt && ./manage.py makefixtures"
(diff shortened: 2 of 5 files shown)


TaiWilkin:
Taking a look now!

TaiWilkin (approved):
This works well and looks good! Successfully ran and got 90262 facilities.

jwalgran:
Thanks for the review.

Issue #1228, opened 2021-02-02
Cookie preference section appears on all profiles, not just the profile of the logged in user


jwalgran:
## Overview

The Cookie preference section added in #1213 appears on all profiles, not just the profile of the logged in user.

Reported by @katieashaw 

### Expected Behavior

The cookie preference section only appears on the profile page for the currently logged in user.

### Actual Behavior

The cookie preference section appears on all profiles, whether a user is logged in or not.

### Steps to Reproduce

* Log out
* Browse a profile page

### Demo

![Screenshot 2021-02-02 at 15 03 57](https://user-images.githubusercontent.com/17363/106630496-1ed8d480-6539-11eb-9926-0a92b623ffb8.png)

PR #1229, opened 2021-02-02
Only show cookie preferences for the logged in user

jwalgran:
## Overview

It is confusing to show the cookie preferences section on the profile pages for users other than the currently logged in user since it appears that they would have the ability to change the preferences for other users.

Connects #1228

## Testing Instructions

These steps assume `./scripts/resetdb` has been run

* Log out
* Browse http://localhost:6543/profile/2 and verify that the cookie preferences section is not shown.
* Log in as c2@example.com:password
* Browse http://localhost:6543/profile/2 and verify that the cookie preferences section appears
* Browse http://localhost:6543/profile/3 and verify that the cookie preferences section is not shown.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


TaiWilkin (approved):
Working well!

jwalgran:
Thanks for the review.

Issue #1136, opened 2020-10-15
Add model and API methods to allow for reporting a facility as closed or reopened


jwalgran:
## Overview

Add models and an API to support mark a facility as closed.

For reference, here are the wireframes of the full "report as closed" feature: https://app.abstract.com/projects/441633f0-7d8c-11e9-af1e-972cbddca60a/branches/c06f3eda-8595-456e-a80f-397a63bbdcec/collections/210f6bf8-3804-4329-94c7-5c5912e7dea8

### Describe the solution you'd like

- Add a `FacilityActivityReport` model
  - required FK to facility
  - `reported_by_user` required FK to user
  - `reported_by_contributor` required FK to contributor
  - `closure_state` required choice OPEN or CLOSED
  - `created_at` and `updated_at` auto fields
  - `approved_at` nullable date time
  - `status` required default PENDING
    - PENDING, CONFIRMED, REJECTED
  - `status_change_reason `nullable text
  - `status_change_by` nullable FK to user 
  - `status_change_date` nullable datetime
  - Historical model integration
- Add a nullable `is_closed` field to Facility 
- Add an API to submit a facility as open or closed
  - Add method(s) to the facility ViewSet to handle `POST /facilities/{id}/report` and `POST /facilities/{id}/report/{id}/approve` and `POST /facilities/{id}/report/{id}/reject`
    - `approve` and `reject` view functions should accept a `status_change_reason` string and set the `status`, `status_change_by`, and `status_change_date` on the `FacilityActivityReport` and set the `is_closed` field on the `Facility` appropriately.
    - `approve` view function should set `approved_at`
  - Include the required docstring comments and related code so that Swagger API docs show the new endpoint.

designmatty:
@jwalgran @dboyer Can you take a look at the designs here https://share.goabstract.com/ca589b7e-f23f-49a5-9f50-595d38a596af

jwalgran:
Thanks for linking these designs. I added comments to Abstract.

PR #1231, opened 2021-02-03
Allow for reporting a facility as closed or reopened

TaiWilkin:
## Overview

Adds models and an API to support marking a facility as closed or opened.

Connects #1136 

## Demo

Screen Shot 2021-02-03 at 11 39 59 AM
Screen Shot 2021-02-03 at 11 39 36 AM
Screen Shot 2021-02-03 at 11 39 13 AM

## Testing Instructions

* Run `vagrant ssh` and `./scripts/manage migrate` and `./scripts/server`
* Go to [:6543](http://localhost:6543/) and save the OAR ID of a facility 
- Go to http://localhost:8081/api/docs/#!/facilities/facilities_read and authenticate your user.
- Fetch the facility by the ID you saved
    - The facility should have `is_closed: null`
- Go to http://localhost:8081/api/docs/#!/facilities/facilities_report
- Enter the facility id and 
    ``` 
    {
         "closure_state": "CLOSED",
         "reason_for_report": "This facility was closed."
     } 
     ```
- Submit. 
    - The request should return a FacilityActivityReport with `closure_state: ‘CLOSED’` and `status: ‘PENDING’` and appropriate user, contributor, and created/updated date times. Note the ID.
- Go to http://localhost:8081/api/docs/#!/facility-activity-reports/facility_activity_reports_approve and enter the report ID and ```{
    "status_change_reason": "The facility report was confirmed."
}```
- Submit. 
    - The request should return the same FacilityActivityReport with `status: “CONFIRMED”`, all of the status_change_ fields filled out, and `approved_at` set to the current date time.
- Go to http://localhost:8081/api/docs/#!/facilities/facilities_read and fetch the facility again
    - It should have `is_closed: true`
- Repeat the process with a new report, but enter `"closure_state": “OPEN”`
    - Approve the facility report. The facility should now have `is_closed: false`
- Create a new facility report with `"closure_state": “CLOSED”`
- Go to http://localhost:8081/api/docs/#!/facility-activity-reports/facility_activity_reports_reject 
- Enter the facility ID and ```{
    "status_change_reason": "The facility report is incorrect."
}```
    - Submit. The response should have `"status": "REJECTED”` and the status_change_ fields should be filled out appropriately. 
- Fetch the facility. 
     - Its status should not have changed. 

## Checklist

- [ ] CI passes after rebase


jwalgran:
Looking at this now.

src/django/api/views.py
```diff
@@ -3263,3 +3309,110 @@ def retrieve(self, request, pk=None):
...
+
+        {
+            "status_change_reason": "CLOSED"
+        }
+    """
+    queryset = FacilityActivityReport.objects.all()
+    serializer_class = FacilityActivityReportSerializer
+    permission_classes = (IsRegisteredAndConfirmed,)
+
+    @action(detail=True, methods=['PUT'],
```
jwalgran, line 3359: For consistency with the rest of the API methods we should consider changing the path+verb combination for the approve and reject endpoints.

Pending facility matches have /approve and /reject endpoints that response to POST. It is conventional that "action" endpoints, where the path ends in a word rather than a resource ID, are POST requests.

PUT is actually a reasonable verb choice in this context, because we are updating an existing resource, not adding a new one. If we continue to use PUT, we should consider changing the path to `PUT http://localhost:8081/api/facility-activity-reports/1`.

I don't have strong feelings, but I lean toward changing the verb to POST for consistency with the facility match approve/reject endpoints.

src/django/api/views.py
```diff
@@ -3263,3 +3309,110 @@ def retrieve(self, request, pk=None):
...
+            permission_classes=(IsRegisteredAndConfirmed,),
```
jwalgran, line 3360: We should change this permission class to `IsAdminUser` since we don't want to allow people to approve their own reports.

src/django/api/views.py
```diff
@@ -2084,6 +2098,38 @@ def dissociate(self, request, pk=None):
...
+            facility = Facility.objects.get(pk=pk)
+        except Facility.DoesNotExist:
+            raise NotFound('Facility with OAR ID {} not found'.format(pk))
+
+        contributor = request.user.contributor
+        facility_activity_report = FacilityActivityReport.objects.create(
+            facility=facility,
+            reported_by_user=request.user,
+            reported_by_contributor=contributor,
+            closure_state=request.data.get('closure_state'),
```
jwalgran, line 2126: I was able to submit a report with `closure_state` set to "FOO". As I understand it Django's validation does not kick in unless `full_clean` is called on the model. 

Screen Shot 2021-02-04 at 10 23 21 AM

jwalgran (commented):
The structure of this update is solid and I was able to successfully run through the testing instructions and exercise the workflow. There are a few changes that I think we need to make before merging.

src/django/api/views.py
TaiWilkin: I went with `POST` for consistency. Thanks for explaining why it's used in this case!

src/django/api/views.py
TaiWilkin: I added a `full_clean` before `save()` for this call. I ended up adding `blank=True` to a few of the fields that are nullable because they were giving validations errors after adding the check, so I updated the migration.

TaiWilkin:
Ready for another look

src/django/api/views.py
```diff
@@ -2125,6 +2125,7 @@ def report(self, request, pk=None):
             reported_by_contributor=contributor,
             closure_state=request.data.get('closure_state'),
             reason_for_report=request.data.get('reason_for_report'))
+        facility_activity_report.full_clean()
```
jwalgran, line 2128: It looks like we need to go a bit further with this validation. When I tried POSTing "foo" as a status the API returned a full Django HTTP unhandled exception page.

Screen Shot 2021-02-08 at 11 08 00 AM

We have some existing code that handles a similar situation, but we will want to use a different exception.

https://github.com/open-apparel-registry/open-apparel-registry/blob/34ecf9d3e4a71235c4cd0fa3f8bbe71fd613e078/src/django/api/serializers.py#L68-L76

In summary, we have the catch the django.core.ValidationError that is raised by `full_clean` and raise an `api.exceptions.BadRequestException` instead. DRF will catch this and render a JSON response instead of returning HTML.

jwalgran (commented):
Updates to the model and the HTTP verbs look great. There is one more adjustment to the status code validation that I would recommend to ensure that the error response is JSON and has a 400 status code.

src/django/api/views.py
TaiWilkin: Updated this in c65d115. Thanks for the assistance!

TaiWilkin:
Ready for another look.

jwalgran (approved):
Nice fixes. Sorry for the start/stop review process and thanks for working through the changes.

TaiWilkin:
Thank you for all the help on this!

Issue #1242, opened 2021-02-16
Display issues on confirm / reject page


katieashaw:
## Overview

There are display issues with the way that potential matches are being presented to users on the confirm / reject page

### Expected Behavior

Each potential match should align with the associated confirm / reject buttons

### Actual Behavior

The alignment of confirm / reject buttons is off. 

### Demo

![Screenshot 2021-02-14 at 16 54 07](https://user-images.githubusercontent.com/46344729/108051240-8ff0af80-7042-11eb-80f4-3101c3def5da.png)

jwalgran:
We started work on fixing this yesterday and we should have it resolved today (#1243) and deployed in the near future.

jwalgran:
Thanks for reporting this. I am closing this specific issue because we tracked our work fixing is as a continuation of #1185.

PR #1243, opened 2021-02-16
Align Facility Matches

TaiWilkin:
## Overview

Previously, the facility list confirmation row component contained two
rows:
- one row for the facility item
- one row containing all of the facility matches and their headers

This caused the confirm/reject buttons to become misaligned with the
facility matches to which they referred.

Now, the facilities are split into: 
- one row for the facility item,
- one row for the facility match headers,
- plus one row for each facility match. 

This keeps the items aligned.

Connects #1242, #1189

## Demo

Screen Shot 2021-02-16 at 8 30 23 AM

## Testing Instructions

* Run `./scripts/server`
* View list items with [potential matches](http://localhost:6543/lists/8?status=POTENTIAL_MATCH&page=1&rowsPerPage=20)
    * The items should be aligned.
    * Clicking the match names should take you to the match facility. 
    * Clicking confirm/reject should correctly confirm or reject the match. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Removed
 
 ### Fixed
+- Align Facility Matches [#1243](https://github.com/open-apparel-registry/open-apparel-registry/pull/1243)
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


jwalgran (approved):
This is excellent work. A simplification of the existing code that also fixes a layout bug. I tested single and multi row matches and uploads a list with a parse error to exercise "error rows." I did not find and regressions in functionality.

TaiWilkin:
Thanks for the review!

Issue #1245, opened 2021-02-16
Add a waffle switch around the public facing facility report UI


jwalgran:
Add a waffle switch around the public-facing facility report UI. 
- Add a waffle switch for `report_a_facility`
- Wrap the public-facing facility report UI components (the status ribbons, status list, and 'Report this facility' button) in FeatureFlags utilizing the new waffle switch
  - The APIs don't need to be hidden
  - The admin dashboard doesn't need to be hidden

See [here](https://github.com/open-apparel-registry/open-apparel-registry/pull/1041/commits/4d3a026fc28021439d727039676924fc18a9f94d) for implementation examples.

PR #1246, opened 2021-02-17
Add report waffle switch

TaiWilkin:
## Overview

Add a waffle switch around the public facing facility report UI so that these features can be reviewed and approved before being publicly accessible in production. 

Connects #1245 

## Demo

Screen Shot 2021-02-17 at 11 13 36 AM

## Testing Instructions

* On `develop`, ensure you have at least one closed facility.
* On this branch, run `./scripts/update` and `./scripts/server`
* Navigate to [Switches in Admin](http://localhost:8081/admin/waffle/switch/)
    * report_a_facility should be inactive
* Navigate to [facilities](http://localhost:6543/)
    * You should not see any closure ribbons
* Navigate to a facility which is closed
     * You should not see any status ribbon
     * You should not see a status list
     * You should not see the ‘report facility closure’ button
* Navigate to [Switches in Admin](http://localhost:8081/admin/waffle/switch/) and set report_a_facility to active
* Navigate to [facilities](http://localhost:6543/)
    * You should see closure ribbons
* Navigate to a facility which is closed
     * You should see a status ribbon
     * You should see a status list
     * You should see the ‘report facility closure’ button

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add API for reporting a facility as closed or reopened [#1231](https://github.com/open-apparel-registry/open-apparel-registry/pull/1231)
 - Confirm or reject status reports [#1235](https://github.com/open-apparel-registry/open-apparel-registry/pull/1235)
 - Report Facility Status from Facility Details [#1239](https://github.com/open-apparel-registry/open-apparel-registry/pull/1239)
+- Add Report Waffle Switch [#1246](https://github.com/open-apparel-registry/open-apparel-registry/pull/1246)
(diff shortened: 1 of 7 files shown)


jwalgran (approved):
Confirmed. :+1:

Did you trip over anything when you were implementing this or wish some part of the implementation was easier? I want to make sure this system is as frictionless as possible so we are encouraged to use it more often as we build out more larger features this year.

TaiWilkin:
It was fairly straightforward! With the current implementation of the switches, I don't feel like we need to shy away from using them as needed. I do think that we should remove the FeatureFlag components for the following production release after we're certain the feature in question is good to go, to avoid having unnecessary components in the code.

Issue #1248, opened 2021-02-17
Add is_closed to the facility CSV


jwalgran:
## Overview

Add `is_closed` to the facility CSV download

### Describe the solution you'd like

* Include `is_closed` to the facility CSV download

PR #1249, opened 2021-02-17
Add Closure to CSV Download

TaiWilkin:
## Overview

Add the facility closure status to the CSV downloads in the case that
the `report_a_facility` switch is on.

Connects #1248 

## Demo

Screen Shot 2021-02-17 at 3 10 53 PM

## Testing Instructions

* Run `./scripts/server`
* Turn off the Reports switch and download the facilities CSV
    - [ ] It should not have an 'is_closed' column
* Turn on the Reports switch and download the facilities CSV
    - [ ] It should have an 'is_closed' column with appropriate data

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/actions/logDownload.js
+++ b/src/app/src/actions/logDownload.js
@@ -38,6 +38,7 @@ export function logDownload() {
 
             const vectorTileFlagIsActive = get(featureFlags, 'flags.vector_tile', false);
             const ppeIsActive = get(featureFlags, 'flags.ppe', false);
+            const reportsAreActive = get(featureFlags, 'flags.report_a_facility', false);
 
             const path = `${window.location.pathname}${window.location.search}${window.location.hash}`;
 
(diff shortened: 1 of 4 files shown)


src/app/src/actions/logDownload.js
```diff
@@ -38,6 +38,7 @@ export function logDownload() {
 
             const vectorTileFlagIsActive = get(featureFlags, 'flags.vector_tile', false);
             const ppeIsActive = get(featureFlags, 'flags.ppe', false);
+            const reportsAreActive = get(featureFlags, 'flags.report_a_facility', false);
```
jwalgran, line 41: 👍 Nice integration with our new switch

jwalgran (approved):
Nice, clear, correct implementation. I like seeing a new JS test.

While this does match the spec of the feature as listed in the issue, when I take a look at the CSV after closing a facility and reopening one I see an opportunity to make this a little friendlier to less technical users

Screen Shot 2021-02-17 at 8 14 05 PM

An alternate formatting I came up with is to output the string "CLOSED" in the `is_closed` column if `Facility.is_closed` is `True`, otherwise leave the column blank. This reduces the number of possible states from 3 to 2 and avoids some of the "double negativeness" of having True representing closed. I'm interested in what you think about this alternate formatting, if you have another suggestion, or think it is more appropriate to keep it as is currently implemented.

TaiWilkin:
I agree that 'CLOSED' is going to be clearer for non-technical users. I've updated the code to use 'CLOSED' when True and otherwise display nothing, and updated the test to check for handling both True and False.

Issue #1255, opened 2021-02-19
Manage `embed` Query Parameter


TaiWilkin:
The embedded map view will be denoted by the query parameter `embed=1`. 
- Add a consistent method to access the `embed` parameter as a prop
- Update the existing filter management code as necessary to ensure the `embed` parameter isn't lost when filters change
- The contributor shouldn't be cleared when the reset button is hit in embed mode

PR #1257, opened 2021-02-23
Manage Embed Query Parameter

TaiWilkin:
## Overview

The embedded map view is denoted by the query parameter `embed=1`.

- Adds the embed parameter as a key in a new embeddedMap reducer
- Updates the existing filter management code to ensure the embed parameter isn't lost when filters change
- Prevents the contributor from being cleared when the reset button is hit in embed mode

Connects #1255 

## Testing Instructions

* Run `./scripts/server` in vagrant
* Navigate to http://localhost:6543/ and search with various combinations of filters
    * The site should behave as before
* Navigate to http://localhost:6543/?contributors=8 and click 'reset'
    * The site should behave as before
    * The contributor filter should be cleared
* Navigate to http://localhost:6543/?embed=1
    * The embed code should remain in the url
    * In the Developer console, confirm that state.embeddedMap.embed = "1".
* Click 'reset'
    * The embed code should remain in the url
* Navigate to http://localhost:6543/?contributors=8&embed=1
    * Click 'reset'
    * The embed code and the contributor filter should remain selected

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/reducers/FiltersReducer.js
+++ b/src/app/src/reducers/FiltersReducer.js
@@ -71,7 +71,12 @@ export default createReducer({
     [updatePPEFilter]: (state, payload) => update(state, {
         ppe: { $set: payload },
     }),
-    [resetAllFilters]: () => initialState,
+    [resetAllFilters]: (state, payload) => update(initialState, {
+        contributors: {
+            $set: payload ? state.contributors
+                : initialState.contributors,
+        },
+    }),
     [updateAllFilters]: (_state, payload) => payload,
     [completeFetchContributorOptions]: maybeSetFromQueryString('contributors'),
     [completeFetchContributorTypeOptions]: maybeSetFromQueryString('contributorTypes'),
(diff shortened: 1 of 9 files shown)


src/app/src/reducers/FiltersReducer.js
```diff
@@ -71,7 +71,12 @@ export default createReducer({
     [updatePPEFilter]: (state, payload) => update(state, {
         ppe: { $set: payload },
     }),
-    [resetAllFilters]: () => initialState,
+    [resetAllFilters]: (state, payload) => update(initialState, {
```
jwalgran, line 74: Although it is conventional to call the second argument `payload` I think we should consider using `isEmbedded` in this case. It is more self documenting and I can argue that since we are not putting the payload into the state, just using it to change behavior, that the break with convention is justified.

jwalgran (approved):
Worked perfectly in my testing. Clear, well written extension of the existing functionality. Nicely done.

Issue #983, opened 2020-03-05
Enable users to filter down to individual list level


katieashaw:
## Overview

Enable users to filter down to the individual list level on a Contributor page

### Is your feature request related to a problem? Please describe.

Whilst Contributors are already able to upload multiple lists to their OAR accounts, when searching the OAR users cannot filter down to the individual list level. 

### Describe the solution you'd like

- Add a `facility-lists` API that returns the active lists for one or more contributors
  - If multiple contributor IDs are passed, the response should contain all active lists from all specified contributors
  - Return the list name, and list ID for each list. Sort by list name.
- When one or more contributors are selected on the search page show a contributor list filter dropdown below it
  - Indent it to emphasize that it is related to the contributors list (and consider indenting the existing "Show only shared facilities" checkbox to match)
  - Populate it with the list API response
- Add a `lists` query string filter parameter that behaves like the `contributors` query string filter parameter.
- Add filtering by list ID(s) to `GET /api/facilities`
  -  Ensure that the option appears correctly in `/api/docs/#!/facilities/facilities_list`
- Add links to the lists shown on the contributor profile page that set the contributor and list query string arguments appropriately so that clicking the link runs the appropriate search. 

### Additional Context

Examples for use of this filtering would be the Accord on Fire and Building Safety in Bangladesh consolidating their two separate accounts into one, with users being able to filter to each separate upload; users being able to filter by the different lists for tiers of supply chains uploaded by brands; certification schemes having separate lists for different levels of certification, e.g. Gold, Silver, Bronze, which users can then filter by.

katieashaw:
A question for @jwalgran on this: would it make any sense to incorporate this into the work on the map plug-in?

jwalgran:
If we are talking about the "embedded map" feature, where contributors can add an OAR-created map of their facilities to their own website, then yes. We plan to allow country and contributor type filters to still work on the embedded map, and it would make sense to also include this new list filter.

katieashaw:
Yes, that's what I meant - so that's great to hear, thank you!

dboyer:
Related to #454

katieashaw:
During the first week of February, we received two enquiries in quick succession relating to historical data in the OAR. Where organisations choose to have multiple active lists associated with their accounts, the ability to filter down to the individual list level would provide those users with the information they are looking for.

[...]

PR #1272, opened 2021-03-08
Fix Filtering Bug

TaiWilkin:
## Overview

In the development database, many associated models share the same id; however, in staging and production this is significantly less common. When introducing list filtering, facilities were being filtered based on matching the id to the incorrect model, which was not caught due to the ids matching in all development cases. The filtering is now based on correct models. 

In addition, a new set of tests was added with manually set, non-matching ids in order to verify the fix is working locally. 

Connects #983 

## Notes

Much of the code setting up the tests for this is a duplicate of the code in the `FacilityAPITestCaseBase`, but as I was applying manually set IDs to the data, I felt it was better to separate out the new test cases. 

## Testing Instructions

* Visually confirm that the test cases are accurately testing the use case and the believed source of the bug. 
* Run the tests and confirm they are passing.
    * (Note: If you replace line 1237 in modes.py with `facilitylistitem__source_id__in=lists).values_list(`, the previous text of that line, and run the tests, you can see that they do not pass with the previous code)
* Complete the tests previously described in [#1264](https://github.com/open-apparel-registry/open-apparel-registry/pull/1264)
  - Navigate to [/api/docs/#!/contributor-lists/contributor_lists_list](http://localhost:8081/api/docs/#!/contributor-lists/contributor_lists_list)
  - Enter 2 as the contributor ID 
      - [x] You should get a list of contributor lists as ID/name pairs
  - Navigate to http://localhost:8081/api/docs/#!/facilities/facilities_list
  - Set the list's param to 2
      - [x] You should get a list of facilities from the Summer 2019 Compliance List
  - Navigate to http://localhost:6543/
      - [x] You should not see the contributor lists
  - Select Service Provider A
      - [x] You should see the contributor lists select
  - Select the Summer 2019 Compliance List and search
      - [x] You should get a list of facilities from that list
      - [x] The map should load correctly
  - Select Civil Society Organization A
      - [x] The previously selected listed should no longer be selected
      - [x] You should see multiple lists in the dropdown
  - Select Summer 2019 Apparel List and Summer 2019 Compliance List and search
      - [x] You should get a list of facilities from both lists
  - Refresh the page
      - [x] Both contributors and both lists should remain selected
  - Navigate to http://localhost:6543/profile/2 
      - [x] The Facility List names should be links
  - Click the link
      - [ ] The map should open with the appropriate contributor and list selected

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/serializers.py
+++ b/src/django/api/serializers.py
@@ -226,9 +226,17 @@ class UserProfileSerializer(ModelSerializer):
 
 
 class FacilityListSummarySerializer(ModelSerializer):
+    contributor_id = SerializerMethodField()
+
     class Meta:
         model = FacilityList
-        fields = ('id', 'name', 'description')
+        fields = ('id', 'name', 'description', 'contributor_id')
+
+    def get_contributor_id(self, facility_list):
+        try:
+            return facility_list.source.contributor.id
+        except Contributor.DoesNotExist:
+            return None
 
 
 class FacilityListSerializer(ModelSerializer):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Align Facility Matches [#1243](https://github.com/open-apparel-registry/open-apparel-registry/pull/1243)
 - Update Reject Button Text [#1244](https://github.com/open-apparel-registry/open-apparel-registry/pull/1244)
 - Fix vagrant provisioning failure by pinning pip python version [#1262](https://github.com/open-apparel-registry/open-apparel-registry/pull/1262)
+- Fix list filtering [#1272](https://github.com/open-apparel-registry/open-apparel-registry/pull/1272)
 
 ### Security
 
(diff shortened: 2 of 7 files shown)


src/django/api/serializers.py
```diff
@@ -226,9 +226,17 @@ def get_is_verified(self, user):
...
     class Meta:
         model = FacilityList
-        fields = ('id', 'name', 'description')
+        fields = ('id', 'name', 'description', 'contributor_id')
+
+    def get_contributor_id(self, facility_list):
+        try:
+            return facility_list.source.contributor.id
+        except Contributor.DoesNotExist:
+            return False
```
jwalgran, line 239: Consider returning `None` rather than `False` here. For a function that returns a number, `None` is a more natural "couldn't find it" value.

jwalgran (approved):
Nice use of tests to ensure the fixes worked. I retested and did not find any regressions. I opened up a tech debt issue to cover updating `resetdb` to give us a better chance of not running into bugs like this again.

I would like to include this as the final commit in a 2.40.0 release. Is it possible to change the target of this to `develop` before merging?

Issue #1252, opened 2021-02-18
Upgrade CRA and craco and ensure Fast Refresh works


jwalgran:
## Overview

Upgrade CRA and craco and ensure Fast Refresh works

### Describe the solution you'd like

- Upgrade CRA and craco and ensure Fast Refresh works

### Additional Context

We recently did this on another project https://github.com/azavea/pwd-stormwater-allocation/pull/2213

PR #1284, opened 2021-03-15
Adjust settings so fast refresh works in development

jwalgran:
## Overview

After upgrading to CRA 4 and craco hot module reloading was not working. We referenced another project that was using fast refresh and copied in settings to match.

We removed the environment variables from the `start` script in `package.json` because they duplicate variable set in docker-compose.yml.

Within `docker-compose.yml` we reduced `CHOKIDAR_INTERVAL`, set `working_dir`, and switched to node 12 to match another project using fast refresh.

On other projects we have kept the versions of react and react-dom in sync, so we updated the version on `react-dom` in `package.json`.

Connects #1252

## Testing Instructions

* Stop any running services
* Browse the site, edit a JSX file, and verify that fast refresh works.
 
## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Adjust settings so fast refresh works in development [#1284](https://github.com/open-apparel-registry/open-apparel-registry/pull/1284)
+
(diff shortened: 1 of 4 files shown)


TaiWilkin (approved):
Great job getting this working! It's running beautifully.

jwalgran:
Thanks for the review.

Issue #928, opened 2019-12-13
Email notification for users once their list has been processed


katieashaw:
## Overview

Could we send an automated email to users to alert them when the OAR has finished processing their list after upload? 

### Is your feature request related to a problem? Please describe.

The length of time the system takes to process a list upload varies dependent on multiple factors. To save users having to repeatedly check back to see if their upload has finished processing, and to prompt them to complete the Confirm / Reject process, could we set up an automated email notification letting users know when their upload has been processed?

### Describe the solution you'd like

- Add a 4th batch processing step dependent on the match step
  - Send email to the contributor including a direct link to the completed list
 
### Additional Context

Katie can provide text for the automated email, when this request is worked on.

PR #1290, opened 2021-03-23
Notify Contributors when List Completes Processing

TaiWilkin:
## Overview

The length of time the system takes to process a list upload varies
dependent on multiple factors. To save users having to repeatedly check
back to see if their upload has finished processing, and to prompt them
to complete the Confirm / Reject process, we send an automated email
notification letting users know when their upload has been processed.

Connects #928 

## Demo

Screen Shot 2021-03-23 at 12 22 12 PM
Screen Shot 2021-03-23 at 12 23 12 PM

## Notes

We intend to later adjust this to send a different message to contributors who have additional lists already in the database. 

## Testing Instructions

* In vagrant, run `./scripts/manage batch_process -l 7 -a "notify_complete"`
    - [ ] An email should print with the correct contributor email (list admin). 
    - [ ] The email text should match [provided text](https://docs.google.com/document/d/1PlrtRFlf_FiKtMB93YQM7ON7BPBqrB4ZZVmAJuC0pQ4/edit)

## Checklist

- [ ] Temporary `DROP` commit has been removed 
- [ ] CI passes after rebase

### DIFF
--- a/deployment/terraform/iam.tf
+++ b/deployment/terraform/iam.tf
@@ -121,6 +121,11 @@ resource "aws_iam_role_policy_attachment" "ec2_service_role" {
   policy_arn = "${var.ec2_service_role_policy_arn}"
 }
 
+resource "aws_iam_role_policy_attachment" "ses_send_email_from_batch" {
+  role       = "${aws_iam_role.container_instance_ec2.name}"
+  policy_arn = "${aws_iam_role_policy.ses_send_email.policy.arn}"
+}
+
 resource "aws_iam_instance_profile" "container_instance" {
   name = "${aws_iam_role.container_instance_ec2.name}"
   role = "${aws_iam_role.container_instance_ec2.name}"
--- a/deployment/terraform/iam.tf
+++ b/deployment/terraform/iam.tf
@@ -121,9 +121,10 @@ resource "aws_iam_role_policy_attachment" "ec2_service_role" {
   policy_arn = "${var.ec2_service_role_policy_arn}"
 }
 
-resource "aws_iam_role_policy_attachment" "ses_send_email_from_batch" {
+resource "aws_iam_role_policy" "ses_send_email_from_batch" {
+  name       = "ses${var.environment}EmailSendingPolicy"
   role       = "${aws_iam_role.container_instance_ec2.name}"
-  policy_arn = "${aws_iam_role_policy.ses_send_email.policy.arn}"
+  policy     = "${data.aws_iam_policy_document.ses_send_email.json}"
 }
 
 resource "aws_iam_instance_profile" "container_instance" {
--- a/src/django/api/mail.py
+++ b/src/django/api/mail.py
@@ -1,6 +1,7 @@
 from django.core.mail import send_mail
 from django.conf import settings
 from django.template.loader import get_template
+from api.models import FacilityList
 
 from api.countries import COUNTRY_NAMES
 
@@ -21,6 +22,19 @@ def make_facility_url(request, facility):
         make_oar_url(request),
         facility.id,
     )
+
+
+def make_facility_list_url(list_id):
+    if settings.ENVIRONMENT == 'Development':
+        protocol = 'http'
+        host = 'localhost:6543'
+    else:
+        protocol = 'https'
+        if settings.ENVIRONMENT == 'Staging':
+            host = 'staging.openapparel.org'
+        else:
+            host = 'openapparel.org'
+    return '{}://{}/lists/{}'.format(protocol, host, list_id)
 
 
 def make_claimed_url(request):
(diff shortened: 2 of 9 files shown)


src/django/api/mail.py
```diff
@@ -336,6 +348,31 @@ def send_report_result(report):
         subj_template.render().rstrip(),
         text_template.render(report_dictionary),
         settings.DEFAULT_FROM_EMAIL,
-        [settings.NOTIFICATION_EMAIL_TO],
+        [report.reported_by_contributor.admin.email],
```
TaiWilkin, line 351: This is an existing bug - the email should have been being sent to the contributor admin

jwalgran:
Looking at this now.

jwalgran:
Code looks good. I am going to push up a temporary commit to test the Batch behavior in staging.

jwalgran (commented):
Testing batch stuff is tricky. After pushing a commit to get this deployed to staging I tested an upload and saw that the job failed. Looks like our batch job container needs to run with an extra permission.

>botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the SendRawEmail operation: User `arn:aws:sts::249322298638:assumed-role/ecsStagingContainerInstanceProfile/i-0224722b5f11575e7' is not authorized to perform `ses:SendRawEmail' on resource `arn:aws:ses:eu-west-1:249322298638:identity/staging.openapparel.org'

We already have Terraform to add this permission to our `app_task_role`

https://github.com/open-apparel-registry/open-apparel-registry/blob/a0742b77cc04fc9ca883d2d3cbf05cfd5b76d117/deployment/terraform/iam.tf#L43-L47

I think it will be relatively straightforward to enable this permission for our "cli" role as well.

jwalgran:
If you would like to pair on figuring out the additional Terraform statements required send me a calendar invitation. I would like us to attempt to fix it before bringing in ops and if we are successful, having them review our changes.

Since we have this branch deployed to staging, I think we can iteratively make changes and run `terraform plan` in development (pointed at staging).

jwalgran:
I paired on this with @TaiWilkin and thought that the change below may be sufficient to get the email permission applied correctly. We tried manually planning and applying this to staging from development but due to what I believe is user error, did not have a successful test.

```diff
diff --git a/deployment/terraform/iam.tf b/deployment/terraform/iam.tf
index 8f04892..dfaa782 100644
--- a/deployment/terraform/iam.tf
+++ b/deployment/terraform/iam.tf
@@ -46,6 +46,12 @@ resource "aws_iam_role_policy" "ses_send_email" {
   policy = "${data.aws_iam_policy_document.ses_send_email.json}"
 }
+resource "aws_iam_role_policy" "ses_send_email_from_batch" {
+  name   = "SESSendEmailFromBatch"
+  role   = "${aws_iam_role.container_instance_batch.name}"
+  policy = "${data.aws_iam_policy_document.ses_send_email.json}"
+}
+
 data "aws_iam_policy_document" "batch_describe_and_submit" {
   statement {
     effect = "Allow"
```

colekettler:
(Apologies for the force push, realized I had an intermediate commit in there, not something I would typically do on somebody else's PR)

deployment/terraform/iam.tf
```diff
@@ -121,6 +121,12 @@ resource "aws_iam_role_policy_attachment" "ec2_service_role" {
   policy_arn = "${var.ec2_service_role_policy_arn}"
 }
 
+resource "aws_iam_role_policy" "ses_send_email_from_batch" {
+  name       = "ses${var.environment}EmailSendingPolicy"
+  role       = "${aws_iam_role.container_instance_ec2.name}"
+  policy     = "${data.aws_iam_policy_document.ses_send_email.json}"
+}
```
colekettler, line 128: By using the [`aws_iam_role_policy`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role_policy) resource, we're creating an inline policy attached to the container instance role. It reuses the contents of the `ses_send_email` policy document, but it is creating separate inline policy objects for our application container role and our Batch container role!

src/django/api/mail.py
```diff
@@ -23,6 +24,17 @@ def make_facility_url(request, facility):
     )
 
 
+def make_facility_list_url(list_id):
+    if settings.ENVIRONMENT == 'Development':
+        protocol = 'http'
+        host = 'localhost:6543'
+    else:
```
jwalgran, line 31: When I tested this on staging I realized that the link in the email pointed to production rather than staging. For correctness and ease of testing we should consider something like this:

```python
if settings.ENVIRONMENT == 'Development':
    protocol = 'http'
    host = 'localhost:6543'
else:
    protocol = 'https'
    if settings.ENVIRONMENT == 'Staging':
        host = 'staging.openapparel.org
    else:
        host = 'openapparel.org'
```

jwalgran (approved):
I correctly received and email from staging when my test list was processed.

Screen Shot 2021-03-31 at 11 07 08 AM

I suggested a correction to the link when we are in staging but otherwise this looks good.

colekettler (approved):
Looks like this works and the Terraform changes are in line with our typical approach. Thank you @TaiWilkin and @jwalgran for kicking off the Terraform changes!

src/django/api/mail.py
TaiWilkin: Good catch, thanks!

Issue #828, opened 2019-09-24
Add a "Back to facility claims" button at the top of each individual facility claim page


katieashaw:
## Overview

At the moment, there is only a link to click back to the Dashboard at the top of an individual "facility claim" page in the admin interface. Could we add a link to take you back to the list of facility claims, rather than just the Dashboard link, to reduce the number of clicks required when working through facility claims, please. 

### Is your feature request related to a problem? Please describe.

Often, a factory group will claim multiple facilities on the tool in one go, which requires working through several submissions when reviewing claims. At the moment, having reviewed a claim - which may just be adding a holding note about the status of the claim - the admin has to:

- Enter a claim
- Review, approve, deny, add a note etc
- Click back to the Dashboard
- Click "View facility claims"
- Being the review process on the next facility claim

### Describe the solution you'd like

Add a link back to the "View facility claims" at the top of each individual facility claim page, in a similar way to how you can quickly reset on the "merge facility matches" or "adjust facility matches" pages, avoiding navigating back through the main Dashboard.

katieashaw:
Now that we're starting to receive more facility claims to review, I'm reminded of this request. Could we bump it up the priority list, please @dboyer @jwalgran

PR #1307, opened 2021-04-08
Add Link Back to Facility Claims

TaiWilkin:
## Overview

Previously, there was only a link to click back to the Dashboard at the top of an individual "facility claim" page in the admin interface. Adds a link to take you back to the list of facility claims, rather than just the Dashboard link, to reduce the number of clicks required when working through facility claims.

Connects #828 

## Demo

Screen Shot 2021-04-08 at 1 33 27 PM

## Testing Instructions

* Run `./scripts/server`
* Create a new facility claim
* Login as `c1@example.com`
* Navigate to `http://localhost:6543/dashboard/claims`
* Select a claim to go to the claim details page
     - [ ] There should be a link to the facility claims page
     - [ ] Clicking the link should take you to the facility claims page

## Checklist

- [ ] CI passes after rebase


jwalgran:
Looking at this now.

jwalgran (approved):
This works well and is a nice extension of the existing system.

Going a "third level" deep in this navigation prompted me to think about how it is possible that Material design 3.x may have has a component for this, and sure enough they do:

https://v3.material-ui.com/lab/breadcrumbs/

The last example on that page even shows an integration with React Router.

Since this was such a modest change, I think we should consider spending a little extra time looking at `Breadcrumbs` so see if we could quickly drop it in and drop some custom code in favor of library code.

TaiWilkin:
@jwalgran Would appreciate another look at this - using the Material UI method, the code is DRYer but less configurable

src/app/src/components/Dashboard.jsx
```diff
@@ -81,107 +99,60 @@ function Dashboard({ userWithAccessHasSignedIn, fetchingSessionSignIn }) {
...
-                                        'Facility Claim Details',
-                                    )()}
-                                </FeatureFlag>
-                            )}
-                        />
-                        <Route
-                            exact
-                            path={dashboardClaimsRoute}
-                            render={() => (
-                                <FeatureFlag
```
jwalgran, line 124: In general I like how Breadcrumbs dried things up but share your concerns about configurability. We don't actually need these specific `FeatureFlag`s any more, but we would be blocked from easily adding one in the future.

jwalgran:
My vote is to not switch to Breadcrumbs in this case. Thanks for looking into it.

TaiWilkin:
Thanks for the review!

Issue #1312, opened 2021-04-14
Connect embed section of setting page to API


jwalgran:
## Overview

Make sure embed setting can be loaded and saved

### Describe the solution you'd like

- No redux needed
- Fetch fields from the API in #1313 and incorporate them into the list of contributor fields along with a hard-coded set of standard fields
  - parent_company
  - type_of_product
  - number_of_workers
- Fetch embed config from the EmbedConfig API and use it to populate the config fields ('width', 'height', 'color', 'font', 'contributor', 'show_other_contributor_information', 'embed_fields') and contributor fields for the configuration settings
  - The list of contributor fields should include both the fields from the nonstandard fields API and the fields from the embed config, but should not contain duplicates. Fields from the embed config take priority in case of duplicates. 
- Update/create the embed config when the user changes the config fields
- Update the contributor embed fields when the user changes the contributor fields

TaiWilkin:
@jwalgran I expanded a bit on this for the sake of completeness. Is there any additional information you think we need here before we officially call it 'undrafted'?

jwalgran:
👍 This looks good. Thanks for adding the additional detail.

Issue #1311, opened 2021-04-14
100% width embed support


jwalgran:
## Overview

Users should be able to set the embedded map to 100% width. This will involve updating the example embedded map code and the sample map on the settings page. 

### Describe the solution you'd like

TODO

Talk to @lederer about implementation details

PR #1329, opened 2021-05-06
Connect embed section of settings page to API

TaiWilkin:
## Overview

Connects the Embed Configuration settings tab to the API, allowing
data fetching, creation, and update. Additionally, adds some UI error
handling, and updates the components to reflect API data structure.

A small, unrelated second commit fixes an error in the proptypes for Feature Flags. 

Connects #1312 

## Demo

Screen Shot 2021-05-06 at 1 44 07 PM

## Testing Instructions

* Login as c1@example.com and navigate to [settings](http://localhost:6543/settings). Select EMBED.
	- [x] You should see an error message.
* Login as a different user, with no nonstandard fields, and navigate to [settings](http://localhost:6543/settings). Select EMBED.
	- [x] The page should show no extra fields.
	- [x] The other settings should be the defaults. 
* Update each of the values on the page. 
	- [x] The fields should update as before. 
* Refresh the page. 
	- [x] The updated values should be maintained. 
* Login as a user with nonstandard fields. 
	- [x] The fields should be listed. 
* Update some (not all) of the fields and refresh the page. 
	- [x] The updated fields should show the new values. 
	- [x] The untouched fields should still appear correctly. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/components/EmbeddedMapConfigWrapper.jsx
+++ b/src/app/src/components/EmbeddedMapConfigWrapper.jsx
@@ -1,0 +1,143 @@
+import React, { useState, useEffect, useCallback } from 'react';
+import { connect } from 'react-redux';
+import { toast } from 'react-toastify';
+
+import { userPropType } from '../util/propTypes';
+import { makeNonStandardFieldsURL, makeEmbedConfigURL } from '../util/util';
+import apiRequest from '../util/apiRequest';
+import EmbeddedMapConfig from './EmbeddedMapConfig';
+import {
+    formatExistingConfig,
+    formatExistingFields,
+    combineEmbedAndNonstandardFields,
+    formatEmbedConfigForServer,
+    getErrorMessage,
+} from '../util/embeddedMap';
+
+function EmbeddedMapConfigWrapper({ user }) {
+    const [loading, setLoading] = useState(true);
+    // Used for field-specific validation errors
+    const [errors, setError] = useState(null);
+    const [embedConfig, setEmbedConfig] = useState(
+        formatExistingConfig(user.embed_config),
+    );
+    const [embedFields, setEmbedFields] = useState(
+        formatExistingFields(embedConfig.embed_fields),
+    );
+
+    const handleError = e => {
+        const errorMessage = getErrorMessage(e);
+        if (Array.isArray(errorMessage)) {
+            toast.error(`Error: ${errorMessage.join(', ')}`, {
+                position: 'bottom-left',
+            });
+        } else {
+            setError(errorMessage);
+        }
+        setLoading(false);
+    };
+
+    const handleSuccess = () => {
+        setLoading(false);
+        setError(null);
+    };
+
+    const fetchFields = useCallback(() => {
+        setLoading(true);
+        apiRequest
+            .get(makeNonStandardFieldsURL())
+            .then(({ data }) => {
+                setEmbedFields(
+                    combineEmbedAndNonstandardFields(
+                        user.embed_config.embed_fields,
+                        data,
+                    ),
+                );
+                handleSuccess();
+            })
+            .catch(e => handleError(e));
+    }, [user]);
+
+    const createConfig = () => {
+        setLoading(true);
+        apiRequest
+            .post(
+                makeEmbedConfigURL(),
+                formatEmbedConfigForServer(embedConfig, embedFields),
+            )
+            .then(({ data }) => {
+                setEmbedConfig(formatExistingConfig(data));
+                handleSuccess();
+            })
+            .catch(e => handleError(e));
+    };
+
+    const updateConfig = () => {
+        setLoading(true);
+        apiRequest
+            .put(
+                makeEmbedConfigURL(embedConfig.id),
+                formatEmbedConfigForServer(embedConfig, embedFields),
+            )
+            .then(({ data }) => {
+                setEmbedConfig(formatExistingConfig(data));
+                handleSuccess();
+            })
+            .catch(e => handleError(e));
+    };
+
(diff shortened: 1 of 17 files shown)


jwalgran:
Looking at this now.

src/app/src/components/EmbeddedMapConfigWrapper.jsx
```diff
@@ -0,0 +1,130 @@
...
+    // additional changes are made, update/create the embed config in the database.
+    useEffect(() => {
+        const handler = setTimeout(() => {
+            if (loading) return;
+            if (embedConfig.id) {
+                updateConfig();
+            } else {
+                createConfig();
+            }
+        }, 500);
```
jwalgran, line 93: Did we choose 500ms for any specific reason?

I am seeing "polling" like behavior where a PUT requests are being made repeatedly. I was expecting that PUT requests would only be made in response to a field changing.

src/app/src/util/constants.js
```diff
@@ -517,3 +517,5 @@ export const PPE_FIELD_NAMES = Object.freeze([
     'ppe_contact_email',
     'ppe_website',
 ]);
+
+export const OARFont = 'ff-tisa-sans-web-pro,sans-serif';
```
jwalgran, line 521: There are a few other components that reference this font string. Consider spreading the use of this constant to those component to ease the burden on a future maintainer that needs to change the font.

Screen Shot 2021-05-11 at 2 07 37 PM

src/app/src/util/embeddedMap.js
```diff
@@ -0,0 +1,100 @@
...
+            font: OARFont,
+            includeOtherContributorFields: false,
+            height: '800',
+        };
+    }
+    return {
+        ...embedConfig,
+        ...formatExistingWidth(embedConfig),
+        color: embedConfig.color.length ? embedConfig.color : '#3d2f8c',
+        font: embedConfig.font.length ? embedConfig.font : OARFont,
```
jwalgran, line 30: Why are we using `.length` on these fields rather than a basic check for a falsey value? Could we ever end up in an exceptional situation because `color` or `length` are null/undefined?

src/app/src/util/embeddedMap.js
```diff
@@ -0,0 +1,100 @@
...
+
+const formatEmbedFieldsForServer = fields =>
+    sortBy(fields, f => f.order).map((f, i) => ({
+        visible: f.visible,
+        order: i,
+        display_name: f.displayName,
+        column_name: f.columnName,
+    }));
+
+const formatWidthForServer = ({ width, fullWidth }) => {
```
jwalgran, line 85: When I check the 100% box I am seeing the width field being set to 100.

jwalgran (commented):
The persistence is working, but there are issues with the mechanics of how the updates happen and how errors are reported that I think we need to address.

src/app/src/components/EmbeddedMapConfigWrapper.jsx
```diff
@@ -0,0 +1,130 @@
...
+    };
+
+    // Fetches and merges nonstandard fields from the database with fields
+    // from the user's EmbedConfig
+    useEffect(() => fetchFields(), [user, fetchFields]);
+
+    /* eslint-disable react-hooks/exhaustive-deps */
+    // Disabled exhaustive dependencies warning because we don't want to rerun
+    // in response to the loading state changing
+    // When the embed cofig or fields are updated, wait for 500ms. If no
```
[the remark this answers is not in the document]
TaiWilkin: Thanks for the catch - updated in 9d913b8

src/app/src/components/EmbeddedMapConfigWrapper.jsx
```diff
@@ -0,0 +1,130 @@
...
+        formatExistingConfig(user.embed_config),
+    );
+    const [embedFields, setEmbedFields] = useState(
+        formatExistingFields(embedConfig.embed_fields),
+    );
+
+    const handleError = e => {
+        setError(getErrorMessage(e));
+        setLoading(false);
+        setTimeout(() => setError(null), 5000);
```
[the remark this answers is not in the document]
TaiWilkin: We're currently using `react-toastify` elsewhere in the app, so I've switched to that here - it definitely is an improvement in terms of workflow and appearance.

src/app/src/components/EmbeddedMapConfigWrapper.jsx
TaiWilkin: The polling behavior was caused because the update and create functions are being recreated on each render. Removing them from the dependency array restored expected behavior. The 500ms delay is meant to prevent repeated calls to the database while the user is actively typing, but the actual duration of the debounce is fairly random.

src/app/src/util/constants.js
TaiWilkin: Updated in 5726f0f.

src/app/src/util/embeddedMap.js
TaiWilkin: Updated in 9d913b8 and also moved some values to constants for clarity/DRYness. I was thinking that 0s would come out as 'false' but since the numbers are strings, that isn't actually an issue.

src/app/src/util/embeddedMap.js
TaiWilkin: Updated to make the behavior a little bit more logical in 9d913b8, but I did want to note as well that full-width embed isn't set up for the preview or embed code yet (#1311 will add that), so the required/expected behavior here will likely undergo additional changes when that is implemented.

TaiWilkin:
Ready for another look.

jwalgran (approved):
Changes look good. Thanks.

Issue #1218, opened 2021-01-28
Update facility details API


dboyer:
## Overview

The embedded map feature supports a contributor generating an iframe of a map showing just the facilities they have contributed as well as potentially additional fields they have added that are not visible on OAR.

For this task, add additional contributor fields to the facility details API. 

Part of #317 

### Describe the solution you'd like

Update facility details API

- [ ] Add embed and contributor arguments (`/api/facilities/{id}/?embed=1&contributor={id}`)
- [ ] Conditionally include extra fields, as defined by that contributor's EmbedConfig, when embed is 1, in the format: 
```{
    "id": "OAR_ID",
    "type": "Feature",
    "geometry": {
        "type": "Point",
        "coordinates": [1, 1]
    },
    "properties": {
       ...regular data,
       "contributor_fields": [
       {
            "label": "Test Field",
            "value": "1234"
        },
       ...
     ]
    }
}```

PR #1333, opened 2021-05-11
Update facility details API

TaiWilkin:
## Overview

The embedded map feature supports a contributor showing additional
fields they have added that are not visible on OAR. When a contributor
id and embed=1 are passed to the facility details API, an array of
additional fields, as defined by that contributor's embedded map
settings, are returned with the facility details.

Connects #1218 

## Demo

Screen Shot 2021-05-11 at 1 24 36 PM

## Notes

We previously discussed the format of the fields being: `{ "label": "Test Field", "value": "1234" }` This works well when the contributor has only submitted the facility once; however, in some cases contributors submit a facility multiple times, which can result in multiple available values for a field. 

Currently, instead, all of the available values for the field are returned as an array `{ "label": "Test Field", "value": ["1234", "abc"] }`. This seemed like the most honest presentation of the available data, but other options would be to only use the most recently submitted FacilityListItem for the facility (regardless of whether the submission includes all of the fields in question), or to use the most recently submitted FacilityListItem where the field is present for each field. 

## Testing Instructions

* Run `vagrant ssh` and `./scripts/test`. 
     - [x] All tests should pass.
* Sign in and submit a facility by API with custom fields. Ensure that it has been matched or created.
* Submit a facility via list. 
* Manually add the custom fields as EmbedFields for the account you're signed into, ensuring they are set to 'visible'. 
* `GET` the details for the facilities you submitted, passing embed=1 and your contributor id as query params. 
     `http://localhost:6543/api/facilities/{facility id}/?embed=1&contributor={id}`
     - [x] The contributor fields should be provided.
     - [x] The fields should show the display name and the provided value(s). 
     - [x] The fields should be in the order defined by the EmbedFields. 
   
## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/serializers.py
+++ b/src/django/api/serializers.py
@@ -1,4 +1,6 @@
 from itertools import chain
+import csv
+import json
 
 from django.conf import settings
 from django.core import exceptions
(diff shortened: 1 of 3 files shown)


src/django/api/serializers.py
```diff
@@ -615,6 +677,31 @@ def get_activity_reports(self, facility):
...
+        if contributor.embed_config is None:
+            return []
+
+        config = contributor.embed_config
+        fields = EmbedField.objects.filter(
+            embed_config=config, visible=True).order_by('order')
+
+        list_items = FacilityListItem.objects.filter(
+                facility=facility,
+                source__contributor=contributor)
```
jwalgran, line 701: Not captured in the original issue description is the need to allow contributors to dissociate from facilities they no longer do business with, which we allow for via the `is_active` flags. I expect that adding 2 additional filter statements to this call to `.filter()` should take care of this

```
source__is_active=True, 
facilitymatch__is_active=True
```

Although the same facility could be submitted multiple times and therefore have multiple line items matched to it, I think that we should assume that the most recent, active submission is the one from which we should pull the nonstandard field values. I expect that adding an `order_by('-created_at')` and then taking just the first matching row will take care of this.

src/django/api/tests.py
```diff
@@ -6494,3 +6495,143 @@ def test_querydict_in_api_data(self):
         # We plan to only store raw data as CSV or JSON in the future. Legacy
         # data can be ignored
         self.assertNotIn('extra_2', content)
+
+
+class ContributorFieldsApiTest(APITestCase):
```
jwalgran, line 6500: Love to see a new test case appear.

src/django/api/serializers.py
```diff
@@ -493,13 +495,73 @@ def format_source(source):
...
+                                    item, contributor_fields
+                                )
+        else:
+            contributor_fields = get_list_contributor_field_values(
+                                    item, contributor_fields
+                                 )
+
+    return [{
+                'values': f['values'], 'label': f['label']
+            } for f in contributor_fields]
```
jwalgran, line 554: Although the same facility could be submitted multiple times and therefore have multiple line items matched to it, I think that we should assume that the most recent, active submission is the one from which we should pull the nonstandard field values. I mentioned this at the point in the code where the QuerySet is built and, as a result, we should be able to take a single list item as a parameter to this function and use a single `'value': None` to initialize the return rather than `'values': []'`.

jwalgran (commented):
I was able to `curl` facilities with additional field values and splitting the data processing into a set of smaller helper functions helps with readability.

I made some suggestions around multiple submissions and dissociation that I think we want to handle but weren't fully described in the initial issue.

TaiWilkin:
Thanks for the review! Updated in 6b04a55 to use only the most recent, active item for the data. Ready for another look!

TaiWilkin:
Ready for another look.

jwalgran (approved):
I verified the changes. Looks great. 👍

Issue #1327, opened 2021-05-05
Nonstandard fields API appears to timeout on staging


jwalgran:
## Overview

Requests to https://staging.openapparel.org/api/nonstandard-fields/ return a 504

### Steps to Reproduce

* Log in to staging
* Browse https://staging.openapparel.org/api/nonstandard-fields/

### Demo

<img width="1250" alt="Screen Shot 2021-05-05 at 7 57 24 AM" src="https://user-images.githubusercontent.com/17363/117162484-a839b600-ad77-11eb-8752-d32370b6f3c3.png">

### Additional context

I suspect that our method of combing through the entire database looking for fields is too slow for production use. We may need to shift to finding these non-standard fields when data is submitted and  writing them to a table.

### To do
Performance boost the custom field list:
- [x] Add a table of custom field column names. Use `unique_together` instead of ID
     - `contributor_id` (Foreign Key)
     - `column_name`
- Write to field table when creating a facilityListItem 
     - [x] FacilityList
     - [x] POST facility API)
- Update the nonstandard fields API query:
     - [x] Query the NonstandardFields model
     - [x] Include four fixed fields (to be selectable in settings)
          - parent_company
          - type_of_product
          - number_of_workers
          - type_of_facility

PR #1340, opened 2021-05-17
Add Nonstandard Fields Model

TaiWilkin:
## Overview

Our previous method of looping through a contributor's facility list
items and parsing them for the contributor's nonstandard fields was
timing out on staging. This adds a NonstandardField model, which gets
created when a user creates a facility via API or creates a list,
allowing us to access the user's nonstandard fields more quickly.

Connects #1327 

## Demo

Screen Shot 2021-05-17 at 9 08 05 AM

## Testing Instructions

* Run `vagrant ssh` and `./scripts/update`
* Run`./scripts/test`
     - [x] The tests should pass
* Run `./scripts/server`
* Navigate to http://localhost:8081/api/nonstandard-fields/
     - [x] You should see the default fields: parent_company, type_of_product, number_of_workers, type_of_facility
* Login as `c2@example.com` and [submit a list](http://localhost:6543/contribute) with an extra column
* Get an API token from [settings](http://localhost:6543/settings)
* In [the Swagger API](http://localhost:8081/api/docs/#!/facilities/facilities_create), authorize and then submit a facility with a different extra field
* Navigate to the [nonstandard fields api](http://localhost:8081/api/nonstandard-fields/)
     - [x] You should see both extra fields listed, as well as the default fields

## Checklist

- [x] CI passes after rebase


src/django/api/models.py
```diff
@@ -2049,6 +2049,35 @@ class Meta:
...
+        return (
+            'Column Name {column_name} - Order: {order} ').format(
+            **self.__dict__)
+
+
+class NonstandardField(models.Model):
+    """
+    Nonstandard data fields available to include on facilities in an
+    embedded map
+    """
```
caseycesari, line 2061: Are non-standard fields different than contributor/custom fields?

TaiWilkin: Yes/No. Previously, we were grabbing nonstandard/extra fields that a contributor sent with facilities and, on the fly, getting a collection of them and returning them for use on the settings page, but that was too slow. Now we are creating an NonstandardField object whenever a contributor submits a facility, so that we can just query the NonstandardField objects for that contributor when we go to the settings page. 
Then, on the settings page, we allow the contributor to select which items from their collection of NonstandardFields they want to include on their facilities (creating EmbedField objects), and optionally choose alternative display names for them. 
Finally, we calculate and return the selected display names and the facility-specific values to the facility details API for presentation to users ('contributor_fields' in the facility details request). 
So I'm thinking of it as, we have nonstandard fields that are submitted, then we create embed configuration fields, and then we send the actual facility's contributor-submitted fields data... but it's possible I'm making a false distinction and/or other naming conventions might be clearer.

src/django/api/views.py
```diff
@@ -2179,6 +2182,27 @@ def get_link(self, path, method, base_url):
         return None
 
 
+def create_nonstandard_fields(fields, contributor):
+    unique_fields = list(set(fields))
+
+    existing_fields = NonstandardField.objects.filter(
+        contributor=contributor).values_list('column_name', flat=True)
+    new_fields = filter(lambda f: f not in existing_fields,
+                        unique_fields)
```
caseycesari, line 2191: Consider lowercasing all of the field names in the csv before comparing them to existing fields.

I accidentally edited the CSV template, and capitalized the `C` in `country`. After uploading a facility, `Country` now shows up as a non-standard field:

image

caseycesari (changes requested):
Generally working well. Found one issue.

src/django/api/views.py
TaiWilkin: I have added lowercasing the fields when comparing to the standard fields (country, address, etc), to prevent users from accidentally adding one of our standard fields with a casing mistake. 
My thoughts are that we want to leave the casing as submitted for actually saving the column_names, as we use the column_name to access user-submitted data. If the user has a facility with { "column_ABC": "this is data" }, in order to access that data we will need to use facility['column_ABC'] with the same casing. 
I am torn on whether to map over the list of existing fields and lowercase them when comparing, so that having 'column_ABC' will prevent a user from adding 'column_abc'. I am leaning towards not, because the addition of the alternatively cased field to their embed settings will potentially help users see that the field was added incorrectly, especially since they can choose to ignore/hide that field or alter its display field.

caseycesari: That approach makes sense to me.

caseycesari (approved):
Working well!

Issue #1258, opened 2021-02-24
Hide other contributor information in embed mode


jwalgran:
## Overview

Remove other contributor data on the facility details page when in embedded map view

### Describe the solution you'd like

- Hide other contributors in detail API response
     - Remove other locations/names/addresses from the API response that are not from the contributor
     - Remove (or filter?) contributors section from the API response (they may list data from other contributors)
- Remove “other” checkbox
- Add migration to remove "other" column from `EmbedConfig`

PR #1341, opened 2021-05-17
Show custom facility fields when in embed mode

caseycesari:
## Overview

When viewing the details of a facility in embed mode, render the custom fields in the details sidebar.

Only the custom fields that were added by the contributor to a facility are displayed. Custom fields added by another user to the same facility will not be displayed.

Connects #1334

## Demo

*The `test` field is not displayed because it has a null value*
image

Screen Shot 2021-05-17 at 4 03 28 PM

## Testing Instructions

- Log in as a contributor, and go to Settings > Embed
- View the example map, and find a facility that you have added custom fields to.
- View the facility details, and ensure the custom fields are shown in the sidebar. 
- Ensure any fields that have a `null` value are not displayed.
- Toggle some of the custom field checkboxes in the "Include these fields" config section, and ensure the fields that are unchecked do not appear in the sidebar.

## Checklist

- [x] `fixup!` commits have been squashed
- [x] CI passes after rebase
- [x] CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines

PR #1342, opened 2021-05-18
Remove "show other contributor info" embed option and hide other contributor information in embed mode

caseycesari:
## Overview

Add a migration to remove the "show other contributor info" option from the embed config model, and remove related code from the back-end. The front-end control for this option is also removed.

Hide other facility locations, names, addresses, and contributors on the facility details panel in embed mode.

Connects #1258 

## Demo

| Embed mode  | Main app |
| --- | --- |
| image | image |

## Testing Instructions

- Run `./scripts/update` to apply the migration
- Visit Settings > Embed
- Scroll down to the "Include these fields" section, and verify the "Information from other contributors" checkbox is gone.
- Scroll to the map, and view a facility that has multiple other locations/names/addresses/contributors. Ensure none of those fields are displayed.
- Visit the main app, and view the same facility. Ensure any other locations/names/addresses/contributors are displayed.

## Checklist

- [x] CI passes after rebase


src/django/api/serializers.py
```diff
@@ -55,6 +55,16 @@ def _get_parent_company(claim):
...
 
+def is_embed_mode_active(serializer):
+    request = serializer.context.get('request') \
+        if serializer.context is not None else []
+
+    if request is not None or request.query_params is not None:
+        if request.query_params.get('embed') == '1':
+            return True
+
+    return False
```
caseycesari, line 66: This function is based off of [this code](https://github.com/open-apparel-registry/open-apparel-registry/pull/1342/files#diff-9d16a5dca79134fade5a462b1028f0d3187305adcd0cc171e952baa8ebfce98eR705-R713). I'm not entirely sure of all the scenarios it is trying to protect against, but I thought the same conditions should be applied here.

TaiWilkin:
Taking a look now

TaiWilkin:
I've found a bug which I think was actually initially introduced in [#1341](https://github.com/open-apparel-registry/open-apparel-registry/pull/1341) that I missed previously. If you navigate directly to [a facility](http://localhost:6543/facilities/CN2021103JB0N0C?q=Nantong%20Jackbeanie&embed=1) with embed set, the app is freezing. I think this is because we're using the signed in user to fetch the facility details before the user information is set. Actually, we should be using a contributor query param and not the authenticated user's id, because a contributor's map will be accessed by other people. The URL for the embedded map will be in the format: "http://localhost:6543/?contributor=2&embed=1" (with the contributor id being whichever contributor 'owns' that map) - in this case, we would send '2' as the contributor id for the facility details.

src/app/src/reducers/FiltersReducer.js
```diff
@@ -84,10 +84,10 @@ export default createReducer(
             update(state, {
                 lists: { $set: payload },
             }),
-        [resetAllFilters]: (state, payload) =>
+        [resetAllFilters]: (state, isEmbedded) =>
             update(initialState, {
                 contributors: {
-                    $set: payload
+                    $set: isEmbedded
                         ? state.contributors
```
caseycesari, line 91: Makes the change that was suggested here: https://github.com/open-apparel-registry/open-apparel-registry/pull/1257/files#r581397227

TaiWilkin (approved):
This is working well. Thanks for resolving the contributors query param issue, despite my sometimes-conflicting information provided. Nice job debugging!

caseycesari:
Thanks for your help and review!

Issue #1052, opened 2020-07-16
The raw_data field on FacilityListItem is being populated a stringified QueryDict or dict instead of JSON


jwalgran:
## Overview

The raw_data field on FacilityListItem is being populated with either a QueryDict or a stringified Python dict instead of JSON.

Addressing this will require a fix to make sure all incoming data is saved as JSON, and after this fix is deployed, optionally cleaning up the incorrectly serialized legacy data with a migration

### Expected Behavior

No matter how data is POSTed to the facility endpoint the body of the request is saved in the `raw_data` field as JSON.

### Actual Behavior

The `raw_data` field on `FacilityListItem` is being populated with either a `QueryDict` or a stringified Python dict instead of JSON.

### Steps to Reproduce

* POST a valid JSON object to /api/facilities

### Demo

<img width="1198" alt="Screen Shot 2021-04-26 at 12 53 29 PM" src="https://user-images.githubusercontent.com/17363/116142160-6a110800-a68e-11eb-80d7-c72b6df1256a.png">

<img width="856" alt="Screen Shot 2020-07-16 at 11 15 59 AM" src="https://user-images.githubusercontent.com/17363/87707409-e8cb3300-c755-11ea-804b-58515d6d8f24.png">

PR #1343, opened 2021-05-19
Fix Raw Data Formatting

TaiWilkin:
## Overview

The raw data sent with a facility creation request was being stored in
a variety of formats, including QueryDicts and stringified
Python objects. It is now stored as JSON.

Connects #1052 

## Notes

We are not attempting to migrate legacy data due to current time constraints. Legacy data on production has minimal to no extra fields supplied by contributors, which is the main portion of the raw_data we are interested in accessing. 

## Testing Instructions

* Run `vagrant ssh` and `./scripts/test`
     - [x] All tests should pass
* Run `./scripts/server`
* Submit a facility via API 
* The returned facility information will include an item_id. Find that item in the [facility list items](http://localhost:8081/admin/api/facilitylistitem).
     - [x] The raw data should be formatted as JSON. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - Adjust settings so fast refresh works in development [#1284](https://github.com/open-apparel-registry/open-apparel-registry/pull/1284)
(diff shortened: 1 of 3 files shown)


caseycesari (approved):
Confirmed the raw data is valid JSON for a newly submitted facility.

PR #1352, opened 2021-05-25
Add full-width embedded map support

TaiWilkin:
## Overview

Adjusts the embedded map preview and sample code to allow users
to set the embedded map to full width, and implements sensible
defaults when moving between full width and set width.

Per a conversation with Scott, some small changes have been made to the
styling of the size section of the embedded map settings.

The width and height text inputs were not allowing users to completely clear them, which made it difficult to enter an entirely new size. There is now a validation message presented in response to empty text inputs instead. 

Connects #1311 

## Testing Instructions

* Run `./scripts/server`
* Login and navigate to [settings](http://localhost:6543/settings) under the embed tab
* Update the width and height of the map.
     - [x] The preview size should change appropriately. 
     - [x] The embed code, when copied, should create the same iframe as the preview.
* Set the width to full-width.
     - [x] The width input should be hidden.
     - [x] The checkbox label should say '100% Width' instead of '100%'
     - [x] The aspect ratio of the preview should be maintained. 
     - [x] The embed code, when copied, should create the same iframe as the preview.
* Uncheck the full-width checkbox.
     - [x] The width input should reappear with the width set to 1000.
     - [x] The aspect ratio should be maintained. 
* Delete the text in the width or height input. 
     - [x] There should be an error message, but you should be able to clear the input. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/util/embeddedMap.js
+++ b/src/app/src/util/embeddedMap.js
@@ -2,16 +2,15 @@ import sortBy from 'lodash/sortBy';
 
 import { OARFont } from './constants';
 
-const DEFAULT_WIDTH = '1000';
-const DEFAULT_HEIGHT = '800';
+export const DEFAULT_WIDTH = '1000';
+export const DEFAULT_HEIGHT = '800';
 
-const getHeight = embedConfig =>
-    embedConfig.height ? embedConfig.height : DEFAULT_HEIGHT;
+const getHeight = embedConfig => (embedConfig.height ? embedConfig.height : '');
 
 const formatExistingWidth = ({ width = DEFAULT_WIDTH }) => {
     const fullWidth = width[width.length - 1] === '%';
     if (!width) {
-        return { fullWidth, width: DEFAULT_WIDTH };
+        return { fullWidth, width: '' };
     }
     return { fullWidth, width };
 };
@@ -84,7 +83,7 @@ const formatEmbedFieldsForServer = fields =>
 
 const formatWidthForServer = ({ width, fullWidth }) => {
     if (fullWidth) return '100%';
-    if (!width || width === '100%') return DEFAULT_WIDTH;
+    if (!width || width === '100%') return '';
     return width;
 };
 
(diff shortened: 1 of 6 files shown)


src/app/src/util/util.js
```diff
@@ -780,9 +780,26 @@ export const getEmbeddedMapSrc = ({ contributor, timestamp }) => {
...
+        src="${getEmbeddedMapSrc({
+            contributor,
+        })}"
+        frameborder="0"
+        allowfullscreen
+        style="position:absolute;top:0;
+               left:0;width:100%;height:100%;"
+        title="embedded-map"></iframe>
+  </div>
+</div>`
```
caseycesari, line 797: Consider changing the indentation over so that this block lines up with the block below.

src/app/src/util/embeddedMap.js
```diff
@@ -2,16 +2,15 @@ import sortBy from 'lodash/sortBy';
 
 import { OARFont } from './constants';
 
-const DEFAULT_WIDTH = '1000';
-const DEFAULT_HEIGHT = '800';
+export const DEFAULT_WIDTH = '1000';
+export const DEFAULT_HEIGHT = '800';
```
caseycesari, line 6: This may be for a follow-up card, but do we want to set a minimum as well? The map is unusable at small sizes.

Screen Shot 2021-05-26 at 11 43 35 AM

caseycesari (approved):
Couple small things, but this is working well.

src/app/src/util/util.js
TaiWilkin: Good call, thanks

src/app/src/util/embeddedMap.js
TaiWilkin: @caseycesari My thought here is that, since users will have direct visual feedback that it looks terrible if they set it at a small size, we can let them self-police their sizing.

TaiWilkin:
Thanks for the review!

Issue #1356, opened 2021-05-26
DRAFT: Avoid flashing language button while loading embedded map


jwalgran:
# Draft

This is a draft issue, and it will need to clarify specs and formalize what
counts as "done" before work begins on it.

Check with the issue's creator before starting work on this issue.

## Overview

Avoid flashing language button while loading embedded map

### Additional Context

![2021-05-26 08 35 57](https://user-images.githubusercontent.com/17363/119689241-8921c780-bdfd-11eb-997c-3c43241329e1.gif)

PR #1357, opened 2021-05-26
Hide translation element on embedded map

caseycesari:
## Overview

Since the app bar is hidden in embed mode, the translation element should also be hidden. Otherwise it is attached to the page on load, but is hidden once the map fully loads, which results in a visual blip.

Connects #1356

## Demo

May-26-2021 15-41-34

## Testing Instructions

- Visit an embedded map. Verify that the "Select Language" is not displayed before the map is loaded.

## Checklist

- [x] CI passes after rebase


jwalgran:
Looking at this now.

jwalgran (approved):
Confirmed 👍

Issue #1362, opened 2021-05-27
DRAFT: Ensure the transparency pledge fields are visibile for lower pricing tiers


jwalgran:
# Draft

This is a draft issue, and it will need to clarify specs and formalize what
counts as "done" before work begins on it.

Check with the issue's creator before starting work on this issue.

## Overview

Ensure the transparency pledge fields are visible for lower pricing tiers

### Describe the solution you'd like

TODO: Decide on adding the custom field UI at lower levels or showing them by default

PR #1365, opened 2021-05-27
Enable the default embed map custom fields by default

jwalgran:
## Overview

We only show the custom field configuration UI to embedded map users at higher price tiers but the default transparency pledge fields are available to any embedded map user. To keep the configuration simple, we modify the client side code to enable these fields default by default and give them nicely formatted labels.

Connects #1362

## Notes

We add a migration to allow for the embed config field on Contributor to be blank since without this change you are not able to save a Contributor record in the Django admin unless an embed config has been created.

To ensure that the list of non-standard field names is unique we combine it with the default field list using sets rather than lists.

## Demo

Screen Shot 2021-05-27 at 12 58 20 PM

## Testing Instructions

* Run migrations
* Log in as an admin and modify the Contributor to have the "Embed" level of access
* Log in as a user/contributor without an embed configuration. I used c4@example.com.
* Open the developer tools network tab
* Browse http://localhost:6543/settings, click the embed tab, and configure the embed. Verify that POST/PUT requests are made with the default transparency pledge fields enabled and labeled.
* Upload 
[test-transparency-pledge-fields.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/6556469/test-transparency-pledge-fields.csv) and fully process it
  * `./scripts/manage batch_process --list-id {id} --action parse`
  * `./scripts/manage batch_process --list-id {id} --action geocode`
  * `./scripts/manage batch_process --list-id {id} --action match`

* Use serve the markup below on localhost and browse it. Verify that the "Some Software Co." facility lists all 4 transparency pledge fields

```
<!doctype html>

<html lang="en">
    <head>
        <meta charset="UTF-8"/>
        <title>OAR Embed Demo</title>
    </head>
    <body>
        <h1>
            <div>
                My embedded map
            </div>
        </h1>
        <li>item one</li>
        <li>item b</li>
        <li>item the third</li>


        <!-- embed -->
        <iframe
            src="http://localhost:6543/?contributors=4&embed=1"
            frameBorder="0"
            style="width:1000px;height:750px"
            title="embedded-map"
        />
        <!-- end -->

    </body>
</html>
```

## Checklist

- [x] CI passes after rebase


TaiWilkin:
This is working as described, but I have one concern with this implementation. Currently, if a user without an existing embed config logs in at a higher embed level and navigates to the settings page, it will display the default fields as visible; however, they haven't actually been enabled until the user interacts with the page and initiates a POST request.

jwalgran:
@TaiWilkin I pushed 67dea9e which adds some server side handling for returning default fields when there is no saved config to override them. Ready for another look.

src/django/api/models.py
```diff
@@ -2070,6 +2070,15 @@ class NonstandardField(models.Model):
     class Meta:
         unique_together = ('contributor', 'column_name')
 
+    # Keys in this set must be kept in sync with
+    # defaultNonstandarFieldLabels in app/src/app/util/embededMap.js
```
TaiWilkin, line 2074: Typo - missing a 'd' in Nonstandard and in embedded

TaiWilkin (approved):
This is working very well! Nice fix-up. 
I noted a couple of small typos, but otherwise this is good to go.

jwalgran:
Thanks for the review and the excellent suggestion. The revised implementation is much better.

PR #1375, opened 2021-06-02
Update embedded map help text

jwalgran:
## Overview

Update embedded map help text provided by the OAR team.


jwalgran:
Merging to expedite a release.

Issue #1371, opened 2021-06-02
Embed lacks default size


lederer:
## Overview

When the embed feature is enabled for a contributor, the size fields on the embed settings page are empty and the embed doesn't have a default width and height. As a result, the preview is awkwardly tiny and looks broken.

### Expected Behavior

Open the settings > embed page for the first time, the size fields are pre-filled with reasonable default values (say, 640x480) and the embed preview dimensions reflect those values.

### Actual Behavior

Open the settings > embed page for the first time, the size fields are empty and the preview is awkwardly tiny.

### Demo

![image](https://user-images.githubusercontent.com/128699/120486893-7837ff80-c383-11eb-9c75-c9ea954de044.png)

PR #1376, opened 2021-06-02
Show messages when the embed config is incomplete

jwalgran:
## Overview

The embed preview and code do not render properly unless a size is selected. To handle that we show informational messages rather than content until the minimum required configuration has been entered.

Connects #1371

## Demo

Screen Shot 2021-06-02 at 1 23 28 PM

## Testing Instructions

* Enable embed for a contributor who has never used it
* Browse config and verify that content is hidden until color, width, and height are selected.

## Checklist

- [x] CI passes after rebase


jwalgran:
I am going to merge this early to expedite deployment, but I would like a brief post-merge review from @lederer and @TaiWilkin.

lederer:
Overall good. Thanks for doing that. 

One wrinkle: When I select a color, two instances of "Error: This field may not be blan." appear above the size fields. With the preview and code messaging that this PR introduces, I think it would be best to just eliminate those error messages entirely.

jwalgran:
Thanks, @lederer. I removed the now redundant error messages in 54d3b94 per your suggestion.

TaiWilkin (commented):
This looks good! I confirmed it's working as expected on `develop`. Hiding the code as well as the preview is a nice touch.

Issue #1346, opened 2021-05-21
Additional report request


katieashaw:
## Overview

We'd like to request some additional monthly reports

### Is your feature request related to a problem? Please describe.

We need additional reports to enable us to track against our external and internal KPIs

### Describe the solution you'd like

Could we please have the following reports:

- [x] Percent of facilities claimed
- [ ] Platform % Uptime (Is there a way we can see this?)
- [x] Duplicates merged
- [x] Facilities marked as closed
- [x] Names and addresses promoted ~(quarterly)~ monthly
  - Count history rows where `created_from` changed from one non-null value to another.
- [ ] Seconds per submitted item to geocode (the time from when an item was submitted to when it was processed (including time waiting in a queue))
  - Monthly and weekly versions
- [ ] Seconds per submitted item to process a match (the actual processing time, excluding any wait time)
  - Monthly and weekly versions
- [x] Number of API users (cumulative)
  - Distinct users in the request log
- [x] Number of Embedded map users
  - Cumulative create_date on the embdconfg row

cc @hlennett

dboyer:
@hlennett Could you put these in a most important to least important priority order so we add them in an order that's most helpful to you?

hlennett:
Yes! Here you go @dboyer: 

- Percent of facilities claimed
- Platform % Uptime (Is there a way we can see this?)
- Duplicates merged
- Facilities marked as closed
- Names and addresses promoted (quarterly) monthly
- Seconds per submitted item to geocode (the time from when an item was submitted to when it was processed (including time waiting in a queue))
- Seconds per submitted item to process a match (the actual processing time, excluding any wait time)
- Number of API users (cumulative)
- Number of Embedded map users

jwalgran:
I rearranged the checklist in the issue description to match the priority order provided by Hannah.

PR #1391, opened 2021-06-10
Add new admin reports

TaiWilkin:
## Overview

Adds new administrative reports for:
- Percent of facilities claimed (monthly)
- Duplicates merged per month
- Facilities marked as closed per month
- Names and addresses promoted per month

Some reports have proven difficult or inefficient to create using pure SQL, so the names and addresses promoted per month report introduces a new method of running reports which uses a custom Python function on a per-report basis. 

Connects #1346 

## Demo

Screen Shot 2021-06-08 at 11 37 39 AM
Screen Shot 2021-06-10 at 9 23 30 AM
Screen Shot 2021-06-10 at 9 24 15 AM
Screen Shot 2021-06-10 at 8 35 07 AM

## Testing Instructions

* Run `./scripts/server`
* Claim a facility
* Merge two facilities
* Mark a facility as closed
* Promote a facility address
* Confirm that each report's logic appears sound 
* Confirm that the facility actions you have taken show in the reports
* Confirm that the downloaded CSV files are correct

## Checklist

- [x] **DROP** any commits that change Jenkinsfile
- [x] CI passes after rebase

### DIFF
--- a/src/django/api/reports.py
+++ b/src/django/api/reports.py
@@ -4,15 +4,37 @@ from glob import glob
 from urllib.parse import quote
 
 from django.db import connection
-
+from api.models import HistoricalFacility
 
 _root = os.path.abspath(os.path.dirname(__file__))
+
+
+def monthly_promoted_name_and_address():
+    data = dict()
+    historical_facilities = HistoricalFacility.objects \
+        .exclude(created_from_id__isnull=True) \
+        .order_by('id', 'history_date') \
+        .values_list('id', 'history_date', 'created_from_id')
+    for i, h in enumerate(historical_facilities):
+            if i != 0:
+                prev = historical_facilities[i - 1]
+                if prev[0] == h[0] and h[2] != prev[2]:
+                    month = h[1].strftime('%Y-%m')
+                    data[month] = data.get(month, 0) + 1
+    return [['month', 'facilities promoted'],
+            sorted(data.items(), key=lambda x: x[0])]
+
+
+NON_SQL_REPORTS = {
+    'monthly_promoted_name_and_address': monthly_promoted_name_and_address
+}
 
 
 def get_report_names():
     file_paths = glob(os.path.join(_root, 'reports', '*.sql'))
     files = [os.path.basename(f) for f in file_paths]
     names = [os.path.splitext(f)[0].replace('_', '-') for f in files]
+    names += NON_SQL_REPORTS.keys()
     names.sort()
     return names
 
@@ -27,6 +49,31 @@ def quote_val(val):
 
 
 def run_report(name):
+    if name in NON_SQL_REPORTS.keys():
+        return run_non_sql_report(name)
+    else:
+        return run_sql_report(name)
+
+
+def create_csv_data(columns, rows):
+    csv_rows = [','.join([quote_val(c) for c in r]) for r in rows]
+    csv_lines = [','.join([quote_val(c) for c in columns])] + csv_rows
+    return quote('\n'.join(csv_lines))
+
+
+def run_non_sql_report(name):
+    columns, rows = NON_SQL_REPORTS[name]()
+    csv_data = create_csv_data(columns, rows)
+    return {
+        'name': name,
+        'sql': 'N/A: Custom Report',
+        'columns': columns,
+        'rows': rows,
+        'csv_data': csv_data,
+    }
+
+
+def run_sql_report(name):
     report_filename = os.path.join(
         _root, 'reports', '{}.sql'.format(name.replace('-', '_')))
     with open(report_filename, 'r') as report_file:
@@ -35,9 +82,8 @@ def run_report(name):
             cursor.execute(report_sql)
             columns = [col[0] for col in cursor.description]
             rows = cursor.fetchall()
-            csv_rows = [','.join([quote_val(c) for c in r]) for r in rows]
-            csv_lines = [','.join([quote_val(c) for c in columns])] + csv_rows
-            csv_data = quote('\n'.join(csv_lines))
+            csv_data = create_csv_data(columns, rows)
+
             return {
                 'name': name,
                 'sql': report_sql,
--- a/src/django/api/reports/duplicates_merged.sql
+++ b/src/django/api/reports/duplicates_merged.sql
@@ -1,0 +1,5 @@
+SELECT to_char(a.created_at, 'YYYY-MM') as month, COUNT(*) as duplicates_merged
+FROM api_facilityalias a
+WHERE reason = 'MERGE'
+GROUP BY month
+ORDER BY month;
--- a/src/django/api/reports/facilities_marked_closed.sql
+++ b/src/django/api/reports/facilities_marked_closed.sql
@@ -1,0 +1,5 @@
+SELECT to_char(approved_at, 'YYYY-MM') as month, COUNT(*) as facilities_closed_in_month
+FROM api_HistoricalFacilityActivityReport
+WHERE status = 'CONFIRMED' AND closure_state = 'CLOSED'
+GROUP BY month
+ORDER BY month;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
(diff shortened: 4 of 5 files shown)


src/django/api/reports/duplicates_merged.sql
```diff
@@ -0,0 +1,4 @@
+SELECT to_char(a.created_at, 'YYYY-MM') as month, COUNT(*) as duplicates_merged
+FROM api_facilityalias a
+WHERE reason = 'MERGE'
+GROUP BY month;
```
jwalgran, line 4: We should include `ORDER BY month` so this is sorted like our other reports.

src/django/api/reports/facilities_marked_closed.sql
```diff
@@ -0,0 +1,4 @@
+SELECT to_char(approved_at, 'YYYY-MM') as month, COUNT(*) as facilities_closed_in_month
+FROM api_HistoricalFacilityActivityReport
+WHERE status = 'CONFIRMED'
+GROUP BY month;
```
jwalgran, line 4: We should include `ORDER BY month` so this is sorted like our other reports.

src/django/api/reports/facilities_marked_closed.sql
```diff
@@ -0,0 +1,4 @@
+SELECT to_char(approved_at, 'YYYY-MM') as month, COUNT(*) as facilities_closed_in_month
+FROM api_HistoricalFacilityActivityReport
+WHERE status = 'CONFIRMED'
```
jwalgran, line 3: I think we also want to have `AND closure_state = 'CLOSED'` here so that we do not count a confirmed reopening report.

src/django/api/reports.py
```diff
@@ -4,15 +4,35 @@
...
 
 _root = os.path.abspath(os.path.dirname(__file__))
 
 
+def monthly_promoted_name_and_address():
+    data = dict()
+    fs = Facility.objects.all()
+    for f in fs:
+        hs = f.history.exclude(created_from_id__isnull=True)
+        for h in hs:
```
jwalgran, line 17: Since we only have one reference to each of these QuerySet variables, consider moving them into the `for` definitions.

```
    for f in Facility.objects.all():
        for h in f.history.exclude(created_from_id__isnull=True):
```

src/django/api/reports.py
```diff
@@ -4,15 +4,35 @@
...
+            prev = h.prev_record
+            if prev is not None and prev.created_from_id is not None \
+                    and h.created_from_id is not prev.created_from_id:
```
jwalgran, line 20: I think we should prefer `!=` over `is not` when compare numeric values. I recall a previous usage of `is not` with numbers resulting in a bug (https://realpython.com/python-is-identity-vs-equality/).

jwalgran (commented):
Nice refactoring to allow for Python function-based reports. I ran the large SQL query in production and the results look correct to me. I made some suggestions for changes to the smaller reports and the Python-based one.

When the changes are made I would suggest also adding a commit to temporarily add PR-1391 to the list of deployable branches in the `Jenkinsfile` so that we can test with a more realistic data set.

src/django/api/reports/duplicates_merged.sql
TaiWilkin: Updated in  44a9a1f

src/django/api/reports/facilities_marked_closed.sql
TaiWilkin: Updated in 44a9a1f

src/django/api/reports/facilities_marked_closed.sql
TaiWilkin: Updated in 44a9a1f

src/django/api/reports.py
TaiWilkin: Updated in 44a9a1f

src/django/api/reports.py
TaiWilkin: Thanks! I need to keep that difference in mind. 
Updated in 44a9a1f

TaiWilkin:
@jwalgran It seems like the regular SQL reports are working well on staging, but the [non-SQL report](https://staging.openapparel.org/admin/reports/monthly_promoted_name_and_address/) is returning a 504 error. Do you have any thoughts on how to proceed?

jwalgran:
I reverted the "Test on staging" commit to prevent deployments while I test a release.

jwalgran:
>It seems like the regular SQL reports are working well on staging, but the non-SQL report is returning a 504 error. Do you have any thoughts on how to proceed?

As an optimization, we could experiment with running a single `values_list` query against the `FacilityHistory` table that sorts by facility ID, then `history_date`. This should avoid multiple trips to the database and avoids instantiating model instances for each row returned. If that query is fast enough and we can iterate over it fast enough, we may be able to produce a result before the request times out.

jwalgran (commented):
Fixups look good.

Some encouragement that a single query / values_list solution may work for our Python report. I wrote up this QuerySet to fetch the data we need

```
    HistoricalFacility.objects \
                      .exclude(created_from_id__isnull=True) \
                      .order_by('id', 'history_date') \
                      .values_list('id', 'history_date', 'created_from_id')
```

Then used `.query` to get the SQL

```
SELECT "api_historicalfacility"."id", "api_historicalfacility"."history_date", "api_historicalfacility"."created_from_id" 
FROM "api_historicalfacility" 
WHERE NOT ("api_historicalfacility"."created_from_id" IS NULL) 
ORDER BY "api_historicalfacility"."id" ASC, "api_historicalfacility"."history_date" ASC
```

Then ran the query with an `EXPLAIN ANALYZE` prefix on production.

```
                                                            QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=7826.71..7992.77 rows=66425 width=28) (actual time=472.848..497.029 rows=66433 loops=1)
   Sort Key: id, history_date
   Sort Method: external merge  Disk: 2480kB
   ->  Seq Scan on api_historicalfacility  (cost=0.00..2506.25 rows=66425 width=28) (actual time=1.104..96.617 rows=66433 loops=1)
         Filter: (created_from_id IS NOT NULL)
 Planning Time: 3.245 ms
 Execution Time: 500.046 ms
(7 rows)
```

This is with 66k history rows

```
openapparelregistry=> select count(*) from api_historicalfacility;
 count
-------
 66433
(1 row)
```

Since we can query the data within 500ms, and the rows are narrow, I am optimistic that even when we get to double the facility volume than we have now we will still be able to query and transfer it in less than 2 seconds and that we can loop over it, aggregate it, and return the result well within our 20 second timeout.

TaiWilkin:
@jwalgran I've updated the query using `values_list` as you suggested, with significant time improvements (in development at least). Running it previously locally took ~2-4 seconds, and now takes ~0.00-0.02 seconds, which is in line with our other reports.

src/django/api/reports.py
```diff
@@ -4,15 +4,37 @@
...
+        .order_by('id', 'history_date') \
+        .values_list('id', 'history_date', 'created_from_id')
+    for i, h in enumerate(historical_facilities):
+            if i != 0:
+                prev = historical_facilities[i - 1]
+                if prev[0] == h[0] and h[2] != prev[2]:
+                    month = h[1].strftime('%Y-%m')
+                    data[month] = data.get(month, 0) + 1
+    return [['month', 'facilities promoted'],
+            sorted(data.items(), key=lambda x: x[0], reverse=True)]
```
jwalgran, line 25: I'm glad this optimization worked out and the logic is basically the same as it was.

I am seeing the result sorted in descending order. Our other reports are in ascending order, with the newest row at the bottom, so we should consider changing this report to match the rest.

Screen Shot 2021-06-15 at 10 21 45 AM

jwalgran (approved):
Looks good. Thanks for working through the revisions with me. I made one sorting suggestion and added a checklist item to ensure that we drop changes to `Jenkinsfile` before merging.

src/django/api/reports.py
TaiWilkin: Thanks! I've updated the sort order.

TaiWilkin:
Jenkinsfile commits have been dropped and the sort order has been updated. Thanks for all of your work and troubleshooting of the new Python report with me!

Issue #1297, opened 2021-04-01
Maximum Update Depth Exceeded bug


TaiWilkin:
Going to the facilities list and clicking on a facility causes a `Maximum update depth exceeded` error in the developer tools, and the page freezes. It doesn't occur when going to the facility directly. This error is present on Staging but not Production.

caseycesari:
It doesn't happen every time, but I've run into this when visiting a facility page directly.

jwalgran:
I added `console.log` statements inside each of the the `useEffect` handlers in our `useUpdateLeafletMapImperatively` hook. After clicking a facility in the results list I saw the last `useEffect`, responsible for panning the map to the selected location, stuck in a loop. Commenting out https://github.com/open-apparel-registry/open-apparel-registry/blob/b220433dfc476ac147d96ef3e6c26a3bc3ec187a/src/app/src/util/hooks.js#L123-L162 (and adding `/* eslint-disable no-unused-vars */
`) broke the map panning, but the update depth error and unresponsiveness when clicking Claim this facility (#1392) were resolved.

I will now look deeper into why this `useEffect` is not behaving the way we intend it to.

PR #1393, opened 2021-06-10
Remove state flag from useUpdateLeafletMapImperatively to avoid loop

### LINKED ISSUE #1392: Browser is unresponsive when clicking the Claim this Facility link on the facility details page
## Overview

Browser is unresponsive when clicking the Claim this Facility link on the facility details page

### Steps to Reproduce

* Log in
* Click the "Facilities" tab
* Select a facility
* Click "Claim this Facility"

jwalgran:
## Overview

When navigating from the facility search results list to the facility details view the final `useEffect` block in `useUpdateLeafletMapImperatively` was triggering a render loop. Removing the state flag used within that `useEffect` block eliminated the rendering loop without breaking any desired map pan and zoom functionality.

Connects #1297 
Connects #1392

## Notes

I checked that the behaviors described in #799 are not affected.

- If the user clicks on a facility marker, do not pan & zoom to the
facility
- If the user arrives at the map directly from a URL containing an OAR
ID, set the map view on the facility location, zoomed to level 15
- If the user clicks on a facility in the sidebar and the selected
facility is outside of the bbox OR the zoom level is
within the range in which the facilities grid layer is shown, pan to the
facility and set the zoom level at 15. This ensures that the facility
details marker will be visible even if the map was at grid-layer zoom
levels before

While testing that this also fixes #1392 we saw that there was an error logged to the console and shown in the WDS overlay error viewer when navigating from the map page to a page without the map. It looks like this may be the result of some react-leaflet code trying to operate on the map related DOM elements that are no longer on the page. 

Screen Shot 2021-06-10 at 1 49 18 PM

This error does not appear to affect functionality

## Testing Instructions

* Open the console, browse http://localhost:6543/, click the Facilities tab, select a facility, and verify that no error is logged.
* Use the in-app back button to return to the facility list, select another facility, and verify that it also works.
* Log in as c2@example.com
* On the facility details page, click the "Claim this Facility" link and verify that the claim page loads (you may run into a non-blocking error when running in development. see note above. this error overlay can be dismissed using the X in the upper right.
 
## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Remove state flag from useUpdateLeafletMapImperatively to avoid loop [#1393](https://github.com/open-apparel-registry/open-apparel-registry/pull/1393)
+
 ### Security
 
 ## [42] - 2021-06-02
--- a/src/app/src/util/hooks.js
+++ b/src/app/src/util/hooks.js
@@ -111,55 +111,34 @@ export default function useUpdateLeafletMapImperatively(
         }
     }, [data, shouldSetViewOnReceivingData, setShouldSetViewOnReceivingData]);
 
-    // Set the map view to center on the facility at zoom level 15 if
-    // the user has clicked on a facility list item in the sidebar for a
-    // facility that is currently off-screen or when the vector tile map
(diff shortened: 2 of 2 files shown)


TaiWilkin (approved):
Great fix! I tested the zoom and pan functionality as well as the listed tests, and the map appears to be functioning as before.

jwalgran:
Thanks for the review.

PR #1399, opened 2021-06-16
Add Additional New Admin Reports

TaiWilkin:
## Overview

- Weekly and monthly averages of the number of seconds from
geocoding start to finish (Python report)
- Weekly and monthly averages of the number of seconds from source
creation to geocoding finish (Python report)
- Monthly number of distinct API users in the request log (cumulative) (SQL report)
- Monthly number of embedded map configs created (cumulative) (SQL report)

Connects #1346 

## Demo

Screen Shot 2021-06-16 at 11 14 24 AM
Screen Shot 2021-06-16 at 11 14 34 AM
Screen Shot 2021-06-16 at 11 16 38 AM
Screen Shot 2021-06-16 at 11 16 46 AM
Screen Shot 2021-06-16 at 11 16 57 AM
Screen Shot 2021-06-16 at 11 17 04 AM

## Testing Instructions

* Run `./scripts/server`
* Navigate to [reports](http://localhost:8081/admin/reports/)
* Confirm that [API Users Monthly](http://localhost:8081/admin/reports/api-users-monthly) works as expected
* Confirm that [Embedded Map Users Monthly](http://localhost:8081/admin/reports/embedded-map-users-monthly) works as expected
* Confirm that [Monthly Geocoding Time with Queue](http://localhost:8081/admin/reports/monthly_geocoding_time_with_queue) works as expected
* Confirm that [Monthly Geocoding Time without Queue](http://localhost:8081/admin/reports/monthly_geocoding_time_without_queue) works as expected
* Confirm that [Weekly Geocoding Time with Queue](http://localhost:8081/admin/reports/weekly_geocoding_time_with_queue) works as expected
* Confirm that [Weekly Geocoding Time without Queue](http://localhost:8081/admin/reports/weekly_geocoding_time_without_queue) works as expected

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/reports.py
+++ b/src/django/api/reports.py
@@ -1,12 +1,28 @@
 import os
+import pytz
 
+from datetime import datetime
 from glob import glob
 from urllib.parse import quote
 
 from django.db import connection
-from api.models import HistoricalFacility
+from api.models import HistoricalFacility, FacilityListItem
+from api.constants import ProcessingAction, DateFormats
 
 _root = os.path.abspath(os.path.dirname(__file__))
+
+
+def sort_by_first_column(array):
+    return sorted(array, key=lambda x: x[0])
+
+
+def try_parsing_date(text):
+    for fmt in (DateFormats.STANDARD, DateFormats.SECOND):
+        try:
+            return datetime.strptime(text, fmt)
+        except ValueError:
+            pass
+    raise ValueError('no valid date format found')
 
 
 def monthly_promoted_name_and_address():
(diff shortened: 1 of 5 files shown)


src/django/api/reports.py
```diff
@@ -22,11 +29,80 @@ def monthly_promoted_name_and_address():
...
     return [['month', 'facilities promoted'],
-            sorted(data.items(), key=lambda x: x[0])]
+            sort_by_first_column(data.items())]
+
+
+def geocoding_time_without_queue(date_format):
+    temp_data = dict()
+    listitems = FacilityListItem.objects.values_list(
+        'created_at', 'processing_results')
+    for f in listitems:
```
jwalgran, line 39: To avoid using subscripts, I think that we can use tuple destructuring here to get some more readable variable names.

```python
for (created_at, processing_results) in listitems:
```

src/django/api/reports.py
```diff
@@ -22,11 +29,80 @@ def monthly_promoted_name_and_address():
...
+        for r in f[1]:
+            if r['action'] == ProcessingAction.GEOCODE:
+                started = datetime.strptime(r['started_at'],
+                                            '%Y-%m-%d %H:%M:%S.%f')
```
jwalgran, line 43: We use the `%Y-%m-%d %H:%M:%S.%f` a few times across reports so consider extracting this as a constant.

src/django/api/reports.py
```diff
@@ -1,14 +1,21 @@
...
 from django.db import connection
-from api.models import HistoricalFacility
+from api.models import HistoricalFacility, FacilityListItem
+from api.constants import ProcessingAction
 
 _root = os.path.abspath(os.path.dirname(__file__))
 
 
+def sort_by_first_column(array):
+    return sorted(array, key=lambda x: x[0])
```
jwalgran, line 16: Nice helper 👍

src/django/api/reports.py
```diff
@@ -22,11 +29,80 @@ def monthly_promoted_name_and_address():
...
+
+
+def weekly_geocoding_time_without_queue():
+    rows = geocoding_time_without_queue('%Y-%W')
+    return [['week', 'average_geocoding_time_in_seconds'], rows]
+
+
+def geocoding_time_with_queue(date_format):
+    temp_data = dict()
+    listitems = FacilityListItem.objects.values_list('source__created_at',
```
jwalgran, line 69: Are we using `source__created_at` rather than `created_at` on the list item itself for a specific reason? They are set at nearly the same time, so I would expect the difference between them to be negligible and using `created_at` would avoid a `JOIN`. 

https://github.com/open-apparel-registry/open-apparel-registry/blob/564bca4adb7f3766e8dfefbbaf2a29293f883f5a/src/django/api/views.py#L2373-L2392

src/django/api/reports.py
```diff
@@ -22,11 +29,80 @@ def monthly_promoted_name_and_address():
...
+                                                     'processing_results')
+    for f in listitems:
```
jwalgran, line 71: To avoid using subscripts, I think that we can use tuple destructuring here to get some more readable variable names.

```python
for (created_at, processing_results) in listitems:
```

src/django/api/reports.py
```diff
@@ -22,11 +29,80 @@ def monthly_promoted_name_and_address():
...
+                temp_data[date] = (counts[0] + 1,
+                                   counts[1] + time.total_seconds())
+    data = dict()
+    for d, c in temp_data.items():
+        data[d] = c[1] / c[0]
+    return sort_by_first_column(data.items())
+
+
+def monthly_geocoding_time_with_queue():
+    rows = geocoding_time_with_queue('%Y-%m')
```
jwalgran, line 90: This is a nice use of parameterization to get the same function to aggregate different time groupings 👍

jwalgran (approved):
I verified these run without error and the SQL reports run in production with reasonable results. I made a few small inline comments.

I would like to run the Python reports on staging to test their output but since this is an admin-only feature I am OK with merging and then dealing with any issues we uncover with a follow-up PR.

src/django/api/reports.py
TaiWilkin: I grabbed the other date formats from this file as well, since it seems likely we'll keep using each of them.

src/django/api/reports.py
TaiWilkin: I wasn't sure how significant the time difference would be. Since it's minor, I've moved to just using 'created_at'.

TaiWilkin:
@jwalgran I added a commit to test on Staging prior to merging, just to be on the safe side.

jwalgran (commented):
I tested the geocode time reports on staging and they hit the 504 timeout.

When discussing this with the OAR team there was agreement that having a separate reporting deployment is a good idea long term.

In the short term, they said:
- There is not time urgency for getting these geocode time reports running
- Exact precision was not a priority, so we have the option of only processing a subset of the rows to reduce the amount of input data we have to chew through.

TaiWilkin:
I've pushed a fixup that includes only the most recent 1000 facility items for the geocoding reports. This (temporary fix) will hopefully allow the OAR team to get a feel for the current time of geocoding items, even if it's not the full view of the historical geocoding times, and hopefully also fix the timeout issue.

Issue #1400, opened 2021-06-16
The height of embeds when the width is 100% are too short to be usable


jwalgran:
![2021-06-16 11 18 37](https://user-images.githubusercontent.com/17363/122271783-a72a9700-ce94-11eb-8b0f-a21929675329.gif)

PR #1401, opened 2021-06-16
Set a minimum height when using a 100% width embed

jwalgran:
## Overview

When setting the embed height as a proportion of the page width, the narrow screen widths of mobile devices lead to an embed that is too sort to be functional.

To work around this we introduce additional styling of the embed that uses media query to override the percentage with a fixed, minimum height.

Connects #1400

## Testing Instructions

I used this index.html page served wtih https://www.npmjs.com/package/http-server to test the embed codes generated by this PR

```
<!doctype html>

<html lang="en">
    <head>
        <meta charset="UTF-8"/>
        <meta name="viewport" content="width=device-width">
        <title>OAR Embed Demo</title>
    </head>
    <body>
        <h1>
            <div>
                My embedded map
            </div>
        </h1>
        <li>item one</li>
        <li>item b</li>
        <li>item the third</li>



        <!-- embed -->
        <!-- end -->

    </body>
</html>
```
* Test that the embed code generated for both fixed and full width embeds work and the full width embed works at phone width.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - Correct contributor types [#1397](https://github.com/open-apparel-registry/open-apparel-registry/pull/1397)
+- Set a minimum height when using a 100% width embed [#1401](https://github.com/open-apparel-registry/open-apparel-registry/pull/1401)
 
 ### Security
 
--- a/src/app/src/components/EmbeddedMapConfig.jsx
+++ b/src/app/src/components/EmbeddedMapConfig.jsx
@@ -5,7 +5,10 @@ import { func, shape, string, bool } from 'prop-types';
 
 import { userPropType } from '../util/propTypes';
 import { getEmbeddedMapSrc } from '../util/util';
-import { EmbeddedMapInfoLink } from '../util/constants';
+import {
+    EmbeddedMapInfoLink,
+    minimum100PercentWidthEmbedHeight,
+} from '../util/constants';
 
 import AppGrid from './AppGrid';
 import EmbeddedMapFieldsConfig from './EmbeddedMapFieldsConfig';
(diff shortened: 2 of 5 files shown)


TaiWilkin (changes requested):
The changes to the embed code are working well on various size screens and with various height and width settings. However, the embedded map preview hasn't been updated to match and should be kept consistent.

lederer:
Note this should be added to the `<head>` of the test html file to help simulate mobile layout:

```
<meta name="viewport" content="width=device-width">
```

lederer:
Nice catch @TaiWilkin.

I think we can get away with a more relaxed minimum, so we're not usurping too much of their customers' webpages. `500px` seems okay in my casual testing?

Even still, one wrinkle with our solution is that we run the risk of mooting the customer-specified relative height at _desktop_ widths. Eg, if someone sets `100%` width and `50%` height (fairly short aspect ratio but not unreasonable), then even if we lower our hardcoded minimum to `500px`, the embed height will remain at `500px` all the way up to viewport widths of `1000px`, after which it becomes dynamic/relative.

We could address this through a media query, where we force a static height only for viewports below a certain width (`600px` would match our current mobile breakpoint), and use the customer's relative height otherwise.

```html
<div>
    <style>
        #oar-embed-GUID { /* append some RANDOM GUID? */
            position: relative;
            padding-top: 50%; /* customer-specified relative height */
        }
        @media (max-width: 600px) { /* mobile breakpoint */
            #oar-embed-GUID {
                padding-top: 500px; /* our mobile minumum */
            }
        }
    </style>
    <div id="oar-embed-GUID"> <!-- id for CSS -->
        <iframe
            src="http://localhost:6543/?contributors=6&embed=1"
            frameborder="0"
            allowfullscreen
            style="position:absolute;top:0;left:0;width:100%;height:100%;"
            title="embedded-map">
        </iframe>
    </div>
</div>
```

jwalgran:
I implemented Scott's suggested media query-based solution, reduced the min height to 500px, and fixed the preview. Thanks for the help. This is ready for another review.

TaiWilkin (approved):
This is working very well. Nice incorporation of changes, and the additional code comments will be helpful. I was able to simulate a number of height and width settings successfully.

lederer (approved):
👍

jwalgran:
Thanks, again, for helping me revise this.

lederer:
Thanks for jumping in and fixing this!

Issue #1404, opened 2021-06-17
Create cumulative versions of reports


TaiWilkin:
Create cumulative versions of two reports: 
[Industry Data vs OAR Data](https://openapparel.org/admin/reports/industry-data-vs-oar-data/)
[Percent Data by Contributor Type](https://openapparel.org/admin/reports/percent-data-by-contributor-type/)

PR #1405, opened 2021-06-17
Add cumulative versions of reports

TaiWilkin:
## Overview

Adds cumulative versions of the Industry Data vs. OAR Data report
and the Percent Data by Contributor Type report. In addition, adds
columns for the counts used to calculate the percents, for added
clarity. 

Connects #1404 

## Demo

Screen Shot 2021-06-17 at 3 56 11 PM
Screen Shot 2021-06-17 at 3 56 26 PM

Screen Shot 2021-06-17 at 3 56 34 PM
Screen Shot 2021-06-17 at 3 58 20 PM

## Testing Instructions

* Run `./scripts/server`
* In order to confirm the reports work locally, you will need to have matches submitted by a user with an openapparel.org email and also matches which were created in different months. I replicated this by changing the email of a user with already-submitted matches and by postdating a large number of matches using `./scripts/manage dbshell`.
* Visit [percent data by contributor type](http://localhost:8081/admin/reports/percent-data-by-contributor-type/) and note the count and total for a contributor type across different months. 
* Visit [percent data by contributor type cumulative](http://localhost:8081/admin/reports/percent-data-by-contributor-type-cumulative/) and confirm if you add the counts and totals for the contributor type it adds up to the cumulative amounts given. 
* Repeat with [industry data vs oar data](http://localhost:8081/admin/reports/industry-data-vs-oar-data/) and [industry data vs oar data cumulative](http://localhost:8081/admin/reports/industry-data-vs-oar-data-cumulative/)

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add new admin reports [#1391](https://github.com/open-apparel-registry/open-apparel-registry/pull/1391)
 - Reorder contributor fields [#1398](https://github.com/open-apparel-registry/open-apparel-registry/pull/1398)
 - Add additional new admin reports [#1399](https://github.com/open-apparel-registry/open-apparel-registry/pull/1399)
+- Add cumulative versions of reports [#1405](https://github.com/open-apparel-registry/open-apparel-registry/pull/1405)
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


jwalgran (approved):
I tested these on production and the results look reasonable. I was initially concerned that the total counts were larger than the number of facilities, bit then realized that when multiple contributors are associated with a facility there can be multiple contributor types or industry/oar associations  for a facility and it can therefor be counted multiple times.

Issue #1402, opened 2021-06-17
The other locations section of facility details does not properly anonymize contributor names


jwalgran:
## Overview

The `FacilityDetailsSerializer` does not check the permissions of the user in `get_other_locations` and  always serializes the full name of the contributor.

If the user is in the `can_submit_private_facility` group but is not in the `can_view_full_contrib_detail` group (a "hybrid" user) then the contributor names in the `other_locations` section should be anonymized.

PR #1415, opened 2021-07-06
Anonymize other_locations contributors

TaiWilkin:
## Overview

If the user is in the `can_submit_private_facility` group but is not in
the `can_view_full_contrib_detail` group (a "hybrid" user) then the
contributor names in the `other_locations` section of the
FacilityDetails are anonymized.

Connects #1402 

## Demo

Screen Shot 2021-07-06 at 9 10 30 AM
Screen Shot 2021-07-06 at 9 12 43 AM

## Testing Instructions

- Login as c1@example.com
- Navigate to http://localhost:8081/admin/api/user/3/change/ and move ‘can submit private facility’ to chosen groups and save
- Get a facility id and go to http://localhost:6543/dashboard/updatefacilitylocation and update the location. Set a contributor for the location. 
- Logout and navigate to http://localhost:6543/ and view the facility. You should see the contributor name under ‘other locations’
- Login as c2@example.com. This user has full view permissions. View the facility’s details. You should still see the contributor name. 
- Grab the token for c2@example.com and authenticate with the API. 
- Navigate to http://localhost:8081/api/docs/#!/facilities/facilities_read and fetch the facility’s details. You should see the contributor name. 
- Log out and log in as c3@example.com and view the facility’s details at http://localhost:6543/. You should see anonymized contributor data under ‘other locations’. 
- Generate a new API token and authenticate with the API. 
- Navigate to http://localhost:8081/api/docs/#!/facilities/facilities_read and fetch the facility’s details. You should see anonymized contributor details. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/serializers.py
+++ b/src/django/api/serializers.py
@@ -44,6 +44,7 @@ from api.models import (FacilityList,
                         NonstandardField)
 from api.countries import COUNTRY_NAMES, COUNTRY_CHOICES
 from api.processing import get_country_code
+from api.helpers import prefix_a_an
 from waffle import switch_is_active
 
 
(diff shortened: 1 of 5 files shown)


src/django/api/serializers.py
```diff
@@ -634,9 +657,9 @@ def get_other_locations(self, facility):
                 'contributor_id':
                 l.facility_list_item.source.contributor.admin.id
```
jwalgran, line 658: While testing I discovered that we have an additional requirement that was not captured in the original issue.

The name of the contributor is being correctly anonymized, but we are still creating a clickable link to the profile page where the full contributor details are visible. I think this is easily taken care of by adding a `user_can_see_detail` check here and making a complimentart change to the client code that gracefully handles a null `contributor_id`

NOTE this screencast has a map-related soft crash that is not related to the anonymization issue.

TaiWilkin:
Updated to remove the contributor_id (and links) from the other_locations data.

jwalgran (commented):
I confirmed the fixup. Nice implementation.

jwalgran (approved):

TaiWilkin:
Thanks for reviewing!

Issue #1411, opened 2021-06-30
In Account Drop down, don't indent "Log Out" 


hlennett:
## Overview

When logged into an OAR account, when you click on the account menu in the upper right corner, Log Out is now in line with the rest of the options. We'd like them to line up. 

### Is your feature request related to a problem? Please describe.

It's just visually nicer. 

### Describe the solution you'd like

We'd like all of the items in that drop down to be aligned. 

### Additional Context

![Screen Shot 2021-06-30 at 12 52 28 PM](https://user-images.githubusercontent.com/80268596/124001011-2b574100-d9a2-11eb-8b35-3b20582180aa.png)

cc @katieashaw

PR #1420, opened 2021-07-12
Align logout button

TaiWilkin:
## Overview

The button for the navigation dropdown for logged-in users was out of
line with the links, but is now aligned. Logout, unlike the other items
in the navigation, is a button instead of an anchor, because it doesn't
navigate the user but rather performs an action in response to a click.
This requires it to have different styles, which need to be set using
higher specifity CSS classes in order to override the '!important' tags
on other CSS classes for nav dropdowns.

Connects #1411 

## Demo

Screen Shot 2021-07-12 at 12 26 44 PM

## Testing Instructions

* Run `./scripts/server`, login as any user, and click the user name in the upper right-hand corner to view the log out button. Confirm it is aligned properly at various screen sizes. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - Conditionally hide remove button [#1417](https://github.com/open-apparel-registry/open-apparel-registry/pull/1417)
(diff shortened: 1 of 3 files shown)


jwalgran (approved):
Looks good 👍 
Screen Shot 2021-07-12 at 1 39 05 PM

TaiWilkin:
Thanks for reviewing!

Issue #1310, opened 2021-04-12
Facility history API endpoint raising an unhandled exception


jwalgran:
## Overview

The facility history API endpoint is raising an unhandled exception

### Steps to Reproduce

* Use the token of a production user who has permission to use the history API to make a request to get the history of any facility

### Demo

https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/738/

![history-api-exception](https://user-images.githubusercontent.com/17363/114472629-7170e580-9ba7-11eb-95dd-5fd716333368.png)

PR #1421, opened 2021-07-13
Fix facility history bug

TaiWilkin:
## Overview

There are cases where a facility claim history object references
a facility which has been deleted. When fetching the facility history,
a ObjectDoesNotExist exception was being thrown. We now catch this
exception and exclude the broken objects.

Connects #1310 

## Demo

Test showing error prior to updating `facility_history.py`:
Screen Shot 2021-07-13 at 10 03 44 AM

## Notes

In most cases we would use `facility is not None` instead of
catching an exception. In this situation, most likely because of a
`history` related quirk, that doesn't resolve the error.

## Testing Instructions

* Run `./scripts/test` and confirm all tests pass
* Run `./scripts/server` and confirm that the history API is working as before

## Checklist

- [ ] CI passes after rebase


jwalgran (approved):
Clear, Pythonic implementation. I copied the test into the `develop` branch and verified that it failed there. 👍

TaiWilkin:
Thanks for reviewing!

PR #1424, opened 2021-07-14
Handle error when splitting match w/o location

caseycesari:
## Overview

Show an informational error message when a match can not be split out into its own facility because it does not have a geocoded location.

Connects to https://github.com/open-apparel-registry/open-apparel-registry-clients/issues/38

## Demo

Screen Shot 2021-07-14 at 12 13 59 PM

## Testing Instructions

- Create a new facility via the API or CSV upload
- Create a second facility with a similar name and the same address as the first facility. Ensure it is matched to the first facility.
- Visit the admin, view the list of `FacilityListItems`, and note the id of the matched facility.
- Open the shell (`./scripts/manage shell_plus`).
- Find the matched item `foo = FacilityListItem.objects.get(pk=XXX)`
- Set the location to `None` `foo.geocoded_point = None`
- Save the item `foo.save()`
- Go to the Adjust Facility Matches dashboard page.
- Find the Facility ID of the first facility you created and enter it on the dashboard page.
- Attempt to remove the matched item. You should see an error message about the match not having a location.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/views.py
+++ b/src/django/api/views.py
@@ -1819,6 +1819,10 @@ class FacilitiesViewSet(mixins.ListModelMixin,
 
             list_item_for_match = match_for_new_facility.facility_list_item
 
+            if list_item_for_match.geocoded_point is None:
+                raise ValidationError('The match can not be split because '
+                                      'it does not have a location.')
+
             facility_qs = Facility.objects.filter(
                 created_from=list_item_for_match)
             if facility_qs.exists():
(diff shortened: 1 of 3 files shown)


TaiWilkin (approved):
Looks good. Nice work on testing this out. I'm seeing a small error in that the message is coming back with a missing space between two words ("The match can not be split becauseit does not have a location."), but other than that this should be good to go.

src/django/api/views.py
```diff
@@ -1819,6 +1819,10 @@ def split(self, request, pk=None):
 
             list_item_for_match = match_for_new_facility.facility_list_item
 
+            if list_item_for_match.geocoded_point is None:
+                raise ValidationError('The match can not be split because'
```
TaiWilkin, line 1823: Missing a space here

Issue #1425, opened 2021-07-14
check_api_limits fails on staging


jwalgran:
## Overview

The scheduled run of `check_api_limits` on staging is failing due to a `.get()` ORM call resulting in a `DoesNotExist` exception

https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/742

### Demo

```
 check_api_limits(datetime.datetime.now(tz=timezone.utc))
File "/usr/local/src/api/limits.py", line 101, in check_api_limits
 check_contributor_api_limit(at_datetime, c)
File "/usr/local/lib/python3.7/contextlib.py", line 74, in inner
 return func(*args, **kwds)
File "/usr/local/src/api/limits.py", line 36, in check_contributor_api_limit
 contributor = Contributor.objects.get(id=c.get('contributor'))
File "/usr/local/lib/python3.7/site-packages/django/db/models/manager.py", line 82, in manager_method
 return getattr(self.get_queryset(), name)(*args, **kwargs)
File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 408, in get
 self.model._meta.object_name
api.models.DoesNotExist: Contributor matching query does not exist.
```

PR #1426, opened 2021-07-15
Block users without contributors

TaiWilkin:
## Overview

The `check_api_limits` management command was throwing an error when
attempting to process a user with no contributor (an invalid data
state). Because ApiBlocks are set on contributors, and not on individual
 users, we can't block a user with no contributor via ApiLimit under our
 current system. Instead, we now prevent these users from
 accessing the API entirely and return a clear error response.

Connects #1425 

## Demo

Screen Shot 2021-07-15 at 2 58 46 PM

## Testing Instructions

* Run `./scripts/manage shell_plus` and delete a contributor. 
```
c = Contributor.objects.get(id=6)
c.delete()
```
* Run `./scripts/manage check_api_limits` and confirm it runs without error
* Run `./scripts/server` and login as the user whose contributor you deleted. Get the user's token and authenticate with the API, then send a request. You should see an error message. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/tests.py
+++ b/src/django/api/tests.py
@@ -350,7 +350,7 @@ class FacilityListCreateTest(APITestCase):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(len(response.json()), 1)
 
-    def test_upload_by_user_with_no_contributor_returns_400(self):
+    def test_upload_by_user_with_no_contributor_returns_402(self):
         Contributor.objects.all().delete()
         token = Token.objects.create(user=self.user)
         self.client.post('/user-logout/')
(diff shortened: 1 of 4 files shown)


src/django/api/tests.py
```diff
@@ -350,7 +350,7 @@ def test_get_with_authentication_token_returns_items(self):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(len(response.json()), 1)
 
-    def test_upload_by_user_with_no_contributor_returns_400(self):
+    def test_upload_by_user_with_no_contributor_returns_402(self):
```
TaiWilkin, line 353: We now send a 402 in response to requests for access by users with no contributor.

jwalgran:
Looking at this now.

jwalgran (approved):
Confirmed. Nice, clear implementation with good inline documentation.

Issue #1433, opened 2021-07-21
DRAFT Experiement  search index and evaluate


jwalgran:
## Overview

Create a "search index table in the database and measure the tile query performance when using this table

### Is your feature request related to a problem? Please describe.

Contributor filters and a lot of map scrolling can lock up the application

### Describe the solution you'd like

- Create an "index table" 
  - Either repeating rows for facility+type+contributor+listid combinations or one row per facility with array fields. Is one more performant than the other.
- Add a function to index a facility
  - In a transaction, delete the existing row for a facility, get the full details, write index rows
- Call the indexing function any time a facility changes
  - Any match-related activity
- Update tile query logic to use the index table

PR #1455, opened 2021-08-18
Add facility indexing

TaiWilkin:
## Overview

Adds an indexed table of facilities, incorporating fields which are
utilized by the facilities search. Data such as facility contributors,
which require several relationship traversals to find, is denormalized
into array fields. A GinIndex is used to index the array fields.

Adds a management command to index facilities. To index all facilities,
pass no arguments to the management command. To index one or more
facilities, pass the ids of the facilities to be indexed.

Adds post_save signals to Facility, Source, FacilityMatch, and Contributor to 
index / re-index facilities after related entities have been updated or created. 

Updates the facilities search to utilize the new search index. In a test on staging, 
the new indexed search took an average of ~0.298 seconds. In comparison, 
the previous search took an average of ~0.617 seconds. 

Connects #1433 

## Demo

Queries run on Amazon Athena to analyze the time taken to search:

BEFORE
`SELECT AVG(target_processing_time) as Average, COUNT(target_processing_time) as Count FROM stg_alb_logs WHERE time > '2021-08-17' AND time < '2021-08-18T14:00:24.062714Z' AND request_url LIKE '%tile%';`
Average: 0.6175616475722862
Count: 3666

AFTER
`SELECT AVG(target_processing_time) as Average, COUNT(target_processing_time) as Count FROM stg_alb_logs WHERE time > '2021-08-18T14:30:24.062714Z' AND request_url LIKE '%tile%';`
Average: 0.2980567260940032
Count: 617

## Notes

- We use ArrayFields to store the related data for the facilities. In order to index these, we use a GinIndex. More info: [Django Docs](https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/fields/), [PostgreSQL Docs](https://www.postgresql.org/docs/current/gin-intro.html)
- In order to keep the search changes localized, the search creates a list of facility ids and then returns all of the Facilities with those ids. Testing showed that altering the search to return and the tile search to use the FacilityIndexes directly (avoiding the Facility mapping) was slower than using the Facility mapping. 
- Running the queries produced by the search against `explain analyze` in the dbshell showed that the FacilityIndex search wasn't making use of the GinIndex because on the current dataset it would be slower than use a sequential scan. However, when we turned off sequential scanning functionality in the database, the GinIndex was used successfully (but was in fact slightly slower). It seems likely that with a larger volume of FacilityIndexes the GinIndex will become the faster option and be used in the search as expected.
- The indexes are updating based on post_save signals. Using alternative update methods such as bulk update might in some cases not initiate the re-indexing of facilities.

## Testing Instructions

* Run `./scripts/manage migrate`. The migration should run correctly.
* Run `./scripts/manage index_facilities`.
* Run the search with various combinations of filters. Confirm that the search behaves as expected, and that the tiles and facilities returned are correct.  
* Update contributors, facility details, sources, and facility matches to various states using the Django admin. Confirm that the search reflects the changes to the data. 

## Checklist

- [x] CI passes after rebase


src/django/api/models.py
```diff
@@ -2112,3 +2140,64 @@ def __str__(self):
...
+
+    update_data = data.filter(id__in=existing_indexes)
+    update_fields = ['name', 'country_code', 'location', 'contrib_types',
+                     'contributors', 'ppe', 'lists']
+    FacilityIndex.objects.bulk_update(
+        [FacilityIndex(**kv) for kv in update_data], update_fields)
+
+    create_data = data.exclude(id__in=existing_indexes)
+    FacilityIndex.objects.bulk_create(
+        [FacilityIndex(**kv) for kv in create_data])
```
jwalgran, line 2197: I have about 11k facilities in my database. The first time I ran `./scripts/manage index_facilities` was about 8 seconds. Nearly all of the facilities in my database belong to the same contributor. When I updated the contributor record via the Django admin, the application locked up. I restarted tried running `./scripts/manage index_facilities` again and it ran for over 15 minutes before I forced  it to exit. With some `print` debugging I determined that it was getting stuck on the `bulk_update`. Switching to a delete followed by an insert resolved the issue and I think should have the same effect.

```suggestion
    FacilityIndex.objects.filter(id__in=facility_ids).delete()
    FacilityIndex.objects.bulk_create(
        [FacilityIndex(**kv) for kv in data])
```

src/django/api/models.py
```diff
@@ -2112,3 +2140,64 @@ def __str__(self):
...
+                                    filter=filter),
+                                  lists=ArrayAgg(
+                                    list,
+                                    filter=filter),
+                                  id=F('facility_id'),
+                                  ppe=Concat('facility__ppe_product_types',
+                                             'facility__ppe_contact_phone',
+                                             'facility__ppe_contact_email',
+                                             'facility__ppe_website',
+                                             output_field=CharField()))
```
jwalgran, line 2183: This is a very clever use of ORM features to handle both finding data and formatting it for insert in one pass. Well done. 👍

jwalgran (commented):
Nice work on this implementation. You made some excellent design choices that I may not have considered. I hit a significant performance issue when updating 5-figures worth of records and suggested a simple fix that appears to work well.

src/django/api/models.py
```diff
@@ -1533,6 +1498,64 @@ def activity_reports(self):
...
+    contrib_types = postgres.ArrayField(models.CharField(
+        max_length=200,
+        null=True,
+        blank=False,
+        choices=Contributor.CONTRIB_TYPE_CHOICES,
+        help_text='The categories to which the contributors belong.'))
+    contributors = postgres.ArrayField(models.CharField(
+        max_length=32,
+        null=True,
+        help_text='The contributor who submitted the facility data.'))
```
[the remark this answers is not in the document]
TaiWilkin: Thanks for the help! Updated in 0384630

src/django/api/models.py
```diff
@@ -2112,3 +2140,64 @@ def __str__(self):
         return (
             'Contributor ID: {contributor_id} - ' +
             'Column Name: {column_name}').format(**self.__dict__)
+
+
+@transaction.atomic
+def index_facilities(facility_ids=[]):
```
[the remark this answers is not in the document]
TaiWilkin: Fixed this up - thanks for the information, I wasn't aware that it would get shared between invocations!

src/django/api/models.py
TaiWilkin: Great suggestion! It's both faster and shorter. I've updated to this in 0384630

jwalgran:
Looking at this now.

src/django/api/models.py
```diff
@@ -2143,7 +2141,7 @@ def __str__(self):
 
 
 @transaction.atomic
-def index_facilities(facility_ids=[]):
+def index_facilities(facility_ids=list()):
```
jwalgran, line 2144: I wasn't as clear as I could have been my last comment about providing a default list argument. If we call `list()` here I think we end up in the same problem as using a literal list. A single list is created when Python interprets the function and it shared from then on. The trick to get around this, which is not intuitive, is to use a "callable" as the default value.

```python
def index_facilities(facility_ids=list):
```

jwalgran (approved):
Nice work. I found one small issue with the fixup but am going to approve this since it just involves removing some `()` 👍

src/django/api/models.py
TaiWilkin: Got it, thank you!

TaiWilkin:
Thank you for reviewing and for pairing on this!

Issue #1459, opened 2021-08-19
Allow matches to be moved to a different facility in the dashboard


isaacokunato:
## Overview

Team would like a manual move feature that would help move wrongly matched facilities into new locations

### Is your feature request related to a problem? Please describe.

Some of the wrongly matched facilities are unable to split out due to the fact that they are unable to plot or geocode a location. Therefore moving the wrong matches manually helps sort the problem

### Describe the solution you'd like

- Add a "Move" button on the dashboard page where matches are split. Hide split of there is no geocoded location.
- Clicking "Move" prompts for another facility ID, reuse facility search components from "merge"
- Show the confirmation dialog
- Create a new API endpoint that borrows and modifies logic from split
	- Updates FacilityMatch to point to the new facility
		- Update simple history appropriately
	- Adds to list item processing_results
- Make updates to facility_history.py to handle the new "Move" moderation action
	- Add a test case to the exiting history test class

@jwalgran @hlennett @katieashaw

isaacokunato:
Some of the facilities that need to be split can be found [here](https://docs.google.com/spreadsheets/d/1NRsmCETKOK193J5gjhpJgFVpkRXPFO0OwFFtMnEBA6o/edit#gid=1677014950). There are two sheets (Facilities that need new location) and (Facilities to be matched with others)

Link to the previously spit bug logged issue [here](https://github.com/open-apparel-registry/open-apparel-registry-clients/issues/38)

katieashaw:
Noting here that this work needs to take place ahead of an Embedded Map customer going live with their Embedded Map, to enable the Stakeholder Management team to moderate some of their data ahead of that launch. The launch is 15th September, so the work needs to be completed by 13th September 

cc @jwalgran @dboyer

TaiWilkin:
Additional data after a conversation with Scott:
After the user clicks "Move", a modal will open displaying a search bar. After entering an ID, a facility will appear below the search bar in the modal (but the search bar will remain). Users can then click 'confirm' to complete the move, or 'cancel' or 'x' to close the modal and cancel the move. 
Scott also recommends reconsidering the button text for the matches section to be more descriptive in general, eg "Create a new facility from match", "Move match to a different facility", etc.

PR #1464, opened 2021-09-09
Allow users to transfer matches between facilities

TaiWilkin:
## Overview

In some cases, list items have been incorrectly matched to a facility.
Currently, users will 'split' those list items away from the incorrect
facility. As a second step, they sometimes merge the match to a
different facility.

However, when the list items are not able to be geocoded, it's
impossible to use the 'split' functionality to create a new facility
from the match.

This adds the functionality to move or transfer a list item directly
from one existing facility to another existing facility, avoiding the
need to create a new, intermediate facility.

Connects #1459 

## Demo

Screen Shot 2021-09-09 at 2 41 58 PM
Screen Shot 2021-09-09 at 12 08 31 PM
Screen Shot 2021-09-09 at 12 08 43 PM
Screen Shot 2021-09-09 at 12 13 55 PM

## Notes

Scott recommended using more descriptive / different titles for the buttons to clarify their purpose. I think to do this thoroughly will require a follow-up card to determine the most appropriate titles and do any refactoring / styling needed to improve the display. 

Part of the facility details card was pulled out into a separate component. Initially I hope to reuse the entire component, but the card is actually the basis of the full page, and its functionality is very tied to the data flow for the facility whose matches are being adjusted. Separating the main UI of the card allowed me to reuse just that portion in the transfer modal with a different facility from the rest of the page. 

## Testing Instructions

* Create a match by uploading two similar facility list items on separate lists. 
* Note the id of the facility (the source facility). Get the id of a different facility (the transfer facility) to which you will move the match. 
* Login as an admin and navigate to the [adjust facility match dashboard](http://localhost:6543/dashboard/adjustfacilitymatches).
* Search for the source facility by id. It should display in the card as before. 
* The match you created should appear in the right hand. If it was successfully geocoded, it should show the 'split' button in addition to the 'promote' and 'transfer' buttons. 
* Click 'transfer'. A modal should open. 
* Search for the transfer facility by id and it should render in the modal. 
* Select 'cancel' and the modal should close. The transfer should not occur. 
* Reopen the modal and search again. Press 'transfer'. The modal should close and the match should be transferred. 
* Search for the transfer facility in the adjust facilities dashboard and you should see the match attached to it. 
* Use the history API to get the history of the source facility. The match move should be listed. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/views.py
+++ b/src/django/api/views.py
@@ -699,6 +699,9 @@ class FacilitiesAutoSchema(AutoSchema):
         if 'split' in path:
             return None
 
+        if 'move' in path:
+            return None
+
         if 'promote' in path:
             return None
 
@@ -1793,6 +1796,8 @@ class FacilitiesViewSet(mixins.ListModelMixin,
                         m.facility_list_item.source.contributor.id
                         if m.facility_list_item.source.contributor else None,
                         'match_id': m.id,
+                        'is_geocoded':
+                        m.facility_list_item.geocoded_point is not None,
                         'facility_created_by_item':
                         Facility.objects.filter(
                             created_from=m.facility_list_item.id)[0].id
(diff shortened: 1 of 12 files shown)


lederer (approved):
UI for the dialog looks good.

The 3 buttons in the originating "Matches to adjust" panel are rather cramped, so I'm happy to see there'll be a follow-up card to work on that part of the UI.

jwalgran (commented):
Clear, easy to follow implementation. My test worked and I confirmed that in addition to the new `MOVE` event in the facility history of the secondary facility, the history of the target facility also includes an `ASSOCIATE` event for the new match. Nice work.

I did find was appears to be a bug in the new dialog box. Each keystoke into the text box sends an API request. We should fix this before merging.

TaiWilkin:
@jwalgran I had to rebase in the process of fixing a weird build error that was occurring; apologies for muddying the water for changes. 
In addition to the bug for the input, I found a second error where some matches with locations were not showing the split button. I'll comment inline to clarify where the change was made to correct this.

src/django/api/views.py
```diff
@@ -1793,6 +1796,8 @@ def split(self, request, pk=None):
                         m.facility_list_item.source.contributor.id
                         if m.facility_list_item.source.contributor else None,
                         'match_id': m.id,
+                        'is_geocoded':
```
TaiWilkin, line 1799: It seems that the geocoded location wasn't being passed through the API properly. Because we do not actually need to know the location on the front end, just whether it exists, I've updated to send a boolean instead of the geocoded point.

jwalgran (commented):
Retest successful. 👍

jwalgran (approved):

TaiWilkin:
Thank you for reviewing!

Issue #1468, opened 2021-09-17
Way to show that GPS coordinates were changed but not completely accurate


hlennett:
## Overview

We have been getting reports that facilities are not plotting correctly (see https://github.com/open-apparel-registry/open-apparel-registry-clients/issues/45). As a work-around (until we build a more sophisticated plotting tool), we are looking at using a tool like ESRI or plugging in the coordinates for the specific province or town of the facility in Google Maps, so we know it is plotting in at least a more accurate way than that Google Maps is plotting, even though it's not perfect. However, in these situations, we want to be explicit about the fact that the GPS coordinates are known to not be 100% accurate and we'd like better coordinates reported. We'd love the ability to add a line of text under the GPS coordinates for facilities when we've made an imperfect update to the coordinates in scenarios like this. 

### Is your feature request related to a problem? Please describe.

We are going to be entering imperfect (but better) GPS coordinates for facilities where Google Maps plotting is way off. We want to be explicit about that with our users.

### Describe the solution you'd like

There's likely a back and front-end need here. On the back-end, when we update the GPS coordinates, we may need to check a box or something to indicate that the GPS coordinates entered will need to trigger helper text saying that the coordinates displayed are known to be inaccurate. Then, when that box is checked, text is displayed under the GPS coordinates on that facility profile that read:

"Unable to locate exact GPS coordinates for this facility. If you have access to accurate coordinates for this facility, please report them using the "Suggest a data edit" link below." 

### Additional Context

@katieashaw @nataliegrillon feel free to weigh in with any additional context or updates to the suggested copy. cc @isaacokunato

PR #1488, opened 2021-10-12
Add warning about inexact facility coordinates

TaiWilkin:
## Overview

Sometimes facilities are geolocated incorrectly or are unable to be
geolocated. In that situation, admins will manually update the location;
however, we need to alert the users that the facility coordinates are
known to be inaccurate. In the Django admin, users are now also able to
note that the facility is known to be inaccurate. When coordinates have
been labeled as inexact, we add text to the facility details noting that
the location is known to be inexact.

Connects #1468 

## Demo

Screen Shot 2021-10-07 at 1 23 37 PM

## Testing Instructions

* Run `./scripts/manage migrate` and then `./scripts/server`
* Navigate to a facility in the Django admin, and check the box 'has inexact coordinates' and save.
* Navigate to the facility details page; under the GPS coordinates, there should be a message warning that the coordinates are inexact. Text should match the text in the issue. 
* View a different facility's details page, and confirm that it does not include the warning.

## Checklist

- [x] CI passes after rebase


jwalgran (approved):
Confirmed 👍

Issue #1512, opened 2021-10-29
Correct Header/Footer links for map site


hlennett:
## Overview

Some of the header/footer links on the map site are linking to pages on the old info site. We've set up redirects so nothing appears to be broken, but would prefer to get the correct links in place:

**Header:** 
Home: https://info.openapparel.org/
Getting Started: https://info.openapparel.org/getting-started
Data & Technology: https://info.openapparel.org/data-technology
API: https://info.openapparel.org/api
Embedded Map: https://info.openapparel.org/embedded-map
FAQs: https://info.openapparel.org/faqs
About Us: https://info.openapparel.org/about-us
Funding: https://info.openapparel.org/funding
Press: https://info.openapparel.org/press
Work with Us: https://info.openapparel.org/work-with-us
Contact Us: https://info.openapparel.org/contact-us
Stories + Resources: https://info.openapparel.org/stories-resources

**Footer:**
Contact Us: https://info.openapparel.org/contact-us
Work with Us: https://info.openapparel.org/work-with-us
FAQs: https://info.openapparel.org/faqs
Privacy Policy: https://info.openapparel.org/privacy-policy
Terms of Use: https://info.openapparel.org/terms-of-use

cc @katieashaw 

## Tasks
- Audit every header and footer link (including mobile burger menu links)
  - Make sure the href is correct and the target is set to open in a new window/tab (#1610)

Issue #1510, opened 2021-10-29
All links to info website should open in new tab


hlennett:
## Overview

For any links to info.openapparel.org on the map site, we'd like them to open in a new tab when clicked. This includes almost all the new Header/Footer links. 

cc @katieashaw 

Shares point with #1512

katieashaw:
Can I ask @lederer your views on this. Hannah & I are both slightly on the fence about whether links between info.openapparel.org and openapparel.org should open in new tabs. We think "yes they should open in new tabs" given that technically these are two different sites. But the "no" counter argument is that we're trying to create a more seamless experience between the two sites & the new header menu bar etc does a lot of that work for us. Do you have a view on this?

lederer:
Good question. I think it might depend on the direction… 

If you're starting on the info site, I would posit that "Explore…" should open in the same tab (as it currently does). The rationale being that you're first getting the scoop on what this thing is, figuring out if it's for you, and ok now you've learned enough, let's do it. Done with the info site, don't keep it around, let's get to work in the app.

Whereas once you're in the app, you're doing work. Searching, maybe editing, etc. It might've take a fair amount of work to navigate to _these_ coordinates at _this_ zoom level, looking at results matching _these_ filters. You don't want to lose that work. But you may need some additional context about the OAR in parallel with that. So in that light, clicking from the app to the info site should open in a new tab.

I would advocate for that approach. But I'll include a couple of counterpoints:
- It creates an inconsistency between the app and info site. I think that's reasonable, but nonetheless it has its costs.
- There's a good general argument to be made that most links should not open in a new tab – because if someone wants a link to open in a new tab, they can choose to do that on their own (via right-click menu, etc). It's a good rule of thumb, since _open in new tab_ is overused in the wild. But I think exceptions are reasonable when there's a decent, use case-driven rationale like above.

katieashaw:
This is so helpful, thanks @lederer !

PR #1520, opened 2021-11-03
Correct info site routes

TaiWilkin:
## Overview

Adds the info site routes as constants in the constants file,
instead of using their string value directly. This allows us to
update the links in one place in order to keep them consistent and
avoid outdated links around the site.

Updates the info links throughout the site to use the constant values.

Updates the info site routes to use the most current version of the
URLs.

Confirms the info site links open in a new tab. 

Connects #1512, #1510 

## Notes

Technically the issue only referenced the header and footer links, but it saves time to update the miscellaneous outdated info site links throughout the site at the same time instead of pushing them off to another issue. 

## Testing Instructions

* Run `./scripts/server`
* For each link in the header (in both web and mobile view) and footer, confirm that it matches the appropriate link in the issue and that it opens in a new tab. 
* Click the info site links on the contribute page, a list page, the cookie preferences text, the Guide tab, the and the embedded map config page and confirm they open in a new tab. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Correct info site routes [#1520](https://github.com/open-apparel-registry/open-apparel-registry/pull/1520)
+
 ### Security
 
 ## [51] 2021-11-04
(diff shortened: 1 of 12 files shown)


jwalgran (approved):

TaiWilkin:
Thank you for reviewing!

Issue #1493, opened 2021-10-15
Create ExtendedField rows when creating FacilityMatch rows


jwalgran:
## Overview

Any time we parse a `FacilityListItem` (via list or via API), we should create `ExtendedField`s. We should create `ExtendedField`s when submitting facility claim details. 

Depends on #1492

### Describe the solution you'd like

- Update the `ExtendedField` model to allow null for `facility`

- When parsing a FacilityListItem, look for well known extended field keys in the submitted list item and create `ExtendedField` rows. The `facility` field will be null. Update the `facility` when the `FacilityListItem`s facility is updated.
- When a moderation activity changes the facility on a facility list item, also update `ExtendedFields`s (when promoting a facility list item)

- When `post`ing to `get_claimed_details`, look for well known extended field keys and create_or_update  `ExtendedField`s.  
     - Loop through known claim fields that correlate to known ExtendedFields. If a value is present and not blank/empty, create_or_update an ExtendedField. If it is blank or missing, look for existing ExtendedFields for the given claim for the field and delete them. 

- Create a follow-up card to add `ExtendedField` deletion when deleting a facility.

PR #1527, opened 2021-11-12
Create ExtendedFields

TaiWilkin:
## Overview

Users can submit facility list items individually through the API or
via a list. During parsing, we search for known extended fields and create
ExtendedFields as applicable. Later, we assign a facility id to the
ExtendedFields if one has been assigned to the list item (for MATCHED
and NEW_FACILITY statuses).

For POTENTIAL_MATCH statuses, the item has not yet been assigned to a
facility. When the user confirms or rejects the match, the list item and
ExtendedFields are both linked to a facility.

Extended facility details can also be submitted via facility claim. In
this case, the facility has already been created, so the ExtendedFields
are linked to a facility immediately. A user can later update the
details supplied with their claim; in this case, changed fields are
updated, and removed fields are deleted.

Connects #1493 

## Notes

In the issue, we list `promote` as a location where a ExtendedField might need to be updated. However, while the facility is updated to point to a different FacilityListItem, the FacilityListItem (which the ExtendedField is linked to) still points to the same facility, so it doesn't seem that any changes are needed to the ExtendedField in response. 

## Testing Instructions

* Run `./scripts/manage migrate` 
* Upload a list, then process it with `./scripts/manage batch_process --list-id 16 --action parse`
* On your lists page, get the id of a successfully parsed list item, then look in the ExtendedFields in the admin. There should be five ExtendedFields containing appropriate data. They should not have a facility assigned to them yet. 
* Geocode and match the list. A new facility should be created.
```
```
* Get the OAR ID of the new facility, then view the ExtendedFields. They should now have the facility’s OAR ID assigned to them. (If instead of a new facility, a potential match is created, the fields will not be linked yet. Confirm the match to link the fields.)
* Create a facility using the API http://localhost:8081/api/docs/#!/facilities/facilities_create. ExtendedFields should be created and linked to the new facility (If instead of a new facility, a potential match is created, the fields will not be linked yet. Confirm the match to link the fields.)
```
{
    "country": "China",
    "name": "Nantong Jackbeanie Headwear &amp; Garment Co. Ltd.",
    "address": "No.808,the third industry park,Guoyuan Town,Nantong 226500.",
"number_of_workers": "200 - 1,000",
"native_language_name": "Jackbeanie Headwear"
}
```
* Claim a facility, then update the claim with facility details including the number of workers and the native language name. ExtendedFields should be created from the claim details. 
* Update the claim to have no number of workers listed and updating the native language name. The ExtendedField for the native language name should have been updated, and theExtendedField for the number of workers should have been deleted. 

## Checklist

- [ ] CI passes after rebase


jwalgran (commented):
Everything is working as described. I like the design of butting a set of helper function in an `extended_fields` module. I do want to discuss the decision to make ExtendedField for list item core fields.

From the issue:
> When parsing a FacilityListItem, look for well known extended field keys in the submitted list item and create ExtendedField rows.

The design in my head when I wrote this was that we would only create number_of_workers and native_language_name extended fields for submitted list items, so I was suprised to see country, name, and address. Name and address do make sense for claims. Did you opt to also create them for list items for consistency? 

If we do go ahead with creating extended field rows for all core fields, we will need to add a data migration to create the extended field records for every list item created before this feature. That could be created in a follow up issue/PR if appropriate.

TaiWilkin:
I've updated the number_of_workers to use the suggested regex processing and format the min and max found values. 
I was adding the address / etc for consistency, but I agree that it makes more sense to just add those for claims, so I have removed them from the list item processing.

jwalgran (approved):
Changes work well. 👍

TaiWilkin:
Thank you for reviewing!

Issue #1538, opened 2021-11-30
Add a feature flag for extended fields


jwalgran:
## Overview

Add a feature flag for extended fields using a waffle switch.

### Describe the solution you'd like

- Use previous feature flags as a model for creating a new `extended_fields` switch.

PR #1542, opened 2021-11-30
Add extended profile feature flag

TaiWilkin:
## Overview

Extended profiles is a long-developing feature which which involve
large changes to the UI. In order to enable ongoing maintenance releases
during development of the new feature, we have added a feature flag to
allow us to hide the new components until we are ready to release them.

Connects #1538 

## Demo

Screen Shot 2021-11-30 at 12 38 29 PM

## Testing Instructions

* Run `./scripts/manage migrate`.
* Run `./scripts/server` and confirm the new switch is present in the Django admin. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/migrations/0073_add_extended_profile_switch.py
+++ b/src/django/api/migrations/0073_add_extended_profile_switch.py
@@ -1,0 +1,16 @@
+from django.db import migrations
+
+
+def create_extended_profile_switch(apps, schema_editor):
+    Switch = apps.get_model('waffle', 'Switch')
+    Switch.objects.create(name='extended_profile', active=False)
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('api', '0072_make_extendedfield_facility_nullable'),
+    ]
+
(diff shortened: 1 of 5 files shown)


src/django/api/migrations/0073_add_extended_profile_switch.py
```diff
@@ -0,0 +1,16 @@
+from django.db import migrations
+
+
+def create_embedded_map_switch(apps, schema_editor):
```
jwalgran, line 4: To avoid confusion I would suggest renaming this function `create_extended_profile_switch`

jwalgran (approved):
Confirmed. One suggested fixup.

src/django/api/migrations/0073_add_extended_profile_switch.py
TaiWilkin: Good catch!

TaiWilkin:
Thank you for reviewing!

Issue #1537, opened 2021-11-30
Update extended fields when facilities are split or merged


jwalgran:
## Overview

Update extended fields when facilities are split or merged

### Describe the solution you'd like

- When splitting, change the facility on any extended fields connected to the list item.
- When merging, change the facility on any extended fields connected to the list item.

PR #1544, opened 2021-12-01
Update extended fields when matches are adjusted

TaiWilkin:
## Overview

When merging two facilities, splitting a facility list item into its
own facility, or transferring a list item from one facility to another,
the list items in question are adjusted to point to a new facility.

ExtendedFields which are created from a list item contain both a
reference to the list item and a reference to the facility. In order to
keep the fields and list item in sync, we now update the ExtendedFields
when the list item's facility is changed.

Connects #1537 

## Testing Instructions

* Run `./scripts/server`
* Create extended fields from at least two separate list items (Item A and Item B). Note the ids of the matched / created facilities (Facility A, Facility B). 
* Merge Facility B into Facility A. The extended fields for Item B should point to Facility A and Item B. The fields for Item A should still point to Facility A and Item A. 
* Split Item B away from Facility A. The extended fields from Item B should no longer point to Facility A. 
* Transfer / move Item B to a new facility. The extended fields for Item B should now point to the new facility. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
+- Update extended fields when matches are adjusted [#1544](https://github.com/open-apparel-registry/open-apparel-registry/pull/1544)
+
 ### Changed
 
 ### Deprecated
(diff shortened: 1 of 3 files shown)


jwalgran (approved):
I created extended fields via shell_plus and verified that they were updated correctly when I used the dashboard to split and merge 👍

TaiWilkin:
Thank you for reviewing!

Issue #1545, opened 2021-12-02
Test breaking due to Google API response change


TaiWilkin:
## Overview

When geocoding a facility, if the country code returned by Google Maps is different from the country code provided by the user, we throw a ValueError.

We’re testing this code with an address that previously was getting mapped by Google Maps to China, and testing that we threw an error because the user had entered it as India. Google is now returning that address as being in India, which means that a value error is (correctly, but breaking the test) not raised. 

We need to update the test to mock the API response so that we are testing only our own code, and not third-party responses.

PR #1546, opened 2021-12-02
Fix broken geocoding test

TaiWilkin:
## Overview

When geocoding a facility, if the country code returned by Google Maps
is different from the country code provided by the user, we throw a
ValueError.

We’re testing this code with an address that previously was getting
mapped by Google Maps to China, and testing that we threw an error
because the user had entered it as India. Google is now returning that
address as being in India, which means that a value error was
(correctly, but breaking the test) not raised.

The test now mocks the API response so that we are testing only our
own code, and not third-party responses.

Connects #1545 

## Demo

Broken test:
Screen Shot 2021-12-02 at 11 24 17 AM

Passing tests:
Screen Shot 2021-12-02 at 11 24 57 AM

## Notes

Only one of the three geocoding tests were breaking, but I switched to the mock for all three. 

## Testing Instructions

* Run the tests and confirm they are all passing. 
* Visually confirm the mocked responses are reasonable given the tests' purposes. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Fix broken geocoding test [#1546](https://github.com/open-apparel-registry/open-apparel-registry/pull/1546)
+
 ### Security
 
 ## [52] 2021-11-10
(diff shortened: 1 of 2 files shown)


jwalgran (approved):
Confirmed. Nice use of the built-in mocking helpers. 

It's not every day that you get to report a 135x performance boost :-)

```
Ran 3 tests in 1.058s
```

```
Ran 3 tests in 0.008s
```

TaiWilkin:
Thank you for reviewing!

Issue #1529, opened 2021-11-22
Embedded Map: List of facilities in search results not extending the full length of the Embedded Map


hlennett:
## Overview

In live Embedded Maps, the scrolling results of a search do not extend the full height of the Embedded Map. 

### Expected Behavior

In the full OAR, while we know that the initial page of search filters does not have enough content to fill the full height of the map in all sizes, the search results will then extend to fill the full height of the map. We would expect this to be reflect in Embedded Map. 

### Actual Behavior

In Embedded Maps, search results stop loading halfway-ish down the side of the map, instead of extending the full length. 

### Steps to Reproduce

View live maps:
https://www.textilbuendnis.com/en/lieferkettentransparenz/
https://www.levistrauss.com/sustainability-report/community/supplier-map/

### Demo

Here's how it looks on the OAR (correct behavior):

![Screen Shot 2021-11-22 at 12 40 25 PM](https://user-images.githubusercontent.com/80268596/142855933-89ebf4e5-c80b-4e84-b1d1-03a5addcf2a1.png)

Here's how it looks on an Embedded Map (bug behavior):

![Screen Shot 2021-11-22 at 12 44 12 PM](https://user-images.githubusercontent.com/80268596/142856057-95bca957-c815-4c0d-9c53-c95f8569c7b3.png)

![Screen Shot 2021-11-22 at 12 44 07 PM](https://user-images.githubusercontent.com/80268596/142856071-c224c27e-105e-42b5-aab5-df3c61be892a.png)

You can also see it demoed in this video capture: https://us02web.zoom.us/rec/share/DY7yHahVQjPA4xRbw9YOo-4aaTfg0CyGecRmx3BMrnDESV5NL1a4howqyV0kCtwA.0VoVzHsjVyxUb1bd Passcode: +^Tr0Db1

### Additional context

cc @katieashaw

Issue #1528, opened 2021-11-22
Embedded Map: Powered by graphic flashes big while map is loading and then never shows up below map


hlennett:
## Overview

In the design of the Embedded Map, a small graphic that says "Powered by the OAR" is meant to appear in the bottom corner of each Embedded Map. It is currently not displaying on live Embedded Maps. Instead, when the map is loading, it flashes in a large size and then disappears when the map loads. 

### Expected Behavior

The powered by OAR graphic should appear in the bottom corner of the Embedded Map and not show up in a huge size when loading. 

### Actual Behavior

We are not seeing the powered by OAR graphic on Embedded Maps and the loading behavior with the large graphic is occurring. 

### Steps to Reproduce

Load live Embedded Maps on any of these sites to observe (you'll want to scroll to the part of the page with the map to observe loading:
https://www.levistrauss.com/sustainability-report/community/supplier-map/
https://www.textilbuendnis.com/en/lieferkettentransparenz/

### Demo

Captured a video of the behavior here: https://us02web.zoom.us/rec/share/DY7yHahVQjPA4xRbw9YOo-4aaTfg0CyGecRmx3BMrnDESV5NL1a4howqyV0kCtwA.0VoVzHsjVyxUb1bd Passcode: +^Tr0Db1

cc @katieashaw

hlennett:
@mariel-oar I am no longer seeing this behavior. Feel free to close.

PR #1547, opened 2021-12-02
Correctly position and size embed footer badge

lederer:
## Overview

Set the size and position of the embed footer badge, and adjust embed map height to make room for it. This fixes two issues:

#1528 where the embed badge was appearing at a huge size before the map loads and then disappearing after the map loads
#1529 where the embed list height did not use up the available space

Connects #1528 
Connects #1529 

## Demo

image

## Notes

In troubleshooting this, I noticed significant CSS debt throughout the app, which made it quite challenging to fix, even though the final fix is quite short. Recommend rearchitecting the (layout, if not additional) CSS when possible.

## Testing Instructions

* `./scripts/server`
* Login as a user with an embed plan (might require some admin mojo)
* Settings > Embed
* Ensure OAR badge appears at correct size below map
* Adjust embed width or height to force it to rerender. Ensure badge doesn't appear huge before map appears.
* Switch to Facilities tab. Ensure results list uses up entire available height of sidebar.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/Routes.jsx
+++ b/src/app/src/Routes.jsx
@@ -97,7 +97,7 @@ class Routes extends Component {
     render() {
         const { fetchingFeatureFlags, embed, classes } = this.props;
 
-        const mainPanelStyle = embed ? { top: 0 } : {};
+        const mainPanelStyle = embed ? { top: 0, bottom: '64px' } : {};
 
         return (
             <ErrorBoundary>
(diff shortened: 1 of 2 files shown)


TaiWilkin (approved):
This is working well! I also tested it outside of the preview in a simple HTML page and had no issues. 
I agree that a CSS cleanup pass will be helpful once workloads allow - I've run into similar issues with older CSS adding some additional time to tasks.

PR #1567, opened 2021-12-23
Fix contributor id bug

TaiWilkin:
## Overview

When getting the contributor id for the facilities list from the
request params `contributors`, we were incorrectly grabbing it as a
string instead of a list. Then, when accessing the first item, the
first digit of the contributor's id was selected instead of the first
contributor id. This resulted in the wrong contributor being selected
when the contributor had a multi-digit id. Now we parse the param as
a list.

This bug impacts both the prefer_contributor_name setting and the contributor fields download. 

## Testing Instructions

* Give a contributor with a multi-digit id embed permissions. 
* Using that contributor, submit facilities with extra fields. Use fields / values not belonging to other contributors in your system.
* In the settings, set up their embedded map and view the preview. Confirm that the downloaded values (from the CSV /XLSX) match the ones submitted for this contributor
* Update a facility match for the contributor to use a different name from the currently set facility name. Set prefer_contributor_name to true and view the embedded map preview. The custom name should be shown. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Use newer version of terraform container to avoid CI error [#1566](https://github.com/open-apparel-registry/open-apparel-registry/pull/1566)
 - Add contributor fields to facility downloads [#1565](https://github.com/open-apparel-registry/open-apparel-registry/pull/1565)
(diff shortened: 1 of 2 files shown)


jwalgran (approved):
I confirmed the fix. This was a sneaky byg. Nice detective work.

TaiWilkin:
Thank you for reviewing!

Issue #1549, opened 2021-12-08
Try to make exact string matches on country/name/address before dedupe


jwalgran:
## Overview

Try to do a string match (using "clean" strings that remove punctuation ,etc) before using dedupe.

There are two cases in which this would be helpful

- Avoiding the situation where submitting an updated list results in POTENTIAL_MATCH es because there are multiple matches with about 80% quality
- The situation where submitting the same facility data to the API twice ends up generating 2 new facilities rather than 1 facility and one match. This should not happen, but has appeared in testing and may be caused by each web worker having its own dedupe model.

### Describe the solution you'd like

- Add`clean_name` and `clean_address` on  the `FacilityListItem`
  - Index these new columns together with `country_code`
  - Add a RunPython block to the migration to backfill `clean_name` and `clean_address`
    - Add an empty reverse function
- Populate `clean_name` and `clean_address` these during parsing using existing clean function use by the existing matching code
- Before matching list or single item with dedup, look for exact item matches 
  - Not rejected or pending. We are looking for items with a good match to a facility
  - There could be multiple items that match. Choose the "best" one
    - Active
    - Prefer same contributor, not require
    - Newest
  - Try to make exact matching produce similar output to dedupe so that we can minimize the changes to `save_match_details`.
    - We do want to record that we skipped deduce in the `processing_results` 
  - Exclude items that are successfully "exact matched" from the follow up dedupe match
- Single item match has a different code path. Use helper functions.

PR #1568, opened 2022-01-05
Identify exact matches pre-dedupe

TaiWilkin:
## Overview

When matching facilities, look for exact matches prior to entering
the dedupe process. Exact matches (for name, address, and country)
can be automatically matched, and don't need to enter the dedupe
process.

Adds new fields to FacilityListItems to store the cleaned name and cleaned address, allowing efficient filtering for exact matches. 

If multiple exact matches exist, we prefer first the active list items; then items submitted by the same contributor; then finally the most recent items. 

Connects #1549 

## Demo

Screen Shot 2022-01-05 at 9 27 44 AM

## Notes

PPE handling was not implemented for exact-match items. 

## Testing Instructions

* Run migrations. 
* Login as c2@example.com, and navigate to http://localhost:8081/api/docs/#!/facilities/facilities_create. Submit an exact copy of an existing facility with `create=true`:
```
{
"name": "Alif Embroidery Village Ltd.",
"address": "Bangobandhu Road, Tangabari, Ashulia, Savar, Dhaka 1341",
"country": "Bangladesh"
}
```
* Confirm that the item has matched. Note the facility id. In an incognito browser, login as c1@example.com and navigate to the new FacilityListItem in the [Django Admin](http://localhost:8081/admin/api/facilitylistitem). In the processing results, you should see `exact_match: true`.
* As c1@example.com, navigate to [adjust facility matches](http://localhost:6543/dashboard/adjustfacilitymatches). Transfer the match you just created from the facility it matched to to any other facility. 
* Resubmit the same facility via the Swagger API. It should now match to the facility you transferred the FacilityListItem to. 
* Deactivate both newly created matches. Submit the same facility details via CSV.
[duplicates_test.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/7815785/duplicates_test.csv)
* Process the list:
```
```
* Confirm that the newly uploaded item matched to the original facility via exact match. 

## Checklist

- [ ] CI passes after rebase


src/django/api/processing.py
```diff
@@ -492,3 +494,72 @@ def make_pending_match(item_id, facility_id, score):
...
+            status=FacilityMatch.PENDING,
+            results={})
+
+    all_matches = []
+    for item_id, exact_matches in item_matches.items():
+        item = FacilityListItem.objects.get(id=item_id)
+        item.status = FacilityListItem.POTENTIAL_MATCH
+
+        matches = [make_pending_match(item_id, m.get('facility_id'))
+                   for m in exact_matches]
```
jwalgran, line 534: After running through the testing instructions, the last list item ended up with 3 match rows. I originally thought this was a problem, but I confirmed that this matches the behavior of the normal deduple process where we would create matches for both a 60% and 90% confidence match but the high quality match would be the `AUTOMATIC` winner. 👍 

```
openapparelregistry=# select id, results, status, facility_id, facility_list_item_id from api_facilitymatch where facility_list_item_id = 935;
 id  |                 results                  |  status   |   facility_id   | facility_list_item_id
-----+------------------------------------------+-----------+-----------------+-----------------------
 937 | {"match_type": "multiple_exact_matches"} | AUTOMATIC | BD202200698XDZC |                   935
 938 | {}                                       | PENDING   | US202200661FFPF |                   935
 939 | {}                                       | PENDING   | US202200661FFPF |                   935
(3 rows)
```

jwalgran (approved):
The implementation looks good and I was able to run through the test successfully. We will want to pay attention to how this performs on staging when we have 2 orders of magnitude more list items.

TaiWilkin:
Thank you for reviewing!

Issue #1554, opened 2021-12-09
November figures for Percent Data by Source Type are way off


hlennett:
## Overview

Looking at this report: https://openapparel.org/admin/reports/percent-data-by-source-type/

The percentages between Single and List submissions usually add up to roughly 100%. For November 2021, that figure was way off:

![Screen Shot 2021-12-09 at 9 17 53 AM](https://user-images.githubusercontent.com/80268596/145359479-73e64bab-f13e-4edd-9c87-c519e96058d1.png)

We would love to understand why the numbers are so different for this month and if they are accurate. We will need an answer to this by mid-January 2022 (Jan 14 at the latest), so we can finalize reporting for funders.

hlennett:
cc @katieashaw

katieashaw:
Flagging this for @obrienad & @jwalgran giving the up-coming deadline

jwalgran:
We added this issue to the current sprint are are working on debugging the query behind this report.

TaiWilkin:
I have identified the cause of the data issue. A fix is in-process, but meanwhile we will manually pull a corrected version of the data for the OAR team. 

For each facility, we look at all the times it was submitted and it was successfully matched / created. Some facilities were submitted multiple times by the same contributor, but we don’t want to count duplicates / resubmissions. 

An example of a single facility’s many submissions pre-filtering (data anonymized with false IDs):
<img width="225" alt="Screen Shot 2022-01-10 at 12 43 38 PM" src="https://user-images.githubusercontent.com/21046714/148813125-413db328-8eed-465a-b5da-410b9b0b77b3.png">

When filtering to look at only the initial submission by each contributor, we were filtering the submissions in two different ways when calculating the number of submissions for each source type vs calculating the total number of submissions. 

When calculating the number of facilities submitted via each source type per month, we were asking: What was the first time this contributor submitted this facility via this source? By this measure, you can see that contributor 2 is considered to have submitted the facility twice - once by list in February 2020 and once by API in November 2021. Counting this way, the facility was submitted three times (twice by list and once by API) in November.
<img width="229" alt="Screen Shot 2022-01-10 at 12 44 14 PM" src="https://user-images.githubusercontent.com/21046714/148813210-c4b0a628-9b1f-4c83-9716-eaf95436ebc8.png">

[...]

PR #1575, opened 2022-01-10
Fix the Percent Data by Source Type report

TaiWilkin:
## Overview

The Percent Data by Source Type report was using a different method
to count the total number of submissions per month (only counting the
first instance of each unique contributor/facility combination) than
the method used to count the number of submissions per source type
(only counting the first instance of each unique contributor/facility/
source_type combination), resulting in invalid percentages. Adjusts
the query to use the first instance of each unique contributor/facility/
source_type combination for both sub-calculations for consistency.

Connects #1554 

## Demo

Screen Shot 2022-01-10 at 12 35 27 PM

## Notes

Please see the more detailed write-up describing the bug in [the original issue](https://github.com/open-apparel-registry/open-apparel-registry/issues/1554#issuecomment-1009175227).

## Testing Instructions

* Run `./scripts/server` and confirm [the report](http://localhost:8081/admin/reports/percent-data-by-source-type/) appears correct.
* View the manually pulled report in the management channel in Slack (or manually pull the report yourself on Production) to confirm that the real data appears correct. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/reports/percent_data_by_source_type.sql
+++ b/src/django/api/reports/percent_data_by_source_type.sql
@@ -1,3 +1,6 @@
+-- Calculate the percentage of facilities submitted via API or List each month.
+-- Only count the first instance of each unique combination of facility/contributor/source_type.
+
 SELECT
     match.month,
     source_type,
@@ -9,9 +12,8 @@ FROM (
     FROM api_facilitymatch m
         JOIN api_facilitylistitem i on m.facility_list_item_id = i.id
         JOIN api_source s on i.source_id = s.id
-        JOIN api_contributor c ON s.contributor_id = c.id
     WHERE m.status NOT IN ('REJECTED', 'PENDING')
-    GROUP BY m.facility_id, c.id, s.source_type
+    GROUP BY m.facility_id, s.contributor_id, s.source_type
 ) as match
 JOIN (
     SELECT month, count(*) as total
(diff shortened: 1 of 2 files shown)


src/django/api/reports/percent_data_by_source_type.sql
```diff
@@ -9,9 +12,8 @@ FROM (
     FROM api_facilitymatch m
         JOIN api_facilitylistitem i on m.facility_list_item_id = i.id
         JOIN api_source s on i.source_id = s.id
-        JOIN api_contributor c ON s.contributor_id = c.id
```
TaiWilkin, line 12: This join was unnecessary, because we only used the Contributor id in the grouping query, and don't need any additional contributor details.

src/django/api/reports/percent_data_by_source_type.sql
```diff
@@ -21,9 +23,8 @@ JOIN (
         FROM api_facilitymatch m
             JOIN api_facilitylistitem i on m.facility_list_item_id = i.id
             JOIN api_source s on i.source_id = s.id
-            JOIN api_contributor c ON s.contributor_id = c.id
         WHERE m.status NOT IN ('REJECTED', 'PENDING')
-        GROUP BY m.facility_id, c.id
+        GROUP BY m.facility_id, s.contributor_id, s.source_type
```
TaiWilkin, line 27: Adding the `source_type` to the grouping here prevents the filtering out of rows where the contributor has resubmitted a facility using a different source type.

jwalgran (approved):
Looks good.

I did a double check by changing one of the development sources to `SINGLE`

```
openapparelregistry=# select count(*) from api_facilitylistitem where source_id = 2;
 count
-------
    75
(1 row)

openapparelregistry=# select count(*) from api_facilitylistitem ;
 count
-------
   932
(1 row)

openapparelregistry=# update api_source set source_type = 'SINGLE' where id = 2;
UPDATE 1
```

After that  my report looked like this

Screen Shot 2022-01-17 at 8 15 27 PM

75 / 932 = 0.080472103

There is a discrepancy, but it likely due to not all othe list items being completely processed.

TaiWilkin:
Thank you for reviewing!

Issue #1540, opened 2021-11-30
Add extended field handling to the new detail sidebar


jwalgran:
## Overview

Add extended field handling to the new detail sidebar

https://app.abstract.com/projects/441633f0-7d8c-11e9-af1e-972cbddca60a/branches/df31450b-d39a-4306-803d-c0d926ffd046/commits/72f5e69bf7a06d99182d72a68d99a26dfba4e1e8/files/fc1791e3-b8d2-4182-b240-fef32fabe978/layers/187B1943-8387-4CDB-AD91-71D0DB4C85FF?collectionId=1390af9f-d01b-4801-95a8-d8d14937a521&collectionLayerId=b5a60a72-c436-4d88-be3b-1274f5caa6d4

<img width="321" alt="Screen Shot 2021-11-30 at 9 25 39 AM" src="https://user-images.githubusercontent.com/17363/144086892-e9c0ef28-42fd-4ab3-b9a3-830c19ef8dcc.png">

Depends on #1539

### Describe the solution you'd like

- Clickable shows value count from API and opens modal to show detail.
  - Add clickable buttons for core fields as well (name, address, location)

Issue #1585, opened 2022-01-14
Add Extended Field Support


jwalgran:
# Task 026-A -  Extended Profile - Add Fields
## (1) Contributors can submit to data staging and see extended fields in API responses and the webside sidebar
January 20 - February 2, 2022
- [x]  :one: #1581 
- [x]  :five: #1582 
- [x]  :two: #1584 
- [x]  :one: #1586 
- [x]  :three: #1583 

February 3 - February 16, 2022
- [x]  1️⃣ #1618 (NOTE: This issue was added after the initial creation of the checklist)
- [x] :one: #1570
- [x] :three: #1540 

##  (2)
- [x] #1592 (formerly 3)
- [x] **CHECKPOINT 1** OAR has first group of stakeholders, including API users, upload and test the data on staging
- [x] **CHECKPOINT 2** February 8: OAR can show EM (on staging) to current customers so they know what changes to expect
- [x] February 10: OAR provides taxonomy feedback
- [x] February 11: OAR provides feedback from users

February 17 - March 2, 2022
- [x] #1623 
    - [x] :one: #1634
    - [x] :two: #1655
    - [x] :one: #1653
    - [x] 2️⃣ : #1638
- [x] :two: #1541 
- [x] :two: #1590 
    - [x] :two: #1648 
    - [x] :two: #1649
    - [x] :one: #1669 (added 2022-02-23)
- [x] :three: #1591 

# Task 016 - Search by additional fields
- [x]  :two: #1587 

March 3 - March 16, 2022
- [x] **CHECKPOINT 3** March 3: Everything is on production except for search. Test users & API users can upload to production and receive query results. The sidebar remains switched off (prior to official launch).
- [x] :two: #1562 
- [x] :one: #1588 
- [x] :five: #1589 

# Misc - Task 26
- [x] :two: #1689
- [x] :one: #1690
- [x] :two: #1703
- [x] :one: #1443
- [x] :one: #1706
- [x] :two: #1711

**March 17th Sprint**
- [x] :two: #1561
- [x] :one: #1704
- [x] :one: #1707

TaiWilkin:
Discuss: how do we handle search for extended fields when in embed mode - do we ignore extended field values submitted by other contributors?

Issue #1570, opened 2022-01-05
Include contributor verified status when sorting extended fields and prioritize claim over most frequent


jwalgran:
## Overview

Include contribuor verified status when sorting extended fields and prioritize claim over most frequent

### Describe the solution you'd like

- Rename `ExtendedField.trusted` to `ExtendedField.is_trusted` to be consistent with `Contributor.is_trusted`
- When sorting results
  - Consider `Contributor.is_trusted` in addition to `ExtendedField.is_trusted`
    - Sort by count of trust flags (2, 1, or 0)
  - Prioritize claimed ahead of most frequent when sorting 
- Include the other_names/_addresses/_locations information with the extended_fields when serializing to allow proper server-side sorting and inclusion of contributor data 
- Prioritize the name/address/location from an active/verified facility claim over the facility object's name/address/location when serializing the Facility.name, Facility.address, and Facility.location (unless it is in embed mode and `prefer_contributor_name` is active, in which case prefer the contributor name)
- Adjust the created_from to hide other contributor information in embed mode

PR #1593, opened 2022-01-18
Add extended field handling to facility details

TaiWilkin:
## Overview

Makes the facility details subsections expandable to show additional
content (details from other matches, claims, etc.) when not in embed
mode.

Adds extended field subsections (number of workers, native language
name, and so on) when values are present and the map is not in embed
mode.

Adds nonstandard/contributor field subsections when values are present
in embed mode.

Connects #1540 

## Demo

Closed items
Screen Shot 2022-01-18 at 2 02 03 PM

Opened items
Screen Shot 2022-01-18 at 2 02 12 PM

Extended fields
Screen Shot 2022-01-18 at 2 02 59 PM

Nonstandard / contributor fields (in embed)
Screen Shot 2022-01-18 at 2 21 50 PM

## Notes

Anticipated extended fields have been included in the list of fields to attempt to render if available, in addition to the existing extended fields. However, because these are not yet present in the API response, some tweaking might be necessary once they are being returned from the API to ensure all formatting / etc are working as expected. 

Some changes have been made to the styling of the items and header based on the updated wireframes. 

## Testing Instructions

* Run `./scripts/server`
* View a facility with multiple matches and confirm that other names, addresses, and GPS coordinates are available when expanding the subsections. 
* Submit and approve a claim for the facility, then submit a claim name, address, and number of workers. 
* The new claim details should be displayed in the expandable lists of facility details. 
* Submit and setup nonstandard fields for the facility. 
* Open the facility details in the embedded map preview. No items should be expandable, the contributors sections and expanded fields should be hidden, and the nonstandard fields should be shown. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/components/FacilityDetailSidebarContributors.jsx
+++ b/src/app/src/components/FacilityDetailSidebarContributors.jsx
@@ -1,9 +1,11 @@
 import React from 'react';
 import { withStyles } from '@material-ui/core/styles';
 import Typography from '@material-ui/core/Typography';
-import Divider from '@material-ui/core/Divider';
+import ListItem from '@material-ui/core/ListItem';
+import ListItemText from '@material-ui/core/ListItemText';
 
-import FacilityDetailSidebarDetail from './FacilityDetailSidebarDetail';
+import ShowOnly from './ShowOnly';
+import BadgeVerified from './BadgeVerified';
 
 import { makeProfileRouteLink } from '../util/util';
 
@@ -18,6 +20,9 @@ const detailsSidebarStyles = theme =>
             textTransform: 'uppercase',
             fontWeight: theme.typography.fontWeightMedium,
         },
+        primaryText: {
+            wordWrap: 'break-word',
+        },
         secondaryText: {
             color: 'rgba(0, 0, 0, 0.54)',
             display: 'flex',
@@ -25,8 +30,14 @@ const detailsSidebarStyles = theme =>
             fontSize: '12px',
             justify: 'flex-end',
         },
-        divider: {
-            backgroundColor: 'rgba(0, 0, 0, 0.06)',
+        icon: {
+            color: 'rgb(106, 106, 106)',
+            fontSize: '24px',
+            fontWeight: 300,
+            textAlign: 'center',
+        },
+        contributor: {
+            boxShadow: '0px -1px 0px 0px rgb(240, 240, 240)',
         },
     });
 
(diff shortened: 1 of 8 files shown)


src/app/src/components/FacilityDetailSidebarContributors.jsx
```diff
@@ -38,18 +49,30 @@ const FacilityDetailSidebarContributors = ({ classes, contributors, push }) => {
...
-                    secondary={contributor.list_name}
-                    hasAdditionalContent={!!contributor.id}
-                    hideTopDivider
-                    additionalContentCount=""
-                    onClick={() => push(makeProfileRouteLink(contributor.id))}
-                    key={contributor.id}
-                    verified={contributor.is_verified}
-                />
+            {visibleContributors.map(contributor => (
+                <ListItem
```
TaiWilkin, line 53: Switched from using the FacilityDetailSidebarDetail here because the structure is different enough now to make it overly complex

src/app/src/components/FacilityDetailSidebarDetail.jsx
```diff
@@ -13,56 +11,27 @@ const detailsSidebarStyles = () =>
...
-    onClick,
-    hideTopDivider,
     verified,
     classes,
 }) => (
-    <>
-        <ShowOnly when={!hideTopDivider}>
-            <Divider className={classes.divider} />
+    <ListItem className={classes.item}>
+        <ShowOnly when={verified}>
```
TaiWilkin, line 26: I'm showing the verified badge when an extended field is verified (presumably includes when the contributor is verified, based on upcoming changes). Do we want this, or should there be no badge for verified items?

src/app/src/components/FacilityDetailSidebarHeader.jsx
```diff
@@ -90,7 +90,7 @@ const getContent = ({
         secondary: 'Claim this facility',
         icon: <BadgeUnclaimed />,
         style: {
-            background: 'rgb(9, 18, 50)',
+            background: 'rgb(61, 50, 138)',
```
TaiWilkin, line 93: Updating to new unclaimed header color per new wireframes

src/app/src/util/constants.js
```diff
@@ -715,3 +715,37 @@ export const facilitySidebarActions = {
     CLAIM_FACILITY: 'Claim this facility',
     VIEW_ON_OAR: 'View on the Open Apparel Registry',
 };
+
+export const EXTENDED_FIELD_TYPES = [
```
TaiWilkin, line 719: The constant array of extended field types includes anticipated fields which aren't yet included in the API

src/app/src/components/FacilityDetailSidebarDetail.jsx
jwalgran: It makes sense that we visually indicate the status of these items. I like shoing the badge.

jwalgran (approved):
Nice work on the implementation! The code is clear and I was able to successfully work through the testing instructions without any issues.

I did see some things I did not expect, but I don't think they necessarily fall withing the scope of this PR. Please create follow up issues or add subtasks to one of our existing issues (listed under #1585) if appropriate.

1. After claiming the facility, I would expect the values for the claimed fa facility to be promoted to the top of the list. In my testing it appeared at the end of the list

Screen Shot 2022-01-19 at 12 28 48 PM
 
2. In my testing some addresses appear without a contributor name or date. 

Screen Shot 2022-01-19 at 12 41 52 PM

3. In my test I created an embedded map as "Service Provider A" but when I view a facility that was originally created under a different contributor ("Factory A"), I see that contributor's name. We need to avoid showing any other contributor names.

Screen Shot 2022-01-19 at 12 35 08 PM

TaiWilkin:
>1. After claiming the facility, I would expect the values for the claimed fa facility to be promoted to the top of the list. In my testing it appeared at the end of the list
This is caused by a) the prioritization of the name/address/location on the facility over the extended fields, and b) the combination/prioritization of the other_names/_addresses/_locations with the extended fields.
>2. In my testing some addresses appear without a contributor name or date.
This is caused by the combination of the other_names/_addresses/_locations with the extended fields (they don't include the contribution data).
>3. In my test I created an embedded map as "Service Provider A" but when I view a facility that was originally created under a different contributor ("Factory A"), I see that contributor's name. We need to avoid showing any other contributor names.
This is caused by the prioritization or and inclusion of contribution data for the facility object's name/address/locations. 
These three issues will be fixed by updating the facility details and list APIs, specifically by: 
- Including the other_names/_addresses/_locations information with the extended_fields when serializing to allow proper server-side sorting and inclusion of contributor data 
- Prioritizing the name/address/location from an active/verified facility claim over the facility object's name/address/location when serializing (unless it is in embed mode and `prefer_contributor_name` is active, in which case prefer the contributor name)
- Adjusting the created_from to hide other contributor information in embed mode
These tasks have been added to [#1570](https://github.com/open-apparel-registry/open-apparel-registry/issues/1570)

TaiWilkin:
Thank you for reviewing!

Issue #1580, opened 2022-01-14
Handle fhe case when exact matching to a non matvhed items submitted with create=false


jwalgran:
### Overview

If facility is submitted via API with `create=false` **and there is no match** a facility list item is created, but there is no facility match or facility. Subsequent submission of the same facility details will be an exact match to the item, but an exception is raised.

In this case, we should ignore the item instead and choose the next exact matching item or, if there are no other exact matches, proceed to the dedupe process.

Ensure that that if there is an automatic dedupe match, `FacilityListItem.facility` is set, allowing subsequent exact matches to avoid the exception. 

```
"Traceback (most recent call last):
  File \"/usr/local/src/api/views.py\", line 1301, in create
    facility = Facility.objects.get(id=facility_id)
  File \"/usr/local/lib/python3.7/site-packages/django/db/models/manager.py\", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File \"/usr/local/lib/python3.7/site-packages/django/db/models/query.py\", line 408, in get
    self.model._meta.object_name
api.models.Facility.DoesNotExist: Facility matching query does not exist.
",
```
         
This is a fix to the original feature was added in #1568 where we implemented to address https://github.com/open-apparel-registry/open-apparel-registry-clients/issues/61 
  
### Steps to Reproduce

- Browse /api/docs/#!/facilities/facilities_create authorize with an API token
- Submit a facility with `create=false` twice. The second submission should result in `ERROR_MATCHING`
```
{
    "country": "Turkey",
    "name": "BAŞARI TEKSTİL",
    "address": "Merkezefendi"
}
```

PR #1594, opened 2022-01-18
Handle create=false when exact matching

TaiWilkin:
## Overview

When a facility list item is submitted with `create=false` and there is
no match found, a `FacilityListItem` is created with `facility_id=None`.
When another facility list item is submitted which exact matches that
`FacilityListItem`, an error was being thrown. Instead, we now filter
out `FacilityListItems` where the `facility_id=None`, allowing us to match
to the next best option (if one exists) or pass the `FacilityListItem`
on to the dedupe process.

Connects #1580 

## Testing Instructions

* Not on this branch, browse /api/docs/#!/facilities/facilities_create authorize with an API token
* Submit a facility with `create=false` twice. The second submission should result in ERROR_MATCHING
```
{
    "country": "Turkey",
    "name": "BAŞARI TEKSTİL",
    "address": "Merkezefendi"
}
```
* Switch to this branch and resubmit the facility. The submission should not throw an error. 

## Checklist

- [x] CI passes after rebase


jwalgran (approved):
Confirmed!

TaiWilkin:
Thanks for reviewing!

PR #1596, opened 2022-01-20
Fix ExtendedField admin

TaiWilkin:
## Overview

Loading an individual ExtendedField in the Django admin on staging
causes a server timeout. It seems likely that this is due to the
automatic setup of the admin, which creates dropdown lists for
contributor, facility, facility claim, and facility list item, some of
which include thousands of items. Because we are unlikely to want to
update these fields in the admin (rather, we would create a new
extended field), I have updated them to be readonly, which prevents the
fetching of so much data and should resolve the timeout errors.

## Demo

Screen Shot 2022-01-20 at 10 06 02 AM

## Notes

A second option, if we do think we will want to allow updating any of these fields in the admin, could be to use a raw id field as described [here](https://books.agiliq.com/projects/django-admin-cookbook/en/latest/many_fks.html).

## Testing Instructions

* Run `./scripts/server`
* Visit an ExtendedField in the Django admin and confirm that it loads quickly and without fetching large amounts of related data. 

## Checklist

- [x] CI passes after rebase


TaiWilkin:
Thanks for reviewing!

Issue #1597, opened 2022-01-21
Empty `clean_` fields should not break parsing


TaiWilkin:
## Overview

When there is an error parsing a facility, we attempt to create a list item with an 'ERROR PARSING' status. If this list item has an empty address or name field, it throws an error due to the not-null rule for `clean_name` and `clean_address`. 

### Expected Behavior

The facility list item should be saved successfully with an ERROR_PARSING status.

### Actual Behavior

![Screen Shot 2022-01-21 at 11.44.48 AM.png](https://images.zenhubusercontent.com/5e41c4f13da639155d2ee456/c1c09f8b-6cd5-47c4-9182-6abaaba77603)

### Steps to Reproduce

* Submit a list with empty name and address fields and run `./scripts/manage batch_process --list-id 16 --action parse`
* There will be an error parsing (correctly), but then a second error will be thrown, breaking the processing script completely

### Demo

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.NotNullViolation: null value in column "clean_address" violates not-null constraint
DETAIL:  Failing row contains (960, 24, Tester,China,,,,,, ERROR_PARSING, null, null, [{"error": true, "trace": "Traceback (most recent call last):\n ..., , , CN, null, null, 2022-01-21 16:43:51.37385+00, 2022-01-21 16:44:25.963266+00, null, 19, null, null, null, null, null, null).


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "manage.py", line 15, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/usr/local/src/api/management/commands/batch_process.py", line 67, in handle
    self.process_items(facility_list, action, process)
[...]

PR #1600, opened 2022-01-24
Prevent `clean_` field error when parsing

TaiWilkin:
## Overview

When there is an error parsing a facility, we attempt to create a
list item with an 'ERROR PARSING' status. If this list item has an
empty address or name field, it threw an error due to the not-null
rule for clean_name and clean_address. If a `clean_` field would get
set to `None`, instead set them to an empty string.

Connects #1597 

## Demo

Screen Shot 2022-01-24 at 2 50 50 PM

## Testing Instructions

* Run `./scripts/server` and login as c2@example.com
* Contribute a list with empty fields and note the list id
[empty-fields.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/7929020/empty-fields.csv)
* Process the list using the appropriate list-id
```
```
* The list as a whole should not break, but there should be parse errors on fields with missing or invalid names or addresses
* Submit a facility via API with a missing name or address field and confirm that there is a validation error
* Submit a facility with a name or address that will `clean` to an empty string via API and confirm that there is a validation error
* Submit a valid facility via API and confirm that there is no error

## Checklist

- [ ] CI passes after rebase


jwalgran (approved):
I was unable to reproduce the original exceptions on the develop branch but the more explicit handling of empty values is a valuable improvement. I was able to successfully complete the test steps.

TaiWilkin:
Thank you for reviewing!

Issue #1582, opened 2022-01-14
Add facility type and production type taxonomy to code and write helper functions to look up values


jwalgran:
## Overview

Add facility type and production type taxonomy to code and write helper functions to look up values

Taxonomy: https://docs.google.com/spreadsheets/d/1HlGoYgj0rbtxnhgm0LAWk5lgFCSw3A2V/edit?pli=1#gid=432729308

### Describe the solution you'd like

-  [x]  Add the taxonomy to the codebase as constant data structures (https://docs.google.com/spreadsheets/d/1HlGoYgj0rbtxnhgm0LAWk5lgFCSw3A2V/edit#gid=432729308)
    -   Use [`clean`](https://github.com/open-apparel-registry/open-apparel-registry/blob/85d74bb1da4b88882fc21cf2a56fbfed68c014dc/src/django/api/matching.py#L39-L56) values as lookup keys      
-   [x] Add alias lookup list for common alternate representations of each facility type and processing type    
-   [x] Add [https://pypi.org/project/thefuzz/](https://pypi.org/project/thefuzz/) for string distance calculation
-   [x] Add a function to look up facility and processing types
    -   Arguments is a string that can be either a Facility Type or a Processing Type
    -   Returns tuple of (field_type, match_type, facility_type_value, production_type_value)
        -   Field type is either FACILITY_TYPE or PROCESSING_TYPE or None      
        -   Match type is either EXACT or ALIAS or FUZZY or ALIAS_FUZZY or None
        -   _value fields are the matched taxonomy values
    -   Proposed logic
        -   `clean` the input value
        -   Try to match exact processing type or facility type
        -   If no match try to match alias for production type or facility type
        -   If no match try using thefuzzz.process.extractOne to match against, in turn, processing types, facility types, processing type alias and facility type alias, only accepting a value with a score exceeding a threshold
    -   Add a unit test that uses a loop to assert matching inputs and outputs for many test cases

PR #1601, opened 2022-01-25
Add facility/processing type taxonomy and look-up function

caseycesari:
## Overview

The look-up function takes the input value, cleans it up by removing non-letter characters and extra spaces, and then attempts to find a match in the taxonomy using various methods.

Connects #1582

## Demo

```
>>> get_facility_and_processing_type('shredding')
('PROCESSING_TYPE', 'EXACT', 'Raw Material Processing or Production', 'Shredding')
>>> get_facility_and_processing_type('pulp')
('PROCESSING_TYPE', 'FUZZY', 'Raw Material Processing or Production', 'Pulp making')
>>> get_facility_and_processing_type('wet painting')
('PROCESSING_TYPE', 'FUZZY', 'Printing, Product Dyeing and Laundering', 'Wet roller printing')
>>> get_facility_and_processing_type('wet printing')
('PROCESSING_TYPE', 'ALIAS', 'Printing, Product Dyeing and Laundering', 'Wet roller printing')
>>> get_facility_and_processing_type('assembly')
('PROCESSING_TYPE', 'EXACT', 'Final Product Assembly', 'Assembly')
>>> get_facility_and_processing_type('assemble')
('PROCESSING_TYPE', 'FUZZY', 'Final Product Assembly', 'Assembly')
>>> get_facility_and_processing_type('final assembly')
('FACILITY_TYPE', 'ALIAS', 'Final Product Assembly', 'Final Product Assembly')
```

## Note

`FUZZY_ALIAS_MATCH` was not implemented because it ended up not being useful. All non-exact entries were either matched using the alias or fuzzy methods.

## Testing Instructions

- Run `./scripts/update` to install the new pip dependency.
- Run `./scripts/test` to ensure the new tests pass.
- Start a shell (`./scripts/manage shell`) and manually experiment with the matching function to see if it otherwise is working as intended.
- Review the taxonomy and suggest any edits.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/facility_type_processing_type.py
+++ b/src/django/api/facility_type_processing_type.py
@@ -1,0 +1,397 @@
+from api.matching import clean
+from thefuzz import process
+
+HEADQUARTERS = 'headquarters'
+NO_PROCESSING = 'no processing'
+OFFICE = 'office'
+OFFICE_HQ = 'office hq'
+RETAIL_SALES = 'retail sales'
+SOURCING_AGENT = 'sourcing agent'
+TRADING = 'trading'
+
+OFFICE_PROCESSING_TYPES = {
+    HEADQUARTERS: 'Headquarters',
+    NO_PROCESSING: 'No processing',
+    OFFICE: 'Office',
+    OFFICE_HQ: 'Office / HQ',
+    RETAIL_SALES: 'Retail Sales',
+    SOURCING_AGENT: 'Sourcing Agent',
+    TRADING: 'Trading',
+}
+
+OFFICE_PROCESSING_TYPES_ALIAS = {
+    'hq': OFFICE_HQ,
+    'sales': RETAIL_SALES,
+    'retail': RETAIL_SALES,
+    'sourcing': SOURCING_AGENT,
+}
+
+PACKING = 'packing'
+WAREHOUSING_DISTRIBUTION = 'warehousing distribution'
+
+WAREHOUSING_PROCESSING_TYPES = {
+    PACKING: 'Packing',
+    WAREHOUSING_DISTRIBUTION: 'Warehousing / Distribution',
+}
+
+WAREHOUSING_PROCESSING_TYPES_ALIAS = {
+    'warehousing': WAREHOUSING_DISTRIBUTION,
+    'distribution': WAREHOUSING_DISTRIBUTION,
+}
+
+ASSEMBLY = 'assembly'
+CUTTING = 'cutting'
+CUT_AND_SEW = 'cut & sew'
+EMBELLISHMENT = 'embellishment'
+EMBROIDERY = 'embroidery'
+FINAL_PRODUCT_ASSEMBLY = 'final product assembly'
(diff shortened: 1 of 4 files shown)


src/django/api/facility_type_processing_type.py
```diff
@@ -0,0 +1,397 @@
+from api.matching import clean
```
caseycesari, line 1: I put this stuff in it's own file since there are so many items.

src/django/api/tests.py
```diff
@@ -600,6 +608,68 @@ def test_incorrect_country_code_raises_error(self, mock_get):
         )
 
 
+class FacilityAndProcessingTypeTest(TestCase):
```
caseycesari, line 611: I know the issue said to create tests using a loop, but I found it easier to incrementally add functionality to the look up function with individual tests. I can condense if need be.

src/django/api/facility_type_processing_type.py
```diff
@@ -0,0 +1,397 @@
...
+        if not matched_value or matched_value is None:
+            matched_value = process.extractOne(
+                cleaned_input,
+                ALL_PROCESSING_TYPES.keys()
+            )
+            match_type = FUZZY_MATCH
+
+            # Match must score 85 or higher to be considered usable.
+            if not matched_value or matched_value[1] < 85:
+                return None
```
jwalgran, line 379: We may want to consider returning `(None, None, None, None)` here so that the "shape" of the return values from the function is always the same, which may make the calling code a litter clearer. 
```
>>> (a, b, c, d) = get_facility_and_processing_type('WILL NOT MATCH ANYTHING')
  File "<console>", line 1, in <module>
TypeError: cannot unpack non-iterable NoneType object
```

src/django/api/facility_type_processing_type.py
```diff
@@ -0,0 +1,397 @@
...
+FACILITY_TYPE = 'FACILITY_TYPE'
+PROCESSING_TYPE = 'PROCESSING_TYPE'
+
+# Match types
+EXACT_MATCH = 'EXACT'
+ALIAS_MATCH = 'ALIAS'
+FUZZY_MATCH = 'FUZZY'
+
+ALL_PROCESSING_TYPES = {
+    **OFFICE_PROCESSING_TYPES,
```
jwalgran, line 332: Nice use of `**` here to compactly build these derived data structures.

jwalgran (approved):
Nice implementation. I was excited to see how well this works. I made one suggestion on how we return when there is not match but we could defer that until we actually have client code using it.

src/django/api/facility_type_processing_type.py
caseycesari: I had that at first, but changed it to `None` because `(None, None, None, None)` evaluates to `not None`. But as you point out, I have some questions about the client code implementation that may require changing this.

caseycesari: Updated this. Thanks.

Issue #1534, opened 2021-11-29
Edit embed config to allow overriding the label on the search box


jwalgran:
## Overview

Edit embed config to allow overriding the label on the search box

Depend on  model updates in #1533

### Describe the solution you'd like

- Update API to allow passing `text_search_label` when updating `EmbedConfig`
    - Normalize empty string to null 
- Update embedded map UI to add a "Text search label" 
- Update search sidebar UI to show custom label is set on the `EmbedConfig`

emilyhu0106:
@jwalgran Just for clarification - do we need to update the placeholder for the input here accordingly? ![image.png](https://images.zenhubusercontent.com/60f85b66cbf2b73a7c1809b8/f3b5d5be-9783-4560-924f-e0d162ffd546)

jwalgran:
Yes. We do want to change the placeholder text.

PR #1604, opened 2022-01-27
Edit search label in embed config

emilyhu0106:
## Overview

Update embedded map UI to add a "Text search label" and search sidebar UI to show custom label which is set on the `EmbedConfig`

Connects #1534

### Embedded Map
image

## Testing Instructions

* Run `./scripts/server` and log in as any users with embed permissions (e.g. c2@example.com)
* Go to the Setting - Embed page to set the search label config
    - [x] The info text is descriptive enough for users
    - [x] When no fields are marked searchable, the custom search box is reset to the default text and disabled.
    - [x] No lag during typing
    - [x] If the text search label is blank, it is replaced with the default text
* Go to preview and embed mode map
    - [x] The search label changes accordingly

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/components/EmbeddedMapConfig.jsx
+++ b/src/app/src/components/EmbeddedMapConfig.jsx
@@ -87,6 +87,7 @@ function EmbeddedMapConfig({
         font,
         width,
         fullWidth,
+        textSearchLabel,
         height,
         preferContributorName,
     },
@@ -173,9 +174,17 @@ function EmbeddedMapConfig({
                             setPreferContributorName={updateEmbedConfig(
                                 'preferContributorName',
                             )}
+                            textSearchLabel={textSearchLabel}
+                            setTextSearchLabel={updateEmbedConfig(
+                                'textSearchLabel',
+                            )}
+                            anyFieldSearchable={fields.some(
+                                field => field.searchable,
+                            )}
                             errors={errors}
                         />
                     ) : null}
+
                     <EmbeddedMapThemeConfig
                         color={color}
                         setColor={updateEmbedConfig('color')}
@@ -225,6 +234,7 @@ EmbeddedMapConfig.propTypes = {
         width: string.isRequired,
         height: string.isRequired,
         fullWidth: bool.isRequired,
+        textSearchLabel: string.isRequired,
     }).isRequired,
     setEmbedConfig: func.isRequired,
     setFields: func.isRequired,
--- a/src/app/src/components/EmbeddedMapConfigWrapper.jsx
+++ b/src/app/src/components/EmbeddedMapConfigWrapper.jsx
@@ -108,7 +108,7 @@ function EmbeddedMapConfigWrapper({ user }) {
             } else {
                 createConfig();
             }
-        }, 500);
+        }, 1000);
 
         return () => {
             clearTimeout(handler);
--- a/src/app/src/components/EmbeddedMapFieldsConfig.jsx
+++ b/src/app/src/components/EmbeddedMapFieldsConfig.jsx
@@ -15,10 +15,15 @@ import ArrowDownwardIcon from '@material-ui/icons/ArrowDownward';
 import { withStyles } from '@material-ui/core/styles';
 
 import { DragDropContext, Droppable, Draggable } from 'react-beautiful-dnd';
+import EmbeddedMapSearchLabelConfig from './EmbeddedMapTextSearchLabelConfig';
 
 const styles = {
     section: {
         marginTop: '20px',
+    },
+    sectionDisabled: {
+        marginTop: '20px',
+        opacity: '30%',
     },
     sectionHeader: {
         color: 'rgb(0, 0, 0)',
@@ -104,6 +109,9 @@ function EmbeddedMapFieldsConfig({
     setPreferContributorName,
     classes,
     errors,
+    textSearchLabel,
+    setTextSearchLabel,
+    anyFieldSearchable,
 }) {
     const updateItem = item => {
         const index = fields.findIndex(f => f.columnName === item.columnName);
@@ -380,6 +388,19 @@ function EmbeddedMapFieldsConfig({
                         }
                         label="Use my facility name"
                         classes={{ label: classes.listText }}
+                    />
+                </div>
+                <div
+                    style={
+                        anyFieldSearchable
+                            ? styles.section
+                            : styles.sectionDisabled
+                    }
+                >
+                    <EmbeddedMapSearchLabelConfig
+                        anyFieldSearchable={anyFieldSearchable}
+                        textSearchLabel={textSearchLabel}
+                        setTextSearchLabel={setTextSearchLabel}
                     />
                 </div>
             </div>
--- a/src/app/src/components/EmbeddedMapTextSearchLabelConfig.jsx
+++ b/src/app/src/components/EmbeddedMapTextSearchLabelConfig.jsx
@@ -1,0 +1,102 @@
+import React, { useState, useEffect, useRef } from 'react';
+import Typography from '@material-ui/core/Typography';
+import { withStyles } from '@material-ui/core/styles';
+import ControlledTextInput from './ControlledTextInput';
+import { DEFAULT_SEARCH_TEXT } from '../util/constants';
+
+const styles = {
+    subsectionHeader: {
+        color: 'rgb(0, 0, 0)',
+        fontSize: '15px',
+        fontWeight: 'bold',
+        margin: '10px 0',
+    },
+    reset: {
+        background: 'none',
+        border: 'none',
+        padding: '0',
+        cursor: 'pointer',
+        textDecoration: 'underline',
+    },
+};
+
+function EmbeddedMapTextSearchLabelConfig({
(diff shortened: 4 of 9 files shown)


emilyhu0106:
In the issue #1534 description we have
> Update API to allow passing `text_search_label` when updating EmbedConfig
Normalize empty string to null

When I am implementing this change, I realize that whenever `text_search_label` is empty string or `null`, it will be updated into the default text. Should this normalization be achieved at API level?

src/app/src/components/EmbeddedMapConfigWrapper.jsx
```diff
@@ -108,7 +108,7 @@ function EmbeddedMapConfigWrapper({ user }) {
             } else {
                 createConfig();
             }
-        }, 500);
+        }, 1000);
```
emilyhu0106, line 111: 1000 ms is more practical to finish typing in the text input; 500 ms pause will cause a lag when typing a bit slower

src/app/src/components/FilterSidebarSearchTab.jsx
```diff
@@ -235,14 +239,20 @@ function FilterSidebarSearchTab({
...
+                                embed ? textSearchLabel : DEFAULT_SEARCH_TEXT
+                            }
                         >
                             Search a Facility Name, OAR ID, or PPE Product Type
                         </FeatureFlag>
                     </InputLabel>
                     <TextField
                         id={FACILITIES}
-                        placeholder="Facility Name or OAR ID"
+                        placeholder={
```
emilyhu0106, line 251: The placeholder text is the same as the text search label when it is embed mode and not default text.

src/app/src/components/EmbeddedMapConfig.jsx
```diff
@@ -166,16 +168,28 @@ function EmbeddedMapConfig({
...
+                                fields={fields}
+                                setFields={setFields}
+                                preferContributorName={preferContributorName}
+                                setPreferContributorName={updateEmbedConfig(
+                                    'preferContributorName',
+                                )}
+                                errors={errors}
+                            />
+                            <EmbeddedMapSearchLabelConfig
+                                anyFieldSearchable={fields.some(
```
emilyhu0106, line 182: For the search label input component, it only cares about if any field is searchable

TaiWilkin:
Taking a look now

TaiWilkin:
> When I am implementing this change, I realize that whenever `text_search_label` is empty string or `null`, it will be updated into the default text. Should this normalization be achieved at API level?

I think it's fine that we're managing the change to default text on the UI only, as it allows us to have a single source of truth for the default text.

TaiWilkin (changes requested):
As a whole this is working very well. The search text label change in embed config is being applied smoothly and consistently.
I found a couple of anomalies as I was testing the search input:
- The initial content of the text input is blank, but is updated when I click into it. You might be able to fix that by replacing empty text with the DEFAULT_SEARCH_TEXT [here](https://github.com/open-apparel-registry/open-apparel-registry/blob/4c30cf877ef95a8eb96ae86c74cb6ef344bac8f4/src/app/src/util/embeddedMap.js#L34). The downside of that approach is that if they leave the field blank long enough that it auto-saves, the text will get replaced with the default value. You could work around that by preventing autosaving the embed config while the textbox is blank (and only saving a blank field / replacing the text on blur). Let me know if you'd like to pair on that approach.
- When I enter multiple spaces, then click away from the textbox, it stays blank; however, if I click back into the box and then click out again without changing anything, it is replaced with the default text. I suggest using `.trim()` on the text when you check to see if the field is blank.

I'm not sure about the descriptive content, so I've tagged @lederer for review to give this a look over as well.

src/app/src/components/EmbeddedMapConfig.jsx
```diff
@@ -166,16 +168,28 @@ function EmbeddedMapConfig({
...
+                            <EmbeddedMapFieldsConfig
+                                fields={fields}
+                                setFields={setFields}
+                                preferContributorName={preferContributorName}
+                                setPreferContributorName={updateEmbedConfig(
+                                    'preferContributorName',
+                                )}
+                                errors={errors}
+                            />
+                            <EmbeddedMapSearchLabelConfig
```
lederer, line 181: Seems that `<EmbeddedMapSearchLabelConfig>` might be better off as a child of `<EmbeddedMapFieldsConfig>`, for a couple of reasons:
1. Semantically / structurally, the search label config should be a subsection of the "Fields" section. It looks that way on the rendered page, but in the DOM it's a sibling, not a child.
2. It is dependent on the fields config checkboxes.

src/app/src/components/EmbeddedMapTextSearchLabelConfig.jsx
```diff
@@ -0,0 +1,56 @@
...
+    };
+
+    useEffect(() => {
+        if (!anyFieldSearchable) {
+            setTextSearchLabel(DEFAULT_SEARCH_TEXT);
+        }
+    }, [anyFieldSearchable]);
+
+    return (
+        <>
```
lederer, line 34: This should probably be wrapped in a `<div>`, rather than a Fragment. That way it will be contained in a specific subtree of the DOM and can be considered a proper sibling of the other sub-sections (eg, Promoted Fields).

The wrapping `<div>` should have `marginTop: 20px` to match its siblings.

src/app/src/components/EmbeddedMapTextSearchLabelConfig.jsx
```diff
@@ -0,0 +1,56 @@
...
+            <Typography style={styles.subsectionHeader}>
+                Custom search text label
```
lederer, line 36: "Custom label for search box"

src/app/src/components/EmbeddedMapTextSearchLabelConfig.jsx
```diff
@@ -0,0 +1,56 @@
...
+            </Typography>
+            <Typography>
+                Set the text to display on your side bar when fields are marked
+                as searchable. Default is &quot;{DEFAULT_SEARCH_TEXT}
+                &quot;
```
lederer, line 41: "If you've made any of your custom fields searchable, you can write your own custom label for the search box, so users know what they can search for."

lederer: It would be awesome if instead of "Default is…" you could put a link that says "Reset to default". It would only be visible if the field isn't empty. Onclick, it would clear the field (and disappear itself).

lederer:
This is neat. Thanks for putting this together. I left some comments.

Currently if no "searchable" checboxes are checked, the input is disabled. This is good! Can you also make the entire subsection grayed out? Maybe just set the opacity of it to, say, 60%.

src/app/src/components/EmbeddedMapFieldsConfig.jsx
```diff
@@ -15,11 +15,16 @@ import ArrowDownwardIcon from '@material-ui/icons/ArrowDownward';
...
 import { DragDropContext, Droppable, Draggable } from 'react-beautiful-dnd';
+import EmbeddedMapSearchLabelConfig from './EmbeddedMapTextSearchLabelConfig';
 
 const styles = {
     section: {
         marginTop: '20px',
     },
+    sectionDisabled: {
+        marginTop: '20px',
+        opacity: '60%',
```
lederer, line 26: Turns out this is still too dark. `opacity: 0.3` should do it..

src/app/src/components/EmbeddedMapTextSearchLabelConfig.jsx
```diff
@@ -0,0 +1,92 @@
...
+
+    return (
+        <div>
+            <Typography style={styles.subsectionHeader}>
+                Search box label
+            </Typography>
+            <Typography>
+                Set the text to display on your side bar when fields are marked
+                as searchable.
+                <button
```
lederer, line 70: When this button renders, there is no space before it. This should fix that:

```suggestion
                Set the text to display on your side bar when fields are marked
                as searchable.
                {" "}
                <button
```

src/app/src/components/EmbeddedMapTextSearchLabelConfig.jsx
```diff
@@ -0,0 +1,92 @@
...
+                    onClick={onReset}
+                    type="button"
+                    style={styles.reset}
+                    className="inline-link"
+                >
+                    Reset to default
```
lederer, line 76: Would it be possible to make this "Reset" button only appear while the value in the text box != the default?

lederer (changes requested):
Thanks for making those changes. Few more tweaks requested in some code comments. Almost there!

emilyhu0106:
> Thanks for making those changes. Few more tweaks requested in some code comments. Almost there!

I made the changes accordingly. Please take another look, thank you!

TaiWilkin (approved):
Great job working through those changes! The fix for the default text onChange vs onBlur issue was surprisingly complicated to resolve, but you did a good job working through it.

emilyhu0106:
Thank you both for reviewing!

Issue #1535, opened 2021-11-29
Update facility indexing to include custom text


jwalgran:
## Overview

Update facility indexing to include populating `custom_text` with `EmbededField` values marked as `searchable` and
trigger a re-indexing of the `custom_text` field if embedded map fields change.

Depends on model changes in #1533 

### Describe the solution you'd like

- Avoid complicating the existing process by having a separate flow for indexing `custom_text`
  - When indexing populate `custom_text` with `EmbededField` values marked as `searchable.
  - Have full facility indexing call the separate custom text indexing method

PR #1607, opened 2022-01-28
Index custom text

TaiWilkin:
## Overview

Contributors want to mark specific contributor fields / embed fields to
be included in the text which is searched by the facilities query.
This adds the function to update the `custom_text` field in the
FacilityIndex, which will allow those custom fields to be included in
the facilities search.

Whenever the FacilityIndex is updated, and in post_save for
FacilityListItems, and when updating a user's embed config, the
index_custom_text function is passed an array of facility ids to
update.

We get a list of contributors who have searchable embed configs,
and for each of the facilities in the facility_ids list, find the most
recent FacilityListItem submitted for that facility by each contributor,
if one exists. Then we grab any fields from the `raw_data` of that
facility which are included in the submitting contributor's searchable
fields, and concatenate them into a single text value. Finally, we
update each facility's `custom_text` to the calculated combined text
value.

Connects #1535 

## Notes

The `custom_text_field` is reevaluated for effected facilities whenever the FacilityIndex is updated, but also in two other cases: 
- When a FacilityListItem is updated (because the facility's core data might not change and so not trigger the FacilityIndex, but we want to look at each contributor's most recent active data)
- When an EmbedField is updated. I initially tried putting a `post_save` trigger on the EmbedField model; however, we delete and recreate ALL of the embed fields for a contributor each time the config is updated, caused half a dozen instances of the index_custom_text method to run when it really should only run once. I also tried adding a `post_save` trigger to the EmbedConfig model, but because we first update the EmbedConfig and THEN update the EmbedFields, this caused the indexing to occur before the updates to the fields occurred. I ended up adding the reindexing as the final step in the `create_embed_fields` function, which we use both when creating and when updating (ie deleting and recreating) the EmbedFields, which causes the reindexing to occur once on each config update after the fields have updated. 

I applied several methods to attempt to keep time / overhead of the method lower, due to the relatively large amount of data processing required. 
- Using a dictionary to track contributor fields (allowing us to query for searchable contributor fields just once each time the index is run). 
- Using an `iterator` for the potentially larger queries which are used for loops
- Django's querying superpowers to cut number of FacilityListItems to loop through to a minimum (see code comments)

## Testing Instructions

* These directions assume you have run `./scripts/resetdb`
* In an incognito browser, navigate to [http://localhost:8081/admin/api/contributor/2/change/](the admin) log in as `c1@example.com` 
* Update Service Provider A to have deluxe embed permissions
* In another window, login as c2@example.com
* Contribute [oar-extra-fields.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/7961425/oar-extra-fields.csv)
* Fully process the list
```
```
* Make ‘custom_field_2’ visible and searchable in the embed config settings
* In a new terminal tab (keep the server running), run `./scripts/manage shell_plus` then run the following. You should receive 905 for the total FacilityIndex count and 0 for the filled `custom_text` count.
```
FacilityIndex.objects.count()
FacilityIndex.objects.filter(custom_text__isnull=False).exclude(custom_text=‘’).count()
```
* Setup the Embed Config (including height and width) and mark `custom_field` and `custom_field_2` visible (but NOT searchable) for c2@example.com in the embed config.
* The field values should appear in the Extra Fields List facilities in their details in the embed preview.
* Run the following in the shell. You should still see 0 filled `custom_text` values.
```
FacilityIndex.objects.filter(custom_text__isnull=False).exclude(custom_text=‘’).count()
```
* Mark `custom_field` searchable in the embed config.
* Run the following in the shell. 
```
FacilityIndex.objects.filter(custom_text__isnull=False).exclude(custom_text='').values('custom_text', 'name')
```
* This should return the following. Note that the custom text for the selected field is listed for each facility.
`<QuerySet [{'custom_text': 'Test Value!', 'name': 'Harts Are You'}, {'custom_text': 'Test Field Value', 'name': 'Hats Are Us'}]>`
* Mark `custom_field_2` searchable in the config settings. 
* Run the same command in the shell. 
```
FacilityIndex.objects.filter(custom_text__isnull=False).exclude(custom_text='').values('custom_text', 'name')
```
* This should return the following. Note the combined text value of the two searchable fields.
`<QuerySet [{'custom_text': 'Test Value!', 'name': 'Harts Are You'}, {'custom_text': 'Test Value 2Test Field Value', 'name': 'Hats Are Us'}]>`
* Mark `custom_field` NOT searchable and confirm it was successfully removed from the index using the same shell command.
* Navigate to http://localhost:8081/api/docs/#!/facilities/facilities_create and authenticate as c2@example.com
* Submit:
```
{
"country": "US",
"name": "Hats Are Us",
"address": "140 Simpson Ave Ardmore PA 19003",
"custom_field": "New API value"
}
```
* Run the same shell command again. The custom text should for Hats Are Us should be updated to 'Hats Are Us' (using the newest facility list item for the values). 
* Login as c3@example.com and submit [oar-extra-fields-2.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/7961489/oar-extra-fields-2.csv)
* Fully process the list:
```
./scripts/manage batch_process --list-id 17 --action parse
./scripts/manage batch_process --list-id 17 --action geocode
./scripts/manage batch_process --list-id 17 --action match
```
* Create an embed config for c3@example.com but leave the fields NOT searchable
* Confirm that the custom text values for the facilities have not changed.
* Make the fields searchable for c3@example.com, then view the indexes again.
`<QuerySet [{'custom_text': 'Test Value!C3 Custom 2', 'name': 'Harts Are You'}, {'custom_text': 'New API valueC3 CustomC3 Other', 'name': 'Hats Are Us'}]>`
* Confirm that the custom text now includes the other contributor’s text. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/helpers.py
+++ b/src/django/api/helpers.py
@@ -1,4 +1,6 @@
 import re
+import csv
+import json
 
 CONSONANT_SOUND = re.compile(r'''
 one(![ir])
@@ -20,3 +22,42 @@ def prefix_a_an(value):
     if not CONSONANT_SOUND.match(value) and VOWEL_SOUND.match(value):
         return 'An {}'.format(value)
     return 'A {}'.format(value)
+
+
+def parse_raw_data(data):
+    try:
+        # try to parse as json
+        return json.loads(data)
+    except json.decoder.JSONDecodeError:
+        try:
+            # try to parse as stringified object
+            return json.loads(data.replace("'", '"'))
+        except json.decoder.JSONDecodeError:
+            return {}
+
+
+def get_csv_values(csv_data):
+    values = []
+    csvreader = csv.reader(csv_data.split('\n'), delimiter=',')
+    for row in csvreader:
+        values = values + row
+    return values
+
+
+def get_single_contributor_field_values(item, fields):
+    data = parse_raw_data(item.raw_data)
+    for f in fields:
+        value = data.get(f['column_name'], None)
+        if value is not None:
+            f['value'] = value
+    return fields
+
+
+def get_list_contributor_field_values(item, fields):
+    data_values = get_csv_values(item.raw_data)
+    list_fields = get_csv_values(item.source.facility_list.header)
+    for f in fields:
+        if f['column_name'] in list_fields:
+            index = list_fields.index(f['column_name'])
+            f['value'] = data_values[index]
+    return fields
--- a/src/django/api/models.py
+++ b/src/django/api/models.py
@@ -24,7 +24,9 @@ from api.constants import FeatureGroups
 from api.countries import COUNTRY_CHOICES
 from api.oar_id import make_oar_id
 from api.constants import Affiliations, Certifications, FacilitiesQueryParams
-from api.helpers import prefix_a_an
+from api.helpers import (prefix_a_an,
+                         get_single_contributor_field_values,
+                         get_list_contributor_field_values)
 
 
 class ArrayLength(models.Func):
@@ -667,6 +669,12 @@ class FacilityListItem(PPEMixin):
         help_text='The cleaned address of the facility.')
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
+
+    @staticmethod
+    def post_save(sender, **kwargs):
+        instance = kwargs.get('instance')
+        if instance.facility is not None:
+            index_custom_text([instance.facility.id])
 
     def __str__(self):
         return 'FacilityListItem {id} - {status}'.format(**self.__dict__)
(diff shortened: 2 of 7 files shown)


src/django/api/helpers.py
```diff
@@ -20,3 +22,42 @@ def prefix_a_an(value):
     if not CONSONANT_SOUND.match(value) and VOWEL_SOUND.match(value):
         return 'An {}'.format(value)
     return 'A {}'.format(value)
+
+
+def parse_raw_data(data):
```
TaiWilkin, line 27: Moved these functions into the `helpers.py` file to allow them to be used in multiple locations

src/django/api/models.py
```diff
@@ -668,6 +670,12 @@ class Meta:
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
+    @staticmethod
+    def post_save(sender, **kwargs):
+        instance = kwargs.get('instance')
+        if instance.facility is not None:
```
TaiWilkin, line 676: If no facility has been assigned to the FacilityListItem, then there's no FacilityIndex row to add any custom text to, so we don't need to try to index it

src/django/api/models.py
```diff
@@ -2314,6 +2322,73 @@ class ExtendedField(models.Model):
...
+    # Get a list of active FacilityListItems for the given facilities.
+    # Only include list items where the contributors have searchable fields.
+    # Select the most recent item for each facility for each contributor.
+    items_filter = (Q(facility_id__in=facility_ids)
+                    & Q(source__contributor_id__in=contributor_ids)
+                    & Q(source__is_active=True)
+                    & Q(facilitymatch__is_active=True))
+    items = FacilityListItem.objects.filter(items_filter) \
+        .distinct('facility__id', 'source__contributor__id') \
+        .order_by('facility__id', 'source__contributor__id', '-created_at') \
```
TaiWilkin, line 2350: This very useful `distinct` with specific field names is apparently only possible when using PostgreSQL for the DB. The `order_by` is required and needs to include all fields listed in the `distinct`. By adding `-created_at` to the `order_by`, we sort by most recent first, and then when we grab the first unique `facility_id`+`contributor_id` combo we get the most recent active FacilityListItem for each facility for each contributor.

src/django/api/models.py
```diff
@@ -2314,6 +2322,73 @@ class ExtendedField(models.Model):
...
+    for item in items:
+        contributor_id = item.source.contributor.id
+
+        # Calculate a list of searchable fields for the item's contributor
+        if len(contributor_fields[contributor_id]) == 0:
+            contributor_fields[contributor_id] = fields.filter(
+                    Q(embed_config__contributor__id=contributor_id)) \
+                    .values_list('column_name', flat=True)
+
+        formatted_fields = [{'value': '', 'column_name': f} for f
```
TaiWilkin, line 2365: We have to reformat the fields for each facility instead of using the dictionary to prevent getting values from other facilities belonging to that contributor mixed in, but the dictionary still prevents requerying for the contributor fields.

jwalgran (approved):
This is an excellent piece of work. Great attention to detail on a very complex interaction. I was able to work through the testing instructions without issue and saw the index being updated correctly.

TaiWilkin:
Thank you for reviewing!

PR #1619, opened 2022-02-02
Handle country code geocoding errors

TaiWilkin:
## Overview

Submitted facilities are frequently throwing the country code error,
which indicates that the geocoding process has returned a result that
doesn't contain the country which the contributor has submitted for the
facility.

There are two primary instances where this is occurring:

1) The returned geocoding result is imprecise and contains no country
code at all. In this case, we will now accept this as a valid result.

2) There are multiple results returned, and although the top result
has an invalid code, another result contains the correct country code.
In this case, we will now select the result with the correct country
code. (If there are both a result with a correct country code and an
inexact result, we prefer the correct country code.)

We will still return a geocoding error if all results include
non-matching country codes.

Connects [Client #62](https://github.com/open-apparel-registry/open-apparel-registry-clients/issues/62)

## Notes

I added additional mock data for geocoding responses in the test suite. 

## Testing Instructions

* These directions assume that `./scripts/resetdb` has been run. 
* Run `./scripts/server` and upload facilities which previous threw geocoding errors.
[oar-country-geocode-errors.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/7989054/oar-country-geocode-errors.csv)
* Process the facilities.
```
```
* Confirm that the facilities have been successfully geocoded.
* Submit one or more of the addresses (with new names) via the API. Confirm that they have been successfully geocoded. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/geocoding.py
+++ b/src/django/api/geocoding.py
@@ -15,11 +15,11 @@ def create_geocoding_params(address, country_code):
     }
 
 
-def format_geocoded_address_data(data):
-    first_result, *_ = data["results"]
-
-    geocoded_point = first_result["geometry"]["location"]
-    geocoded_address = first_result["formatted_address"]
+def format_geocoded_address_data(data, result=None):
+    if result is None:
+        result = data['results'][0]
+    geocoded_point = result["geometry"]["location"]
+    geocoded_address = result["formatted_address"]
 
     return {
         "result_count": len(data["results"]),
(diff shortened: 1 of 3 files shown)


src/django/api/geocoding.py
```diff
@@ -38,12 +38,37 @@ def format_no_geocode_results(data):
     }
 
 
-def validate_country_code(data, country_code):
-    address_components = data["results"][0]['address_components']
-    for component in address_components:
-        if component['short_name'] == country_code:
-            return True
-    raise ValueError("Geocoding results did not match provided country code.")
+def find_valid_country_code(data, country_code):
```
caseycesari, line 41: Consider adding a short comment explaining what "valid" means in this context.

caseycesari (approved):
Working well. Nice, clean implementation function.

TaiWilkin:
Thank you for reviewing!

Issue #1618, opened 2022-02-02
Show facility_type and processing_type extended field values on the sidebar


jwalgran:
## Overview

Show facility_type and processing_type extended field values on the sidebar

### Describe the solution you'd like

- Use product type as a guide for how to list multiple values
- For field type, show the corresponding values the "matched_values" tuples.

PR #1624, opened 2022-02-03
Display processing and facility type

caseycesari:
## Overview

Show the processing type(s) and facility type(s) for a facility in the facility details sidebar.

Also, reorder the extended fields in the sidebar so that processing type and facility type are next to each other.

Connects #1618

## Demo

Screen Shot 2022-02-03 at 1 07 39 PM

## Testing Instructions

 - If you don't already have some locations with facility and processing types assigned, follow the instructions in #1616.
 - Select a facility on the map, and verify that the facility and processing types are displayed correctly.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Index custom text [#1607](https://github.com/open-apparel-registry/open-apparel-registry/pull/1607)
 - Create extended field values for processing/facility type [#1616](https://github.com/open-apparel-registry/open-apparel-registry/pull/1616)
 - Add batch_process tool [#1625](https://github.com/open-apparel-registry/open-apparel-registry/pull/1625)
+- Display processing and facility type [#1624](https://github.com/open-apparel-registry/open-apparel-registry/pull/1624)
 
 ### Changed
 
--- a/src/app/src/components/FacilityDetailSidebarExtended.jsx
+++ b/src/app/src/components/FacilityDetailSidebarExtended.jsx
@@ -84,8 +84,8 @@ const detailsSidebarStyles = theme =>
 const formatAttribution = (createdAt, contributor) =>
     `${moment(createdAt).format('LL')} by ${contributor}`;
 
-const formatIfList = value =>
-    Array.isArray(value) ? value.map(v => <li>{v}</li>) : value;
+const formatIfListAndRemoveDuplicates = value =>
+    Array.isArray(value) ? [...new Set(value)].map(v => <li>{v}</li>) : value;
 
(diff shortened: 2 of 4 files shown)


jwalgran:
Looking at this now.

jwalgran (commented):
When I tested a file with 2 processing types ([processing-type-2.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/7997456/processing-type-2.csv)) I saw duplicate facility types, which is expected based on our implementation

Screen Shot 2022-02-03 at 12 08 11 PM

II was able to resolve this by adding some logic and renaming our existing list formatting helper


```diff
diff --git a/src/app/src/components/FacilityDetailSidebarExtended.jsx b/src/app/src/components/FacilityDetailSidebarExtended.jsx
index bad608f..666e9b1 100644
--- a/src/app/src/components/FacilityDetailSidebarExtended.jsx
+++ b/src/app/src/components/FacilityDetailSidebarExtended.jsx
@@ -84,8 +84,8 @@ const detailsSidebarStyles = theme =>
 const formatAttribution = (createdAt, contributor) =>
     `${moment(createdAt).format('LL')} by ${contributor}`;
 
-const formatIfList = value =>
-    Array.isArray(value) ? value.map(v => <li>{v}</li>) : value;
+const formatIfListAndRemoveDuplicates = value =>
+    Array.isArray(value) ? [...new Set(value)].map(v => <li>{v}</li>) : value;
 
 /* eslint-disable camelcase */
 const formatExtendedField = ({
@@ -96,7 +96,7 @@ const formatExtendedField = ({
     id,
     formatValue = v => v,
 }) => ({
-    primary: formatIfList(formatValue(value)),
+    primary: formatIfListAndRemoveDuplicates(formatValue(value)),
     secondary: formatAttribution(updated_at, contributor_name),
     verified,
     key: id,

```

jwalgran (approved):
Other than the duplication this works as intended. If you like my suggested diff and include it, consider this approved.

caseycesari:
@jwalgran Good catch. I forgot that having two processing types will now add two facility types. I'll integrate your diff.

caseycesari:
@jwalgran I implemented your fix, and did some more testing. In the case when two facilities have been matched, and they have the same processing or facility type, this is what it will look like:

Screen Shot 2022-02-03 at 3 33 56 PM

Do we want to remove duplicate values across extended field type entries (the current code only does it within the same entry)?

A similar thing happens with product type as well:

Screen Shot 2022-02-03 at 3 37 09 PM

jwalgran:
Handling the cases when a contributor submits the same extended field values multiple times is something that we should address, but is outside the scope of this PR since, as you mentioned, it is not just related to the processing type and facility type fields. 

I created #1626 to follow up on this.

PR #1627, opened 2022-02-04
Update embed facility detail rendering to support extended fields

jwalgran:
## Overview

Embedded map owners have the ability to change the order of and labels for the fields they have submitted. We add support for this to extended fields and support for mixing the order of extended fields and other contributor fields by introducing a new `renderEmbedFields` method that references the embed config and calls the appropriate function to render either an extended field or an arbitrary contributor field.

Connects #1592

## Demo

Screen Shot 2022-02-04 at 3 01 33 PM

## Notes

Prior to the general availability of extended fields embedded maps had special handling for transparency pledge fields. As part of removing this special handling we rename "default" to "extended" and update the field keys to align with the extended field keys.

Note that although we have implemented support for native language name it is not part of the publicly acknowledged lists of extended fields that will be available when the extended profile feature is launched. As a result, we do not included it in `NonstandardField.EXTENDED_FIELDS`.

In order to show extended fields on an embedded map we need the ability to filter them so that only values added by the embedded map contributor appear.

We also remove handling for extended fields from the `get_contributor_fields` serializer method because the values are now retrieved as by the `get_extended_fields` serializer method.

When showing an embedded map we only show the most recent extended field values from the embed contributor. As such, we don't need to display the secondary line of information under the field value since it will always display the same contributor name.

## Testing Instructions

These instructions assume that`./scripts/resetdb` has been run

* Log in as c1@example.com, browse http://localhost:8081/admin/api/contributor/2/change/ and set the embed level to "Deluxe/Custom"
* Log in as c2@example.com and submit 
[test-3-fprocessing-types-and-custom-field.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8016856/test-3-fprocessing-types-and-custom-field.csv)
* Fully process the list with `./tools/batch_process 16`
* Browse http://localhost:6543/settings, click the Embed tab
* Enable the building color custom field  and move it so that it is between facility type and processing type.
* Set an embed size to load the embed preview
* In the embed preview text search for "test" and select the test facility. Verify that both the extended field values and the custom field values are displayed.

## Checklist

- [x] CI passes after rebase


src/app/src/util/embeddedMap.js
```diff
@@ -49,12 +49,13 @@ export const formatExistingFields = (fields = []) =>
 const doesExist = (field, existingFields) =>
     existingFields.some(f => f.columnName === field);
 
-// Keys in this object must be kept in sync with the NonstandardField.DEFAULT_FIELDS in django/api/models.py
+// Keys in this object must be kept in sync with the NonstandardField.EXENDED_FIELDS in django/api/models.py
```
jwalgran, line 52: Note that although we have implemented support for native language name it is not part of the publicly acknowledged lists of extended fields that will be available when the extended profile feature is launched. As a result, we do not included it in NonstandardField.EXTENDED_FIELDS.

src/django/api/serializers.py
```diff
@@ -831,13 +831,7 @@ def get_contributor_fields(self, facility):
...
             return []
 
-        # If the contributor has not created any overriding embed config these
-        # transparency pledge fields will always be visible.
-        fields = [
-            EmbedField(column_name=column_name, display_name=display_name)
-            for (column_name, display_name)
-            in NonstandardField.DEFAULT_FIELDS.items()]
-
+        fields = []
```
jwalgran, line 834: We remove handling for extended fields from the get_contributor_fields serializer method because the values are now retrieved as by the get_extended_fields serializer method.

TaiWilkin (commented):
This is working well. I also checked with multiple contributors for a facility's extended fields and outside of embed to confirm that the changes were working in the various modes. 
Once the tests are passing and it's been rebased to develop, it should be good to go

jwalgran:
Thanks for the review. Tests are passing. This just needs a final approval.

TaiWilkin (approved):
Good to go

PR #1631, opened 2022-02-08
Fix custom text processing

TaiWilkin:
## Overview

The custom text field was not being updated to empty when all of the
fields for that facility had been marked un-searchable. We now
correct remove the custom_text for the facility in that case.

## Testing Instructions

* On the branch `eh/custom_text_in_embedded_map`, upload custom text fields for a facility, and mark them searchable. 
* Search for the facility by the custom fields and note that it is found. 
* Mark the custom texts to be NOT searchable, then search for the facility again and note that it is still found. 
* Switch to this branch and update the fields' searchability again by making them searchable and then unsearchable. 
* The facility should now NOT be found when searching. 
* Mark the field searchable again and confirm that the field is now found. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - Handle country code geocoding errors [#1619](https://github.com/open-apparel-registry/open-apparel-registry/pull/1619)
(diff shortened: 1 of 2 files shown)


emilyhu0106 (approved):
Everything works well now! Thank you for identifying this issue!

TaiWilkin:
Thank you for reviewing!

Issue #1640, opened 2022-02-11
Ignore partial OAR ID matches in text search


TaiWilkin:
## Overview

When searching for a number, such as '12', in the OAR text search, all facilities with the number '12' in their OAR ID will be returned.  In general, users who want to match to an OAR ID will search for the entire ID, and not just a portion of the ID. This results in many more facilities being returned for numeric searched than users are expecting.

### Describe the solution you'd like

- Update the search query to only ignore partial matches when matching OAR IDs to text

TaiWilkin:
Searching for '12' before (on Staging):
![Screen Shot 2022-02-14 at 2.53.53 PM.png](https://images.zenhubusercontent.com/5e41c4f13da639155d2ee456/d8591089-30ca-4949-8698-85b72d267ba2)
Searching for '12' after (on Staging):
![Screen Shot 2022-02-15 at 10.35.23 AM.png](https://images.zenhubusercontent.com/5e41c4f13da639155d2ee456/1b5e577d-6d92-49b4-aa55-691ee755dc02)

PR #1641, opened 2022-02-11
Ignore partial OAR ids

TaiWilkin:
## Overview

When searching for a number, such as '12', in the OAR text search, all
facilities with the number '12' in their OAR ID will be returned. In
general, users who want to match to an OAR ID will search for the
entire ID, and not just a portion of the ID. This results in many more
facilities being returned for numeric searched than users are expecting.

We now only consider the query text to be a match to the OAR ID if it's
an exact match.

Working with exact matches revealed a related bug. 
When the search is run for the facilities list, if there is only one
match, the search bar automatically navigates to the facility. However,
when we navigate back to the list from the facility details, this
resulted in being navigated directly back to the facility details.

When rerunning the search in the back navigation for facility details,
replace the 'push' command with a noop to prevent navigation to a
specific facility.

Connects #1640 

## Demo

Results when searching for '5':
Screen Shot 2022-02-11 at 11 27 59 AM

## Notes

## Testing Instructions

- On `develop`:
    - Search for ‘5’ in the text search. You will find over 100 facilities. 
    - Navigate to a facility and copy the OAR ID. Enter the OAR ID in the text search and search. 
    - You will be navigated automatically to a matching facility. Click the back button, but you will not be navigated back to the facilities list.
- On this branch: 
    - Search for ‘5’ in the test search. You will now only find a few facilities (all with ‘5’ in their name).
    - Search for the same OAR ID as before. 
    - You will be navigated automatically to the matching facility. Click the back button, and you will be navigated to the facility list.
    - Repeat the search in an embedded map and confirm that general text search functionality behaves as before. 

## Checklist

- [ ] CI passes after rebase


emilyhu0106 (approved):
Very simple and tidy way to solve the issues! Ignoring the partial OAR ID will be very helpful to for clients to narrow down their search.

TaiWilkin:
Thank you for reviewing!

PR #1647, opened 2022-02-14
Enable remote Django debugging in Visual Studio Code

caseycesari:
## Overview

Allow developers to debug the Python back-end by using Visual Studio Code's debugging capabilities along with the [debugpy](https://github.com/microsoft/debugpy) library.

## Demo

Screen Shot 2022-02-14 at 1 04 52 PM

## Notes

- I tried to get this to work using the Gnuicorn server, but I was unable to. I think it has to do with they way Gnuicorn does multi-threading. I wasn't able to find many examples of people using Gnuicorn in development, let alone with debugpy. 
- This does not work if you are running the app within Vagrant. I was hoping forwarding the port would be enough, but that was not the case. Other attempted fixes also did not work.

## Testing Instructions

- Follow the instructions in the README and see if you can successfully trigger a breakpoint. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,18 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Debug Django",
+            "type": "python",
+            "request": "attach",
+            "pathMappings": [
+                {
+                    "localRoot": "${workspaceFolder}/src/django",
+                    "remoteRoot": "/usr/local/src"
+                }
+            ],
+            "port": 3000,
+            "host": "localhost",
+        }
+    ]
+}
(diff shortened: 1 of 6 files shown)


TaiWilkin (approved):
This is working in my development environment. Because this is run outside of Vagrant, I did have to run the `update` script outside of Vagrant to set up the environment before it would work. In addition, on my MacBook, Django for Mac needs to be running. Note as well that the Django admin is unavailable while using this tool, which means it can't replace our current workflow; however, for many tasks this will be extremely useful and save a lot of time, so as a flagged optional process it's going to be very helpful.

caseycesari:
@TaiWilkin Thanks for the review. I added some more context to the `README` based on your experience.

caseycesari:
Thanks for the reviews!

Issue #1587, opened 2022-01-14
Update FacilityIndex with extended field values


jwalgran:
## Overview

Update FacilityIndex with extended field values

### Describe the solution you'd like

- [ ]   Add columns for extended fields to the search index table
    -   String array fields for
        -   number_of_workers (this is a string array because we index by standardized ranges)
        -   facility_type
        -   processing_type
        -   product_type
        -   parent_company_name
        -   native_language_name
    -   Integer array field for parent_company_id
    -   Use `clean` taxonomy values in the index.
- [ ]   Update facility indexing function to populate extended field columns
    -   For number_of_workers add any of the standardized ranges that overlap with any of the submitted ranged
    -   Use `clean` on product_type values
    -   Use `unidecode` when indexing native_language_name

PR #1651, opened 2022-02-15
Index extended fields

TaiWilkin:
## Overview

In order to later enable searching by extended field, adds extended
field values to the facility index when the indexing function is run
and in response to relevant data changes, including splitting and
merging of facilities.

For number_of_workers, creates a string array including any of the
standardized ranges that overlap with any of the submitted range values.

For facility_type, creates a string array of all submitted
facility_types, including ones calculated based on submitted processing
types.

For processing_type, creates a string array of all submitted processing
types, excluding facility_types.

For product_type, creates a `clean()` string array based on submitted
procust_types.

For parent_company_name, creates a string array using the raw data name
for non-matched parent_company, but uses contributor_name for
matched parent_company.

For parent_company_id, creates an integer array using the
contributor_id for matched parent_company.

For native_language_name, creates a unidecoded string array from
submitted native_language_names.

Connects #1587 

## Demo

```
>>> FacilityIndex.objects.filter(name__in=['Alsico', 'ALFA TAPUTFABRIEK NV', 'Aarova vzw']).values('id', 'name', 'number_of_workers', 'facility_type', 'processing_type', 'product_type', 'parent_company_name', 'native_language_name', 'parent_company_id')
<QuerySet [
{'id': 'BE2022046WHE6D5', 'name': 'Aarova vzw', 'number_of_workers': ['Less than 1000', 'More than 10000', '1001-5000'], 'facility_type': ['Final Product Assembly', 'Warehousing / Distribution', 'Printing, Product Dyeing and Laundering'], 'processing_type': ['Extrusion', 'Collecting'], 'product_type': ['hats'], 'parent_company_name': ['ABC Holdingas'], 'native_language_name': ['Hang Zhou Wan Kai Fa Qu Xing Ci Er Lu Yu Bin Hai Er Lu Cha Kou Heng Yuan Gong Ye Yuan Qu '], 'parent_company_id': []}, 
{'id': 'BE20220460EGHZ5', 'name': 'ALFA TAPUTFABRIEK NV', 'number_of_workers': ['1001-5000'], 'facility_type': ['Final Product Assembly', 'Warehousing / Distribution'], 'processing_type': ['Stitching', 'Assembly', 'Cutting', 'Packing', 'Steaming'], 'product_type': ['short sleeved shirts', 'wovens', 'shorts', 'dresses', 'skirts', 'long sleeved shirts', 'jumpsuits', 'trousers'], 'parent_company_name': ['Brand A'], 'native_language_name': [], 'parent_company_id': [4]}, 
{'id': 'BE2022046NPJFZ2', 'name': 'Alsico', 'number_of_workers': ['Less than 1000'], 'facility_type': [], 'processing_type': [], 'product_type': [], 'parent_company_name': [], 'native_language_name': [], 'parent_company_id': []}
]>
```

## Testing Instructions

* Run `./scripts/manage migrate`
* Run `./scripts/server`. Login as `c2@example.com` and submit 
[OAR Extended Fields.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8073771/OAR.Extended.Fields.csv). Process with `./tools/batch_process {list_id}`
* In `./scripts/manage shell_plus`:
```
FacilityIndex.objects.filter(name__in=['Alsico', 'ALFA TAPUTFABRIEK NV', 'Aarova vzw']).values('id', 'name', 'number_of_workers', 'facility_type', 'processing_type', 'product_type', 'parent_company_name', 'native_language_name', 'parent_company_id')
```
* Compare the values of the indexed items to the values submitted in the CSV to confirm they match, all relevant values are included, and values aren't repeated. 
* Authenticated at http://localhost:8081/api/docs/#!/facilities/facilities_create and post:
```
{
"name": "ALFA TAPUTFABRIEK NV",
"address": "Lindestraat 96, B-8790 Waregem , Belgium",
"country": "Belgium",
"parent_company": "Brand A",
"processing_type": ["Cutting", "Stitching", "Steaming"]
}
```
* Run the same facility index query in `shell_plus` as before. Confirm that the new values from the API submission have been added to ALFA TAPUTFABRIEK NV.
* Claim 'Aarova vzw' as c2@example.com
* In another browser session, login as c1@example.com and approve the claim. 
* Update the claim details in 'My Facilities' as c2@example.com. Submit "杭州湾开发区兴慈二路与滨海二路叉口恒元工业园区" as the native language name (to test that a long unidecoded name works) in addition to filling the other claim fields that we process for extended fields.
* Run the same facility index query in `shell_plus` as before. Confirm that the new values from the claim have been added to Aarova vzw.
* Get the ID for ALFA TAPUTFABRIEK NV. Login as c1@example.com and navigate to `Adjust facility matches`. Enter the id and split ALFA TAPUTFABRIEK NV.
* Run the same facility index query in `shell_plus` as before. Confirm that there are now two facilities named ALFA TAPUTFABRIEK NV. One should contain the extended field values from the CSV item and the other should contain the extended field values from the API submission. 
* Navigate to 'Merge two facilities' and merge the two facilities named ALFA TAPUTFABRIEK NV. 
* Run the same facility index query in `shell_plus` as before. Confirm that there is now only one facility named ALFA TAPUTFABRIEK NV and that it contains both sets of extended field values. 

## Checklist

- [x] CI passes after rebase


src/django/api/helpers.py
```diff
@@ -61,3 +64,41 @@ def get_list_contributor_field_values(item, fields):
             index = list_fields.index(f['column_name'])
             f['value'] = data_values[index]
     return fields
+
+
+def clean(column):
```
TaiWilkin, line 69: Moved this to the `helpers` file to avoid circular dependencies when using this in the `models` file

src/django/api/tests.py
```diff
@@ -7845,7 +7877,10 @@ def setUp(self):
         super(ParentCompanyTestCase, self).setUp()
         self.url = reverse('facility-list')
 
-    def test_submit_parent_company_no_match(self):
+    @patch('api.geocoding.requests.get')
```
TaiWilkin, line 7880: Updated a bunch of the tests to mock the external geocoding API (preventing external dependency in tests)

src/django/api/views.py
```diff
@@ -1879,6 +1880,8 @@ def merge(self, request):
         # be made in the `facility_history.py` module's
         # `create_facility_history_dictionary` function
         merge.changeReason = 'Merged with {}'.format(target.id)
+
+        FacilityIndex.objects.get(id=merge.id).delete()
```
TaiWilkin, line 1884: When testing the extended field indexing, I discovered that FacilityIndex objects weren't being removed after a merge (when the related Facility is deleted). Added a line to remove the stale Index objects

src/django/api/helpers.py
caseycesari: @TaiWilkin Same! https://github.com/open-apparel-registry/open-apparel-registry/pull/1639#discussion_r804803932

src/django/api/constants.py
```diff
@@ -147,3 +147,22 @@ class DateFormats:
     SECOND = '%Y-%m-%d %H:%M:%S'
     MONTH = '%Y-%m'
     WEEK = '%Y-%W'
+
+
+class NumberOfWorkersRanges:
+    STANDARD_RANGES = [{
+        'min': 0,
+        'max': 1000,
+        'text': 'Less than 1000'
```
caseycesari, line 156: Minor, but consider renaming this property `label`, which I think makes the usage [here](https://github.com/open-apparel-registry/open-apparel-registry/pull/1651/files#diff-900f6cb047297d79255398d8083917fe71bcfc43e15e933ef8a322787fcf5508R103) a bit more clear.

src/django/api/helpers.py
```diff
@@ -82,3 +84,21 @@ def clean(column):
...
+    range_max = range.get('max', None)
+    value_min = range_value.get('min', 0)
+    value_max = range_value.get('max', value_min)
+    if value_min < range_min:
+        return value_max >= range_min
+    if value_min >= range_min:
+        return range_max is None or value_min <= range_max
+
+
+def convert_to_standard_ranges(number_of_workers, overlapping_ranges=set):
```
caseycesari, line 100: I think using `set` instead of `set()` here initializes `overlapping_ranges` to the class `set` instead of a new `set` instance:

```py
>>> def foo(bar=set):
...     return bar
...
>>> foo()
<class 'set'>
>>> def foo(bar=set()):
...     return bar
...
>>> foo()
set()
```

However, even with that fix, I think this could potentially not work as intended because the default value of `overlapping_ranges` is initialized once and then reused each time the function is called. 

```py
>>> def add_to_set(elements, foo=set()):
...     foo.add(elements)
...     return foo
...
>>> add_to_set(10)
{10}
>>> add_to_set(20)
{10, 20}
```

More info and a workaround here: https://nikos7am.com/posts/mutable-default-arguments/

caseycesari (changes requested):
This is working as described. Just a couple comments.

src/django/api/constants.py
TaiWilkin: Good call. I've updated the property name.

TaiWilkin:
Thank you for reviewing!

Issue #1638, opened 2022-02-10
Upload issues with Extended Fields


obrienad:
# Draft

1. People had to follow different work flows to upload the same file

- Mariel: format data in excel -> save as csv to a local folder -> upload to OAR -> "An error prevented uploading that file." 
- Natalie: format data in google sheets -> download as csv -> upload to OAR -> success 

- Moving excel CSV to Google sheets caused changes to formatting (headers became capitalized). Had to revert back to lowercase for it to be accepted. 

Opened the excel-formatted CSVs in Text Edit and saw the following
- In the [SanMar data](https://drive.google.com/file/d/1JSof38xjvwGqZY3s8Z4twTO702jJGXD0/view?usp=sharing), there were quotation marks around some of the values but not others. 
- In the [Scotch & Soda data](https://drive.google.com/file/d/1-O2Ph3N8EQayBWnghUNKgXcGMORKotys/view?usp=sharing), there were commas, usually four, at the end of some of the rows. Both data sets, as formatted in excel are attached. 
Despite those formatting, both of these data sets when uploaded to GS, then re-downloaded as CSV then uploaded, worked.

- Upload handling is much more sensitive than we expected it to be. This could be a huge challenge to users, who may use Excel or other applications to format their data. 
 
Issues with the processing_type and facility_type fields: 

- processing_type is not accepting any types that are not explicitly listed in the taxonomy. Our understanding is that processing_type should accept any data and do a fuzzy match against the taxonomy. Any matches with taxonomy would be displayed, any other values would only be available on download. 

- When the processing_type column is removed altogether an error for facility_type values of "Main Production Site or Subcontractor/Processor" resulted in an error. 

We understand that this is because, for facility_type, the intention is that system requires that it have an exact match in the taxonomy. 

Related: OAR has requested information on whether they need to change to processing_type & facility_type fields (see discussion in Slack).

TaiWilkin:
Potentially related Rollbar errors for reference:
https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/203/
https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/594/

obrienad:
Additional data uploads here: https://drive.google.com/drive/folders/162AQ5ryaGIeHKoIWAILq9jwt8ytjvreQ?usp=sharing

mariel-oar:
Hi @jwalgran, 
[...]

PR #1663, opened 2022-02-18
Fix handling of numeric cells in Excel workbooks

jwalgran:
## Overview

With the introduction of number_of_workers we now have a column that can have a numeric value. This presented 2 challenges to the existing Python code.

- Excel parsing library returns these as numeric types, but the code was expecting a string.
- A value that appears as 130 in the workbook can appear as a float 130.0 after parsing.

This commit attempts to fix both of these issues by first manually casing cell values to strings when parsing the workbook, then using an `int(parse(x))` nested cast in the `extract_int_range_value` function to ensure that the decimal portion is removed. 

Connects #1638

## Testing Instructions

* Check out `develop`, log in as c2@example.com, submit 
[extended-template-test.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/8110687/extended-template-test.xlsx). Verify that an error blocks submission.
* Check out `bugfix/jvw/handle-numeric-excel-cells` and attempt to submit the file again. Verify that it succeeds and that `./tools/batch_prcess {is}` completes successfully.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Fix handling of numeric cells in Excel workbooks [#1663](https://github.com/open-apparel-registry/open-apparel-registry/pull/1663)
+
 ### Security
 
 - Bump follow-redirects from 1.14.3 to 1.14.8 in /src/app [#1645](https://github.com/open-apparel-registry/open-apparel-registry/pull/1645)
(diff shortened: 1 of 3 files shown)


TaiWilkin (approved):
I tested this both with the provided list, and the list [Embedded.Map.CSV.Template.-.boohoo.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/issues/1638#issuecomment-1046911702) from the issue. Both lists broke on develop, but were able to be uploaded and processed successfully on this branch. 
Thanks for the clear write-up and code comments explaining the fix for this!

jwalgran:
Thank your for the quick review turnaround.

Issue #1652, opened 2022-02-16
Modify embed custom text search - Only one field and only exact match


jwalgran:
## Overview

Modify embedded custom text search - Only one field and only exact match

### Describe the solution you'd like

- Change searchable field selection to ratio buttons instead of checkbox. Only allow one selection
- Add migrations to change the `custom_text` field to an array string.
   - Use multiple migrations. Add a new column, reindex to set the values, drop the old colum, rename the new column to the old name.
- Update facility indexing. Sore free text searchable value as `{contributor_id}|{value}`
- Update searching of `custom_text` to build the appropriate `{contributor_id}|{value}` string and use an `iexact` rather than `icontains` on the array field.
- When updating the embed map config, only reindex if the searchable field has changed.

PR #1665, opened 2022-02-18
Modify custom_text handling

TaiWilkin:
## Overview

Contributors want search queries in their embedded map to only return
matches for custom text if it is an exact match to custom text which
they submitted.

We have updated the custom_text field to use an array of strings in the
format '{contributor_id}|{custom_text}' (For example, '1|abcd'). We
format the free text query in the search in the same way, and find exact
matches.

In our previous iteration of this feature, given Contributor 1 submitted
a custom text value of "ABCD" for Facility A and Contributor 2
submitted a custom text value of "ABCD" for Facility B, searching for
"D" in Contributor 1's map could return both Facility A and Facility B.

In the current version, searching "D" would return no matches; and
searching "ABCD" in Contributor 1's map would return only Facility A.

Custom text was being reindexed each time the embed config was updated,
which is a lengthy process because we need to reindex custom text for all 
facilities for the contributor each time. The embed config view has been updated to 
compare the previous searchable column_names to the new searchable column_names,
and only run if the searchable column_names have been updated. 

Connects #1652 

## Demo

Previous `custom_text`:
```
<QuerySet [{'custom_text': 'Test Value 2Test Field Value'}, {'custom_text': 'Test Value!'}]>
```
Updated `custom_text`:
```
<QuerySet [{'custom_text': ['3|test value 2', '3|test field value']}, {'custom_text': ['3|test value!']}]>
```

After preventing the custom_text from reindexing on every save, there is significant difference in speed
between updating when there has been a searchability change and when there has not:
Updating searchable fields (causes reindexing): 532.80 ms
Updating map style (does not cause reindexing): 52.45 ms

## Notes

We initially planned to limit contributors to one searchable field. However, due to using an array for the custom text, it doesn't add any additional complexity to the current design to allow multiple searchable fields, so we have decided not to limit the number of searchable fields at this time. 

## Testing Instructions

* On `develop`, sign in as c2@example.com, submit a list with custom fields [oar-extra-fields.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8099773/oar-extra-fields.csv) and fully process it. Assign deluxe embed permissions to c2@example.com.
* Visit c2@example.com's embed settings, and assign a height and width to the map. Make custom_field and custom_field_2 searchable. 
* Run `./scripts/manage shell_plus` and then `FacilityIndex.objects.distinct('custom_text').values('custom_text')`. You will see a list of custom text values. 
* Switch to this branch, and run `./scripts/manage migrate`. The migrations should succeed. 
* Run `./scripts/manage shell_plus` and then `FacilityIndex.objects.distinct('custom_text').values('custom_text')`. You will see a list of array-formatted custom text values. No values should have been lost. 
* Search for 'test' in c2@example.com's map. There should be no results. 
* Search for 'Test Value 2' or 'test value 2' in c2@example.com's map. There should be a result. 
* Login as c3@example.com, submit a list with custom fields 
[oar-extra-fields-alt.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8099793/oar-extra-fields-alt.csv), and fully process the it. Assign deluxe embed permissions to c3@example.com.
* Visit c3@example.com's embed settings, and assign a height and width to the map. Make custom_field and custom_field_2 searchable. 
* Run `./scripts/manage shell_plus` and then `FacilityIndex.objects.distinct('custom_text').values('custom_text')`. You will see a list of custom text values, which should now include the values from c3@example.com's list. 
* Search for 'Test Value 2' or 'test value 2' in c3@example.com's map. There should be no results. 
* Update the searchable fields and note the timing for submitting the request. Update the map style or another field unrelated to the search and note that the time is greatly reduced. 

## Checklist

- [x] CI passes after rebase


src/django/api/models.py
```diff
@@ -1253,13 +1254,14 @@ def filter_by_query_params(self, params):
...
         if free_text_query is not None:
+            custom_text = format_custom_text(contributors[0], free_text_query)
             if switch_is_active('ppe'):
                 if embed is not None:
                     facilities_qs = facilities_qs \
                         .filter(Q(name__icontains=free_text_query) |
                                 Q(id=free_text_query) |
                                 Q(ppe__icontains=free_text_query) |
-                                Q(custom_text__icontains=free_text_query))
+                                Q(custom_text__contains=[custom_text]))
```
emilyhu0106, line 1264: Why are we using `contains` instead of `exact` or `iexact` here for an exact match?

TaiWilkin:
> The only missing piece is the updating the front-end to make only one field to be searchable - using radio button instead of checkboxes.

After speaking to Justin, we've decided to drop the limitation of only being able to search by one field, so we don't need to change to radio button.

emilyhu0106 (approved):
Good job on this improvement and thank you for checking with Justin about the scope of this issue!

TaiWilkin:
Thank you for reviewing!

Issue #1672, opened 2022-02-23
List index out of range exception when running a free-text search


jwalgran:
## Overview

List index out of range exception when running a free-text search

### Steps to Reproduce

* Browse http://localhost:6543/
* Search for "test"

### Additional context

```
ERROR:django.request:Internal Server Error: /tile/facilitygrid/1645640734-1-84e391b0/2/3/2.pbf
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 505, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 465, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 476, in raise_uncaught_exception
    raise exc
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 502, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/decorators.py", line 50, in handler
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/views/decorators/cache.py", line 31, in _cache_controlled
    response = viewfunc(request, *args, **kw)
  File "/usr/local/lib/python3.7/site-packages/waffle/decorators.py", line 50, in _wrapped_view
    return view(request, *args, **kwargs)
  File "/usr/local/src/api/views.py", line 3641, in get_tile
    request.query_params, layer, z, x, y)
[...]

PR #1673, opened 2022-02-23
Fix custom text handling when not in embed mode

jwalgran:
## Overview

When not in embed mode just searching for free text was raising an exception because `contributors` was None in `filter_by_query_params`.

Connects #1652
Connects #1672 

## Testing Instructions

* Check out `develop`
* Search for `test` and verify that an exception prevents the search from working
* Check out this branch, run the same search, and verify that it works.
* Verify that text search in embed mode still works

## Checklist

- [x] CI passes after rebase


TaiWilkin (approved):
Thanks for the quick fix! This is working as expected now

jwalgran:
Thanks for the review.

Issue #1685, opened 2022-03-01
Claim showing incorrect parent company


TaiWilkin:
## Overview

When viewing a claim, the selected `facility_parent_company` is not shown in the dropdown. Selecting and saving the claim with the correct contributor doesn't fix the problem. 

This appears to be a mixup between Contributor.id and User.id. For example, when saving 'Tai Wilkin' as the parent company (admin user id: 835, contributor id: 830) the value returned for `facility_parent_company` is { id: 835, name: 'Tai Wilkin' } but the displayed value in the selector is jwalgran+2021-03-08@azavea.com (contributor id: 835). 

### Expected Behavior

The dropdown should show the correct value, and whether we are using the contributor or user id for parent company values should be consistent throughout the site. 

### Actual Behavior

The incorrect parent company is shown in the dropdown.

### Steps to Reproduce

* Claim a facility and assign a parent company. Save and refresh the page
* The claim parent company will now show the incorrect value.

### Additional context

The correct parent company shows in the facility details, so these values are being handled differently there vs in the claim.

PR #1687, opened 2022-03-02
Fix claim showing incorrect parent facility

TaiWilkin:
## Overview

The claim facility form was correctly submitting and saving the
contributor name and contributor id; however, when serializing the data
for fetching it, we were incorrectly grabbing the user id. This caused
the form to display the incorrect contributor in the dropdown.

Claims now correctly serialize the parent_facility with the contributor
id, not the admin user id.

We have run into similar bugs many times, because in development, the 
default contributor and user IDs match. This has
caused us to miss a number of bugs related to mixing up contributors
and users when working with data (because these two entities are used
somewhat interchangeably). In addition to updating the claim serialization,
the IDs of contributors and their associated users have been updated in our 'fixtures' files,
and now will no longer match after running `resetdb`, which will
hopefully prevent this category of bug.

Connects #1685 

## Demo

Screen Shot 2022-03-02 at 11 24 56 AM

## Testing Instructions

* Checkout this branch, then checkout commit 1073cd3. Run `./scripts/resetdb`. 
* Navigate to a claim and save a parent company using the form, then refresh the page. Note that the displayed parent company is not the one you saved. 
* View the same claim in the Django admin and confirm that the stored parent company is correct.
* Checkout commit 4ae4282 view the same claim's claim form. Confirm that the displayed parent company is correct. 
* Resave with a new parent company and refresh the page and confirm that the correct parent company is saved and displayed. 

## Checklist

- [ ] CI passes after rebase


emilyhu0106 (approved):
Great fix! The two commits help streamline the review logic a lot.

src/django/api/fixtures/contributors.json
```diff
@@ -1,7 +1,7 @@
 [
     {
         "model": "api.contributor",
-        "pk": 2,
+        "pk": 1,
```
emilyhu0106, line 4: Just curious about some context here -  is `pk` the user ID and `fields.admin` being the contributor ID?

TaiWilkin: `pk` stands for "primary key". Since it's on the api.contributor model, it's the primary key / id for the contributor. The admin id is the id for the user who administers the account (so the user id). Hope that makes sense!

TaiWilkin:
Thank you for reviewing!

Issue #1703, opened 2022-03-09
Update to processing of data submitted in facility_type_processing_type


mariel-oar:
### Summary

Contributors are beginning to upload data as part of Extended Profiles, and in many cases they are running into errors in their list uploads related to having unsupported data values in the 'facility_type_processing_type' field. 

### Is your feature request related to a problem? Please describe.

This is a confusing experience because for other fields, such as product type, we accept any value, even if it does not get processed and matched to a product type in our taxonomy. The idea is that facility_type_processing_type should work similarly. 

### Describe the solution you'd like

1. All data submitted as part of facility_type_processing_type should be accepted on data upload 
2. facility_type_processing_type should then be processed to identify matches between the uploaded values and facility type taxonomy 
3. All values submitted in facility_type_processing_type that do not match to a facility type should be assigned to (and display) as part of the processing type field. 

#### Technical implementation

- Do not raise a validation error If there is no match for a facility type or processing type
- Continue to set `raw_values` as before
- For non matching items, set the`matched_values` item to [null, null, null, null]. This ensures that we can determine which raw values did and did not match.
- Update any UI code that renders `matched_values` to show raw values when the matched values are null

PR #1705, opened 2022-03-10
Allow submitting non-taxonomy facility type and processing type

jwalgran:
## Overview

We previously returned a validation error to contributors submitting facility types and processing types that did not match the taxonomy. When we began testing with contributors the poor user experience and friction this added to the process made us decide to accept and display non-matching values.

Connects #1703

## Demo

Screen Shot 2022-03-10 at 3 01 23 PM

## Testing Instructions

* TODO

## Checklist

- [ ] CI passes after rebase


TaiWilkin (approved):
This looks good! Thanks for updating the tests. 
It feels a bit strange to have unmatched values in the facility details alongside matched values. It would be good to have some kind of visual signifier of matched vs unmatched values in the UI. I think that's a design enhancement for the future and not in scope for this task, however.

Issue #1706, opened 2022-03-11
DRAFT: Add missing API endpoint


obrienad:
# Draft

Create endpoint that returns product_type options for the extended profile.

## Overview

A clear description of the feature request.

### Additional Context

Optional. Add any other context, screenshots, gifs, etc about the feature request here.

PR #1708, opened 2022-03-11
Add an API to return suggested product type choices

jwalgran:
## Overview

We allow contributors to submit any value for product type but expect that by suggesting standard values and values that have already been submitted we will have more standardization in the collected data.

We use a helper function so we can share the logic with a public API and with the claim form serializer which was already pulling a list of values. Now that we have migrated claims to use extended fields we can replace the old logic used to look for product types used on other claims.

Connects #1706

## Demo

Screen Shot 2022-03-11 at 1 34 46 PM

## Testing Instructions

* Verify that the new product-types API is available at http://localhost:8081/api/docs/#!/product-types/product_types_list and that it returns a list of unique types.
* Log in as c2@example.com and submit a claim
* In a separate browser log in as c1@example.com, browse http://localhost:8081/dashboard/claims/  and approve the claim
* In the original session browse http://localhost:6543/claimed, edit the claim, and verify that the product types dropdown lists is filled with values.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Filter by extended fields [#1696](https://github.com/open-apparel-registry/open-apparel-registry/pull/1696)
 - Add extended profile search APIs [#1697](https://github.com/open-apparel-registry/open-apparel-registry/pull/1697)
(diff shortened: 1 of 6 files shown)


jwalgran:
There is a test of the claim form product types that I will fix up

jwalgran:
Thanks for the review.

Issue #1711, opened 2022-03-15
Inactive Extended Profile Data Points are Displaying on Facility Pages 


mariel-oar:
## Overview
Per current behavior, when a facility's connection with a stakeholder is made inactive, the connection to that stakeholder is anonymized (e.g. BrandName changes to "A brand or retailer" on facility page). 

(Note: a connection is made inactive when the stakeholder uploads a new list that no longer includes the facility in question and makes a prior list that does include the facility inactive) 

## Description
When this occurs: 
1) All data points associated with that contributor should become anonymized. This includes country, name, and address data in addition to an extended field data points contributed by the user. 
2) All data points should remain on the facility profile with anonymized attribution (e.g. Number of workers: 130 contributed by a brand or retailer) 
3) All data points (active and inactive) should be available via download 

### Additional Context

We are seeing that in staging, when a list is deactivated that the connection to the original contributor is not anonymized.

Issue #1691, opened 2022-03-03
Nirapon list issue 


mariel-oar:
Summary: 
The brand Nirapon uploaded a facility list with a total of 481 facilities to the OAR; however when you search for them on the public side of OAR, they show up as having 482 facilities. 

Additional context: 
They previously had uploaded a list and not completed confirm/reject. That list is not active and we have not split any of their facilities through manual moderation. 

You can see they have 481 facilities in their upload here:https://openapparel.org/lists/1482
You can see the 482 facilities here: https://openapparel.org/?contributors=2139

The original list that they uploaded: 
[OAR_Contributor_Nirapon CAP Closed_Feb 2022_481 Factories.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/8178792/OAR_Contributor_Nirapon.CAP.Closed_Feb.2022_481.Factories.xlsx)

## Techical Implementation

- Review the research in https://docs.google.com/spreadsheets/d/1lz3BLLY6zM2MUOoGnA51hVj1ZEijUivs/edit?usp=sharing&ouid=104202036790313999359&rtpof=true&sd=true
- Ensure that facilities are reindexed when there is a list replacement.
- Reindex the Nirapon facilities in production

TaiWilkin:
When we search by Nirapon, we return 482 facilities. However, when we search by the [Nirapon CAP Closed_Feb 2022](https://openapparel.org/?contributors=2139&lists=1482) list, we only return 474. I'm attaching a csv of the "stale" items that aren't linked to a list for future reference (in case the data state changes by the time we are able to look into this). 
[nirapon.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8179136/nirapon.csv)
My suspicion is that we aren't correctly reindexing some facilities when we deactivate their list, but this will require additional research time to confirm.

hlennett:
Did some investigation into discrepancies between the list that Nirapon uploaded and what's available via search/download in their list via the OAR. You can see the results/notes here: https://docs.google.com/spreadsheets/d/1lz3BLLY6zM2MUOoGnA51hVj1ZEijUivs/edit?usp=sharing&ouid=104202036790313999359&rtpof=true&sd=true

mariel-oar:
@jwalgran @TaiWilkin , do you think it's worth having Nirapon upload their list again to see if this is still happening? I'm wondering if this could somehow have been affected by the other bug tickets that were closed last sprint. 

[...]

PR #1712, opened 2022-03-15
Hide inactive contributor associations

TaiWilkin:
## Overview

The ExtendedField serializer was incorrectly returning non-anonymized
contributor names and ids for contributors who had deactivated the
list through which they contributed the fields.

Deactivating a list or match will now result in the contributor's
details being anonymized in the extended field serializer.

Connects #1711 

## Demo

Screen Shot 2022-03-15 at 2 12 41 PM
Screen Shot 2022-03-15 at 2 12 49 PM

## Notes

My initial thought about this bug was that it was related to issue #1691, and possibly also extended field indexing. 
However, we handle anonymization of association displays in the serializers, not in the indexer, so this bug was unrelated to search or indexing.

## Testing Instructions

* Login as c2@example.com and submit [ExtendedFieldsTestListA.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8256370/ExtendedFieldsTestListA.csv). Fully process the list with `./tools/batch_process {list_id}`
* Open both facilities from the list in a new tab and confirm that they both display Service Provider A for all their fields and under contributors.
* Replace ExtendedFieldsTestListA with [ExtendedFieldsTestListB.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8256372/ExtendedFieldsTestListB.csv) (via submission form) and fully process the list with `./tools/batch_process {list_id}`.
* Refresh both facilities. 'Kept Facility' should still display Service Provider A for its contributor data. 'Deleted Facility' should now show 'a Service Provider' (an anonymized value) for the contributor details.

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Hide inactive contributor associations [#1712](https://github.com/open-apparel-registry/open-apparel-registry/pull/1712)
+
 ### Security
 
 ## [60] - 2022-03-03
--- a/src/django/api/serializers.py
+++ b/src/django/api/serializers.py
@@ -1481,19 +1481,30 @@ class ExtendedFieldListSerializer(ModelSerializer):
                   'contributor_name', 'contributor_id', 'value_count',
                   'is_from_claim', 'field_name', 'verified_count')
 
-    def get_contributor_name(self, instance):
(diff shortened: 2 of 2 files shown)


jwalgran (approved):
The implementation looks good. I repeated the test steps and also verified that manually togging the `is_active` or `is_public` flags on the `Source` via the Django admin also works correctly.

TaiWilkin:
Thanks for reviewing!

Issue #1694, opened 2022-03-03
Remove "workers" suffic from number of workers extended field values in sidebar


jwalgran:
## Overview

Remove "workers" suffic from extended field value in sidebar

### Describe the solution you'd like

<img width="330" alt="Screen Shot 2022-03-03 at 1 33 51 PM" src="https://user-images.githubusercontent.com/17363/156648441-3199f4f9-2a57-46ff-8f2f-58382dd4b5d1.png">

PR #1715, opened 2022-03-16
Remove the redundant workers suffix when showing value

jwalgran:
## Overview

The label above the field already says "Number of workers" so the OAR team asked to remove the redundant suffix, making "1000 wokers" into "1000."

Connects #1694

## Demo

Screen Shot 2022-03-15 at 5 58 28 PM

## Testing Instructions

* Log in as c2@example.com
* Upload [Extended-WorkersTest.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/8258153/Extended-WorkersTest.xlsx) and use `./tools/batch_process {list id}` to process it
* Browse both facilities and verify that the worker count does not have a suffix

## Checklist

- [x] CI passes after rebase


jwalgran:
Thanks for the review.

PR #1716, opened 2022-03-16
Move exact_match sorting to server

TaiWilkin:
## Overview

When a pending facility list item has multiple exact matches in the
database, we choose the match to automatically pair with via sorting
the matches and choosing the top result. We prefer active items, then
items which were submitted by the current contributor, then by the
most recent submission.

Previously, we sorted exact_matches using Django annotations and
order_by. This included a CPU-intensive nested subquery to collect a
list of active list item ids.

Moving the sorting to a server-side
sort allows us to perform the subquery only once, outside of the main
query, reducing the time taken by the query.

## Demo

I uploaded a large list three times using the old query. On the third time, once there were multiple matches for each item, I recorded the length of time taken by the full exact_match_items function and found it took 12.731543 seconds for matching all items. I also ran EXPLAIN ANALYZE in the dbshell and found that the execution time would be 5.394 ms for just the query (per item).

I then uploaded the same list again using the new query. The full exact_match_items function took 3.633751 seconds for matching all items in the list. EXPLAIN ANALYZE estimated that the execution time would be 0.123 ms for just the query (per item). 

OLD query:
```
                                                                                                                                                              QUERY PLAN                                    
                                                                                                                          
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=1503.05..1503.06 rows=1 width=2201) (actual time=5.348..5.350 rows=2 loops=1)
   Sort Key: ((hashed SubPlan 1)) DESC, ((api_source.contributor_id = 1)) DESC, api_facilitylistitem.updated_at
   Sort Method: quicksort  Memory: 29kB
   ->  Nested Loop  (cost=1486.72..1503.04 rows=1 width=2201) (actual time=5.315..5.336 rows=2 loops=1)
         ->  Index Scan using api_fli_match_fields_idx on api_facilitylistitem  (cost=0.28..8.32 rows=1 width=2199) (actual time=0.016..0.022 rows=2 loops=1)
               Index Cond: (((country_code)::text = 'VN'::text) AND (clean_name IS NOT NULL) AND (clean_name IS NOT NULL) AND ((clean_name)::text = 'tng investment and trading jsc padding factory branch':
:text) AND ((clean_address)::text = 'area b song cong industrial zone bach quang ward song cong city thai nguyen'::text))
               Filter: ((facility_id IS NOT NULL) AND ((status)::text = ANY ('{MATCHED,CONFIRMED_MATCH}'::text[])))
               Rows Removed by Filter: 1
         ->  Index Scan using api_source_pkey on api_source  (cost=0.15..8.17 rows=1 width=8) (actual time=0.003..0.003 rows=1 loops=2)
               Index Cond: (id = api_facilitylistitem.source_id)
         SubPlan 1
           ->  Hash Join  (cost=353.17..1479.02 rows=2907 width=4) (actual time=1.458..5.023 rows=924 loops=1)
                 Hash Cond: (u1.id = u0.facility_list_item_id)
                 ->  Hash Join  (cost=25.11..1095.17 rows=3563 width=4) (actual time=0.014..3.333 rows=2633 loops=1)
                       Hash Cond: (u1.source_id = u2.id)
                       ->  Seq Scan on api_facilitylistitem u1  (cost=0.00..1051.26 rows=7126 width=8) (actual time=0.004..2.844 rows=6050 loops=1)
                       ->  Hash  (cost=19.30..19.30 rows=465 width=4) (actual time=0.006..0.006 rows=18 loops=1)
                             Buckets: 1024  Batches: 1  Memory Usage: 9kB
                             ->  Seq Scan on api_source u2  (cost=0.00..19.30 rows=465 width=4) (actual time=0.002..0.004 rows=18 loops=1)
                                   Filter: is_active
                                   Rows Removed by Filter: 4
                 ->  Hash  (cost=255.38..255.38 rows=5814 width=4) (actual time=1.425..1.426 rows=4319 loops=1)
                       Buckets: 8192  Batches: 1  Memory Usage: 216kB
                       ->  Seq Scan on api_facilitymatch u0  (cost=0.00..255.38 rows=5814 width=4) (actual time=0.003..1.014 rows=4319 loops=1)
                             Filter: (is_active AND ((status)::text = ANY ('{AUTOMATIC,CONFIRMED,MERGED}'::text[])))
                             Rows Removed by Filter: 24
 Planning Time: 0.826 ms
 Execution Time: 5.394 ms
(28 rows)
```

NEW query: 

```
                                                                                                                                                         QUERY PLAN                                
                                                                                                                          
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------------------------------------
 Index Scan using api_fli_match_fields_idx on api_facilitylistitem  (cost=0.28..8.57 rows=1 width=2193) (actual time=0.094..0.104 rows=4 loops=1)
   Index Cond: (((country_code)::text = 'VN'::text) AND (clean_name IS NOT NULL) AND ((clean_name)::text = 'tng investment and trading jsc padding factory branch'::text) AND (clean_address IS NOT
 NULL) AND ((clean_address)::text = 'area b song cong industrial zone bach quang ward song cong city thai nguyen'::text))
   Filter: ((facility_id IS NOT NULL) AND ((status)::text = ANY ('{MATCHED,CONFIRMED_MATCH}'::text[])))
 Planning Time: 0.235 ms
 Execution Time: 0.123 ms
(5 rows)
```

## Testing Instructions

ON DEVELOP:
* Run `./scripts/server` and upload 
[1569-nodupes-latlng.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8274806/1569-nodupes-latlng.csv). This list contains lat and lng columns, which allows us to skip the geocoding process. 
* Run `./tools/batch_process {list_id}`.
* Re-upload and process the list several times. Note the timing of the match step.

ON THIS BRANCH:
* Re-upload and process the list several times. Note the timing of the match step.
* Confirm there is no functionality regression by completing the tests from https://github.com/open-apparel-registry/open-apparel-registry/pull/1568 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/matching.py
+++ b/src/django/api/matching.py
@@ -9,7 +9,7 @@ from datetime import datetime
 from django.conf import settings
 from django.contrib.postgres.search import TrigramSimilarity
 from django.db import transaction
-from django.db.models import Q, Max, ExpressionWrapper, BooleanField
+from django.db.models import Q, Max
 
 from api.models import (Facility,
                         FacilityList,
@@ -90,6 +90,18 @@ def get_canonical_items():
     return items
 
 
+def sort_exact_matches(exact_matches, active_item_ids, contributor):
+    def sort_order(f):
+        is_active = f['id'] in active_item_ids
+        has_same_contributor = f['source__contributor_id'] == contributor.id
+        updated_at = f.get('updated_at', None)
+        return (is_active,
+                has_same_contributor,
+                updated_at)
+
+    return sorted(exact_matches, key=sort_order, reverse=True)
+
+
 def exact_match_items(messy, contributor):
     started = str(datetime.utcnow())
 
@@ -103,7 +115,7 @@ def exact_match_items(messy, contributor):
                             FacilityMatch.MERGED],
                 is_active=True,
                 facility_list_item__source__is_active=True) \
-        .values_list('facility_list_item')
+        .values_list('facility_list_item', flat=True)
 
     results = dict()
 
@@ -113,24 +125,21 @@ def exact_match_items(messy, contributor):
         country_code = item.get('country', '').upper()
         empty_text_fields = Q(Q(clean_name__isnull=True) |
                               Q(clean_name__exact='') |
-                              Q(clean_name__isnull=True) |
-                              Q(clean_name__exact=''))
+                              Q(clean_address__isnull=True) |
+                              Q(clean_address__exact=''))
         exact_matches = matched_items.filter(clean_name=clean_name,
                                              clean_address=clean_address,
                                              country_code=country_code) \
             .exclude(empty_text_fields) \
-            .annotate(is_active=ExpressionWrapper(
-                Q(id__in=active_item_ids),
-                output_field=BooleanField())) \
-            .annotate(has_same_contributor=ExpressionWrapper(
-                Q(source__contributor=contributor),
-                output_field=BooleanField())) \
-            .order_by('-is_active',
-                      '-has_same_contributor',
-                      'updated_at') \
-            .values('facility_id')
+            .values('id', 'facility_id', 'source__contributor_id',
+                    'updated_at')
 
         if len(exact_matches) > 0:
+            if len(exact_matches) > 1:
+                exact_matches = sort_exact_matches(exact_matches,
+                                                   active_item_ids,
+                                                   contributor)
+
             results[messy_id] = exact_matches
 
     finished = str(datetime.utcnow())
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - Hide inactive contributor associations [#1712](https://github.com/open-apparel-registry/open-apparel-registry/pull/1712)
+- Move exact_match sorting to server [#1716](https://github.com/open-apparel-registry/open-apparel-registry/pull/1716)
 
 ### Security
 
--- a/src/django/api/tests.py
+++ b/src/django/api/tests.py
@@ -37,7 +37,8 @@ from api.models import (Facility, FacilityList, FacilityListItem,
                         ContributorManager, index_custom_text)
 
 from api.oar_id import make_oar_id, validate_oar_id
-from api.matching import match_facility_list_items, GazetteerCache
+from api.matching import (match_facility_list_items, GazetteerCache,
+                          sort_exact_matches)
 from api.processing import (parse_facility_list_item,
                             geocode_facility_list_item,
                             reduce_matches, is_string_match,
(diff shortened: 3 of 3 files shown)


src/django/api/matching.py
```diff
@@ -90,6 +90,18 @@ def get_canonical_items():
     return items
 
 
+def sort_exact_matches(exact_matches, active_item_ids, contributor):
+    def sort_order(f):
+        is_active = f['facility_id'] in active_item_ids
```
jwalgran, line 95: This comparison looks like it may be incorrect. If I am reading the code property, `active_item_ids` is a list of `FacilityListItem` ids. A `Facility` id read from `f['facility_id']` will never be ina list of `FacilityListItem` ids.

jwalgran (commented):
Thanks for working on this. Here is what I found.

I used `docker stats` to watch the CPU usage of the database container. and processed the large test list with `time ./tools/batch_process {id}`

The first 4 runs were on develop. On runs 3 and 4 the CPU would be in the 70s or 80s peaking at just over 100.

```
run 1
real	3m15.261s
user	0m0.699s
sys	0m0.495s

run 2
real	3m0.887s
user	0m0.565s
sys	0m0.450s

run 3
real	5m9.020s
user	0m0.624s
sys	0m0.472s

run 4
real	6m6.744s
user	0m0.691s
sys	0m0.530s
```

I then ran the file through twice more with this branch checked out. The total run times increased, but I did not see the database container CPU percent get above 40 percent

```
run 5 (on the branch)
real	7m19.256s
user	0m0.758s
sys	0m0.520s

Run 6 (on the branch)
real	9m20.718s
user	0m0.815s
sys	0m0.642s
```
While running through the test instructions in #1568 I had trouble at the end when I tried deactivating the matches to the adjusted facility. My submissions kept being linked to the wrong facility. As I worked through the code I found what looks like a bug, but when I applied the fix below the list match that sorted to the top was to wrong facility.

```diff
diff --git a/src/django/api/matching.py b/src/django/api/matching.py
index e388a588..734528c0 100644
--- a/src/django/api/matching.py
+++ b/src/django/api/matching.py
@@ -92,7 +92,7 @@ def get_canonical_items():
 
 def sort_exact_matches(exact_matches, active_item_ids, contributor):
     def sort_order(f):
-        is_active = f['facility_id'] in active_item_ids
+        is_active = f['id'] in active_item_ids
         has_same_contributor = f['source__contributor_id'] == contributor.id
         updated_at = f.get('updated_at', None)
         return (is_active,
@@ -131,7 +131,7 @@ def exact_match_items(messy, contributor):
                                              clean_address=clean_address,
                                              country_code=country_code) \
             .exclude(empty_text_fields) \
-            .values('facility_id', 'source__contributor_id', 'updated_at')
+            .values('id', 'facility_id', 'source__contributor_id', 'updated_at')
 
         if len(exact_matches) > 0:
             if len(exact_matches) > 1:
```  

We should pair and talk about these results and if the problem is with my testing technique.

jwalgran:
I think I found another small bug in the sorting and I believe this allows me to run through the test steps correctly.

By adding `print` statement I found that `active_item_ids` was returning tuples like `(12302,)`. This meant there were never any matches when trying to find the `FacilityListItem` id in the list. Adding `flat=True` appears to correct the issue.

```diff
diff --git a/src/django/api/matching.py b/src/django/api/matching.py
index e388a588..d4637968 100644
--- a/src/django/api/matching.py
+++ b/src/django/api/matching.py
@@ -92,7 +92,7 @@ def get_canonical_items():
 
 def sort_exact_matches(exact_matches, active_item_ids, contributor):
     def sort_order(f):
-        is_active = f['facility_id'] in active_item_ids
+        is_active = f['id'] in active_item_ids
         has_same_contributor = f['source__contributor_id'] == contributor.id
         updated_at = f.get('updated_at', None)
         return (is_active,
@@ -115,7 +115,7 @@ def exact_match_items(messy, contributor):
                             FacilityMatch.MERGED],
                 is_active=True,
                 facility_list_item__source__is_active=True) \
-        .values_list('facility_list_item')
+        .values_list('facility_list_item', flat=True)
 
     results = dict()
 
@@ -131,7 +131,7 @@ def exact_match_items(messy, contributor):
                                              clean_address=clean_address,
                                              country_code=country_code) \
             .exclude(empty_text_fields) \
-            .values('facility_id', 'source__contributor_id', 'updated_at')
+            .values('id', 'facility_id', 'source__contributor_id', 'updated_at')
 
         if len(exact_matches) > 0:
             if len(exact_matches) > 1:
```

TaiWilkin:
I added the suggested changes and also ran through the `time` and `docker stats` tests. 
For the `time` I'm seeing significantly smaller results than you, interestingly. 
```
real	2m51.825s
user	0m0.593s
sys	0m0.126s
```
For `docker stats`, I see `vagrant_database_1` not going above 45% at the highest, but `vagrant_django_run_5` is staying pretty high (up into the 90s at the highest I saw). I'm not sure whether or not that is a concern. 
(I haven't re-run through the manual sorting tests yet - going to do that next).

jwalgran (approved):
I would expect the higher CPU usage on the Djando contain since we moved our looping there. Because this will be running in a dedicated batch task container, maxing the CPU is ok, since that is the only job it is expected to do.

Issue #1700, opened 2022-03-08
Additional data points missing headers on download 


mariel-oar:
## Overview

When a user downloads additional fields data, in this case from an embedded map account, the additional fields (parent company, number of workers, product type, etc.) are missing column headers. 

### Expected Behavior

The additional data points should include column headers when the data is downloaded. 

### Steps to Reproduce

- Go to the embedded map test site for a contributor that has additional field data 
- download data, results in: https://open-apparel-registry.github.io/embeded-map-test-fixed-size.html?host=openapparel.org&contributors=2340

PR #1720, opened 2022-03-17
Fix CSV download when Extended Profile switch is off

TaiWilkin:
## Overview

When the extended profile switch is turned off, we were filtering out
extended field headers from the embedded map download but ttill
showing the extended field values, causing a mismatch between the
header and rows.

Elsewhere in the embedded map, we treat the extended fields as custom
contributor fields if extended profile is turned off. We now handle both
extended field headers and values as custom contributor fields in the
embedded map facilities download.

Connects #1700 

## Demo

Embedded map download when the extended profile flag is on:
[ExtendedProfileDownload.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8289918/ExtendedProfileDownload.csv)

Embedded map download when the extended profile flag is off:
[NoExtendedProfileDownload.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8289922/NoExtendedProfileDownload.csv)

## Testing Instructions

* As user c2@example.com, upload and process a list with both extended fields and custom fields. 
* Navigate to http://localhost:8081/admin/ as c1@example.com and turn off the extended profile switch. 
* Navigate to http://localhost:6543/settings as c2@example.com and search for the list you just uploaded in the embedded map preview. Download the list as a csv. The list should display the extended map fields as contributor fields. 
* Navigate to http://localhost:8081/admin/ as c1@example.com and turn on the extended profile switch. 
* Navigate to http://localhost:6543/settings as c2@example.com and search for the list you just uploaded in the embedded map preview. Download the list as a csv. The list should display the extended map fields as extended fields. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - Hide inactive contributor associations [#1712](https://github.com/open-apparel-registry/open-apparel-registry/pull/1712)
+- Fix CSV download when extended profile switch is off [#1720](https://github.com/open-apparel-registry/open-apparel-registry/pull/1720)
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


emilyhu0106 (approved):
Everything works great! Special kudo to the detailed explanation on why this one line change work for this issue

TaiWilkin:
Thank you for reviewing!

PR #1731, opened 2022-03-18
Delay loading search field dropdown data

TaiWilkin:
## Overview

As part of our redesign of the sidebar to support extended fields we researched search behavior and hid list frequently used search fields. If the search fields are never shown, their dropdown list option do not need to be fetched.

Connects #1729

## Notes

When refreshing the page after selection both a facility type and a processing type, the facility type options do not populate. This is not a regression and is a non-ideal behavior in the current implementation of filtering options based on the connection between facility type and processing type.  

## Testing Instructions

* Open the dev tools network tab and filter by XHR
* Browse http://localhost:6543/ and verify that API request are not made for
  * /api/contributor-types
  * /api/facility-processing-types
  * /api/product-types
  * /api/worker-ranges
* Click "More filters" and verify that fetches are now make and that all the dropdowns are populated
* Select items in all the extended field dropdowns then refresh the page, verify that the selections are restored

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
+- Delay loading search field dropdown data [#1731](https://github.com/open-apparel-registry/open-apparel-registry/pull/1731)
+
 ### Deprecated
 
 ### Removed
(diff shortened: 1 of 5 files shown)


jwalgran:
I rebased this on develop and force pushed along with some cleanup.

jwalgran:
We want to make other adjustments to fields in this new sub-component so since I effectively did a review of @TaiWilkin 's initial work I am going to approve and merge this. If there are additional clean ups or changes we would like to do they can be made in a follow up PR.

jwalgran (approved):

PR #1734, opened 2022-03-19
Remove product type dropdown from search and claim

rajadain:
## Overview

Removes product type options from search UI list and the claim form.

Connects #1730 

## Testing Instructions

* Open the home page
* Go to extended fields
  - [x] Ensure that the dropdown has been replaced with a creatable text field
  - [x] Ensure you can still search as before
* Log in as c2@example.com and claim a facility
* In a separate session log in as c1@example.com, browse http://localhost:6543/dashboard/claims, and approve the claim
* In the original session browse http://localhost:6543/claimed, open the claim, and exercised the product types fields. 
  * [x] Ensure you can add/remove choices and that that they save and reload on refresh.
  * [x] Use the link at the upper right of the claim to jump to the facility details and verify that the product types appear as an extended field.

## Checklist

- [x] `fixup!` commits have been squashed
- [ ] CI passes after rebase
- [x] CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines

PR #1740, opened 2022-03-20
Remove Parent Company dropdown options

rajadain:
## Overview

From the search UI. Similar to #1734.

Connects #1736 

## Testing Instructions

* Check out this branch
* Go to the search page
  - [x] Ensure the parent company search does not have a dropdown

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Remove contribute type counts from dropdown [#1732](https://github.com/open-apparel-registry/open-apparel-registry/pull/1732)
 - Remove product type dropdown from search and claim [#1734](https://github.com/open-apparel-registry/open-apparel-registry/pull/1734)
+- Remove parent company dropdown from search [#1736](https://github.com/open-apparel-registry/open-apparel-registry/pull/1736)
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


TaiWilkin (changes requested):
This looks great.
Two notes:
- When clicking out of the input, I kept expecting my text to be searched, despite not having pressed enter (since it looks like a regular search input unless you do press enter). We can get that behavior by adding an `onBlur` to `CreatableInputOnly`:
```
    handleBlur = () => {
        const { inputValue } = this.state;
        const { onChange, value } = this.props;
        if (!inputValue) return;
        this.setState({
            inputValue: '',
        });
        onChange([...value, createOption(inputValue)]);
    };
```
- After refreshing the page, my search term is cleared from the input and the URL params. I think we're clearing out any values that aren't in the contributors dropdown on refresh [here](https://github.com/open-apparel-registry/open-apparel-registry/blob/d175fc75a74988dfcda26c86254ddaf54355c2a8/src/app/src/actions/filters.js#L73), but since we allow any value in here we don't want that. This section also has the purpose of mapping IDs to names, but since we're not using the dropdown any IDs will have been explicitly entered by users and probably shouldn't be overwritten anyway. I think we can remove the parent_company mapping here.

jwalgran:
Thanis, @TaiWilkin. To expedite things I added made two suggested change and they appear to work well. Please check them out.

TaiWilkin (approved):
This is working well, thanks for expediting!

Issue #1738, opened 2022-03-20
DRAFT: Embed detail header not showing contributor name


jwalgran:
# Draft

This is a draft issue, and it will need to clarify specs and formalize what
counts as "done" before work begins on it.

Check with the issue's creator before starting work on this issue.

## Overview

A clear description of the feature request.

### Is your feature request related to a problem? Please describe.

<img width="847" alt="Screen Shot 2022-03-19 at 4 59 43 PM" src="https://user-images.githubusercontent.com/17363/159142492-5af47568-3bff-420e-84b5-149ee66b0f69.png">

### Additional Context

Optional. Add any other context, screenshots, gifs, etc about the feature request here.

jwalgran:
I have not been able to recreate this issue.

Issue #1741, opened 2022-03-20
CSV dowloaded from embed map has contributor attribution


jwalgran:
## Overview

CSV dowloaded from embed map has contributor attribution. We only show extended fields from a single contributor, so there should not be any attribution included.

### Steps to Reproduce

* Log in as c2@example.com
* Submit data with extended fields
* Use the Embed tab of https://staging.openapparel.org/settings to configure and view the example embed
* Download the search results from the embed. 

### Demo
<img width="653" alt="Screen Shot 2022-03-19 at 9 20 54 PM" src="https://user-images.githubusercontent.com/17363/159148076-47cab14e-bbbe-4251-a894-25be51598c01.png">

### Additional context

Optional. Add any other context about the problem here.

PR #1742, opened 2022-03-20
Hide contributor details in embed downloads

TaiWilkin:
## Overview

Facilities downloaded from embed maps had contributor attributions.
We only show extended fields from a single contributor, so there should
not be any attributions included.

Hides the attributions the CSV downloads when in embedded map mode.

Connects #1741 

## Demo

[EmbeddedCSV.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8311528/EmbeddedCSV.csv)

## Testing Instructions

* Log in as [c2@example.com](mailto:c2@example.com)
* Submit data with extended fields
* Use the Embed tab of https://staging.openapparel.org/settings to configure and view the example embed
* Download the search results from the embed
     - [x] No contributor should be listed in the embed fields

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Hide contributor details in embed downloads [#1742](https://github.com/open-apparel-registry/open-apparel-registry/pull/1742)
+
 ### Security
 
 ## [62] - 2022-03-18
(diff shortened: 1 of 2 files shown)


TaiWilkin:
Actually, hold off on reviewing - I went for a front-end fix but I just got an idea relating to a possible back-end issue that might be related to a different bug, so I'm going to take a quick pass at a different approach.

TaiWilkin:
We want to return no other contributor details in embed mode, not anonymized details. After https://github.com/open-apparel-registry/open-apparel-registry/pull/1712, we were incorrectly returning anonymized details instead of None in embed mode. 

This fixed the CSV. It's possible it also will fix https://github.com/open-apparel-registry/open-apparel-registry/issues/1738, but I haven't been able to replicate that locally so can't check. 

@jwalgran ready for review

jwalgran:
Looking at this now.

jwalgran (approved):
I confirmed that the embed CSV no longer has contributor attribution but publicly downloaded CSV does. Nice work debugging and finding the focused back end fix.

Issue #1748, opened 2022-03-21
Number formatting on EM Custom Fields (Decimal Points) 


mariel-oar:
## Overview

HL: I'm no longer seeing decimal points in number of workers, but I uploaded another numeric field as a custom field for an embedded map and it has a decimal point. There were not decimal points in the data I uploaded. This is happening in multiple numeric fields. This is a showstopper because when I try to search by that data point as part of the custom functionality, it won't come up unless I add a decimal point. 

### Expected Behavior

When a custom field is entered as a number format, convert to text to remove number-specific formatting (e.g. ".0") 

### Steps to Reproduce

* enter number data as custom field in EM (without a decimal point) 
* search by number data without adding the decimal point 
* search will not return a match unless decimal point is added 

### Demo

Columbia has paid for this functionality and it needs to work: 
<img width="1080" alt="Screen Shot 2022-03-21 at 9 28 56 AM" src="https://user-images.githubusercontent.com/97469356/159273479-2806b512-b657-4df5-a701-52b864724a22.png">

cc: @hlennett @katieashaw

hlennett:
You can see the Embedded Map in this account: https://staging.openapparel.org/dashboard/lists/?contributor=872

PR #1757, opened 2022-03-21
Fix handling of numeric custom cells in Excel

TaiWilkin:
## Overview

Custom cells can have numeric values. When parsing the sheets, a value
that appears as 130 in the workbook can appear as a float 130.0.

This commit attempts to fix this issue by using an
int(float(x)) nested cast to ensure that the decimal portion is removed.

Note that this will convert all floats to integers when submitted as
custom text. If a contributor would like to submit floats in a custom
text field in the future, we will need to implement that as an
enhancement.

The handling of this was based on the [recent fix for number_of_workers](https://github.com/open-apparel-registry/open-apparel-registry/pull/1663).

Connects #1748 

## Demo

Screen Shot 2022-03-21 at 10 17 46 AM

### NOTE: Custom text indexing can take a minute to run, so wait after adjusting any searchable field. 

On `develop`:
* Run `./scripts/server`
* Submit
[Numeric_Custom_List.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/8316140/Numeric_Custom_List.xlsx)
* Process the list using `./tools/batch_process {list_id}`
* Login as c2@example.com and navigate to the settings -> embed, then set a width and height so your map preview displays
* Select custom_field as a searchable field.
* Search for 23876. No results will return. Search for 23876.0 and one result should return. 
* Deselect custom_field as a searchable field.

On this branch:
* Run `./scripts/server`
* Select custom_2 as a searchable field.
* Search for 12345 and one result should return. 
* Select custom_field as searchable. Search for 23876 and one result should return.

## Checklist

- [ ] CI passes after rebase


src/django/api/helpers.py
```diff
@@ -47,12 +47,19 @@ def get_csv_values(csv_data):
     return values
 
 
+def try_parse_float(value):
```
rajadain, line 50: Consider naming this `try_parse_int_from_float` since that's what you're getting in the end.

rajadain (approved):
+1 tested, this is working well!

rajadain (approved):

Issue #1749, opened 2022-03-21
Values outside of Facility type taxonomy are displaying within that field 


mariel-oar:
## Overview

I wanted to test the behavior when uploading a processing_type_facility_type outside the taxonomy. You can see the result here: [https://staging.openapparel.org/facilities/TR2019098ZWGHGF?q=TR2019098ZWGH](https://staging.openapparel.org/facilities/TR2019098ZWGHGF?q=TR2019098ZWGHGF) 

### Expected Behavior
Values that fall outside of the facility taxonomy should not cause a value to display under the facility type field. These values should be displayed under processing type. 

### Steps to Reproduce

* Upload facility with value that is outside of the facility taxonomy 
* See that this value populates under processing type (this is correct) 
* See that this value also populates under facility type (this is NOT correct) 

### Demo

Similarly, see this profile[: ](https://staging.openapparel.org/facilities/VN2019085WSV6YH?contributors=872)[https://staging.openapparel.org/facilities/VN2019085WSV6YH?contributors=872](https://staging.openapparel.org/facilities/VN2019085WSV6YH?contributors=872) You can see that when an item fits in the taxonomy it generates the correct Facility Type, but if it's outside the taxonomy, it does not.

### Additional context

cc: @katieashaw @hlennett

PR #1759, opened 2022-03-21
Filter out invalid facility_types

TaiWilkin:
## Overview

Multiple values can be submitted for a single facility type field.
When some of these values are valid and others are invalid, we need to
filter out the invalid values to only display the valid ones.

Connects #1749 

## Testing Instructions

* Upload and process [ProcessFacilityType.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8317870/ProcessFacilityType.csv). Confirm matches if needed
* View each facility. Confirm that no raw values display under facility type. 
* For facility TEST VALID confirm that "making shirts" is not displayed under facility type, but Raw Material Processing or Production is. 

## Checklist

- [ ] CI passes after rebase


TaiWilkin:
Thank you for the speed-review!

PR #1763, opened 2022-03-22
Reduce pageSize when downloading to 10

jwalgran:
## Overview

At a pageSize of 50 with `detail=true` we are seeing response times near 30s, sometimes timing out.

In some brief testing reducing the pageSize to 10 results in requests that take 3 to 6 seconds. Setting a pageSize of 20 doubles the response time but interestingly setting a pageSize of 5 doe _not_ reduce the request time. There appears to be a performance floor.

### The effects of reducing page size in production

Screen Shot 2022-03-22 at 1 33 36 PM

### Download testing in development

Screen Shot 2022-03-22 at 1 40 20 PM

## Testing Instructions

* Open the network tab
* Download a CSV and verify that the data requests have a page size of 10

## Checklist

- [x] CI passes after rebase
- [X] CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
+- Reduce pageSize when downloading to 10 [#1763](https://github.com/open-apparel-registry/open-apparel-registry/pull/1763)
+
 ### Deprecated
 
 ### Removed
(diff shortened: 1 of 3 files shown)


jwalgran:
@TaiWilkin reviewed this with me on a screenshare so I will merge this.

Issue #1561, opened 2021-12-16
Turn off extended profile fields for search & facility profiles on Embedded Maps if user hasn't submitted


hlennett:
It would be a confusing experience for an Embedded Map if the Embedded Map owner had not contributed Type of Product, but that still appeared as a searchable field on their map. 

As discussed in our check-in, this would be a natural extension of Extended Profile work for Embedded Map. 

"The additional work that would be required would be conditionally checking if someone has not submitted the extended profile fields."

This is not currently a part of the Extended Profile design but would be a very helpful extension. 

cc @katieashaw

katieashaw:
Does this link to #1522 ?

mariel-oar:
As a related item to this ticket, I want to add that the informational text that displays on (i) hover and when additional filters are displayed should not show up for embedded map users. This is because they are custom selecting the fields they want to display and they have had the ability to add additional custom fields (overlapping with extended field data) for some time now.

mariel-oar:
This may be more related to #1522 which is closed, but we are still seeing this issue where fields that do not have data (e.g. the contributed list does not include product type) is still showing up on the profile. 

See these facilities (both contributed by EM users):
KR2020357B89R59
BD202005316M1V9

hlennett:
I think you'll need the map links to see those examples in action, so I'll add those here: 

Look for KR2020357B89R59 in:
https://www.columbiasportswearcompany.com/corporate-responsibility-group/responsible-practices/supply-chain/

Look for BD202005316M1V9 in:
https://www.levistrauss.com/sustainability-report/community/supplier-map/

hlennett:
Noting that the second part of this issue (Fields still displaying on facility profiles in Embedded Maps when they are empty), still persists: 

Look for KR2020357B89R59 in Columbia's test map: https://open-apparel-registry.github.io/embeded-map-test-fixed-size.html?host=staging.openapparel.org&contributors=2340.

jwalgran:
Thank you for the very specific and helpful test details, @hlennett .

I submitted a quick change for review in #1800 that will remove the explanatory text.  There will be additional pull requests to address the profile display issues.

For engineers:
I traced the Ted Baker example and I believe we are creating showing an empty Parent Company because  Ralph Lauren Corporation has submitted a parent company value for the facility but we are filtering it out on Ted Baker's embedded map.

Embed fields and extended fields have separate data sources but we may be able to share "hide the header if there is no values" code between them.

[...]

PR #1794, opened 2022-04-08
Only show contributed extended fields in embed sidebar search

emilyhu0106:
## Overview

In embed mode, we only want the extended profile fields that are contributed by the contributor to appear on the filter sidebar. Perviously, it would be a confusing experience for an Embedded Map if the Embedded Map owner had not contributed Type of Product, but that still appeared as a searchable field on their map.

This PR add extended_fields to embed_configs endpoint and use the returned value to check if the extended profile fields are contributed by the user.

Connects #1561

## Notes

When users contributing facilities with extended fields, we will match the processing type with the associated facility type based on taxonomy (see #1616), which means we may populate the facility type / processing type automatically without explicitly contributing that field.

## Testing Instructions

This PR assumes `resetdb` has been run.
* Check out this branch, run `./script/update`
* Log in as c3@example.com, and make sure it has any level of embed access
    - [x] Open [embed](http://localhost:6543/facilities?contributors=2&embed=1) mode and there should be no "More Filter" since no extended fields are contributed
    - [x] Open OAR [front-page ](http://localhost:6543/) and make sure the sidebar looks as normal
* Contribute [ExtendedFieldsTestListLimited.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8465965/ExtendedFieldsTestListLimited.csv) and fully process it using `./tools/batch-process {list-id}`
    - [x] Open [embed](http://localhost:6543/facilities?contributors=2&embed=1) mode, extend "More Filters", and make sure only "Facility Type", "Number of Workers", "Processing Type", and "Product Type" filters are shown.
    - [x] Open OAR [front-page ](http://localhost:6543/) and make sure the sidebar looks as normal

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/components/FilterSidebarExtendedSearch.jsx
+++ b/src/app/src/components/FilterSidebarExtendedSearch.jsx
@@ -108,6 +108,9 @@ const mapProcessingTypeOptions = (fPTypes, fTypes) => {
     return mapDjangoChoiceTuplesValueToSelectOptions(uniq(pTypes.sort()));
 };
 
+const isExtendedFieldForThisContributor = (field, extendedFields) =>
+    extendedFields.includes(field.toLowerCase());
+
 function FilterSidebarExtendedSearch({
     contributorTypeOptions,
     facilityProcessingTypeOptions,
@@ -127,6 +130,7 @@ function FilterSidebarExtendedSearch({
     fetchingFacilities,
     fetchingExtendedOptions,
     embed,
+    embedExtendedFields,
     classes,
     fetchContributorTypes,
     fetchFacilityProcessingType,
@@ -396,7 +400,7 @@ function mapStateToProps({
     facilities: {
         facilities: { data: facilities, fetching: fetchingFacilities },
     },
-    embeddedMap: { embed },
+    embeddedMap: { embed, config },
 }) {
     return {
         contributorTypeOptions,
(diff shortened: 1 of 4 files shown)


src/app/src/components/FilterSidebarExtendedSearch.jsx
```diff
@@ -194,110 +198,160 @@ function FilterSidebarExtendedSearch({
                 </div>
             </div>
             <div className="form__field">
```
jwalgran, line 200: When testing this with an API submission that included only 1 of the 5 extended fields I saw a large amount of white space in the embed sidebar

Screen Shot 2022-04-11 at 3 19 16 PM


Consider moving these `<div className="form__field">` tags inside the `ShowOnly` tags so that we don't render extra empty divs.

jwalgran (commented):
I was able to this correctly hiding fields from the search form. There is a stray model field and migration that I think we can remove as well as some extra whitesapce that we can avoid by slightly adjusting how we hide the search fields.

src/app/src/components/FilterSidebarExtendedSearch.jsx
emilyhu0106: Nice catch! I did see some white spaces but didn't connect it with the empty `div`

emilyhu0106:
This PR is ready for another look. Please run `update` to make sure `0088` is the latest migration

jwalgran (approved):
I demoed this in our sprint review and it worked well. Thank you.

PR #1800, opened 2022-04-14
Hide extended field messages in embed mode

jwalgran:
## Overview

The explanatory text mentions fields that may not be displayed and the caveat that data may not be available because the feature is new is also not relevant.

Connects #1561

## Demo

Screen Shot 2022-04-14 at 7 12 43 AM

Screen Shot 2022-04-14 at 7 12 34 AM

## Testing Instructions

* Log in as c2@example.com
* Set up the embed config
* Contribute [ExtendedFieldsTestListLimited.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8465965/ExtendedFieldsTestListLimited.csv) and fully process it using `./tools/batch-process {list-id}` so there there 
* Confirm that the embed messages are not  displayed in embed mode but still show in the full map

## Checklist

- [x] CI passes after rebase


emilyhu0106 (approved):
It all works as intended! Thank you for addressing this missing piece in 1561!

jwalgran:
Thanks for the review.

PR #1804, opened 2022-04-16
Use exact match in parent company when creating extended field

emilyhu0106:
## Overview

Use exact match instead of fuzzy match to when creating the parent company field. Create migration to reprocess and index the parent company field.

Tests are updated accordingly, including revising the `ParentCompanyTestCase` inputs to be exactly the same as the contributor, and change the sequence of results in `test_filter_by_name_verified`

Will visit trigram matching at #1805. 

Connects #85

## Demo

image

## Notes

In test_filter_by_name_verified, we suspect that not using `similarity`  changes the sorting of same names, making the second name comes at first when both names are the same. So we mark `c1` as verified to test if sorting by verified is working as expected.

## Testing Instructions

* Checkout this branch and run `./scripts/update`
* Go to the django admin [site](http://localhost:8081/admin/api/contributor/) and log in as c1@example.com
* Make a new contributor - Kith Fashion Limited 
* Go to local [OAR](http://localhost:6543/) page and log in as c2@example.com
* Contribute [RalphLaurenTesting.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8537268/RalphLaurenTesting.csv) and fully process it with `./tools/batch_process {list-id}`
- [x] Go to Company A and Company B detail page, the parent company field should match the original list

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Update DB anonymization script [#1769](https://github.com/open-apparel-registry/open-apparel-registry/pull/1769)
 - Update tileer load test script [#1770](https://github.com/open-apparel-registry/open-apparel-registry/pull/1770)
 - Only show contributed extended fields in embed sidebar search [#1794](https://github.com/open-apparel-registry/open-apparel-registry/pull/1794) [#1800](https://github.com/open-apparel-registry/open-apparel-registry/pull/1800)
+- Use exact match in parent company when creating extended field [#1804](https://github.com/open-apparel-registry/open-apparel-registry/pull/1804)
 
 ### Deprecated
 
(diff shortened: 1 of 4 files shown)


jwalgran:
I looked at the first failing test

```
FAIL: test_search_by_id (api.tests.ParentCompanyTestCase)
----------------------------------------------------------------------
  File "/usr/local/lib/python3.7/unittest/mock.py", line 1256, in patched
    return func(*args, **keywargs)
  File "/usr/local/src/api/tests.py", line 7896, in test_search_by_id
    self.assertEquals(data['count'], 1)
AssertionError: 0 != 1

```

I believe this is failing because we removed fuzzy matching but the test is still using a misspelled contributor name

https://github.com/open-apparel-registry/open-apparel-registry/blob/8a9654a6034f3e03f10a159b7e966cb7c4f429c6/src/django/api/tests.py#L7983

```
            'parent_company': 'TEST CNTRIBUTOR'
```

If I change this line to 

```
            'parent_company': 'TEST CONTRIBUTOR 1'
```

The test passes. 

I suspect similar changes will get the other tests passing.

TaiWilkin (approved):
The suggested CSV processed as expected. I also explored the data in my local shell a bit after migrating, and the facilities appear to have indexed correctly. 
Nice work!

emilyhu0106:
Thank you for reviewing!

PR #1811, opened 2022-04-26
Read percentage as string in Excel

emilyhu0106:
## Overview

Make sure values in excel sheets that appears to be percentage actually read as strings. E.g. in excel sheets, the value of the cell `68%` can be `0.68`, and this PR makes sure that `'68%'` is being read and displayed in the embedded maps.

Connects [#85](https://github.com/open-apparel-registry/open-apparel-registry-clients/issues/85)

## Demo

image

## Testing Instructions

**Baseline**
* Log in as c2@example.com (make sure it has embed access), contribute [XLSX - Testing.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/8561783/XLSX.-.Testing.xlsx) and fully process it.
* Go to embed map setting, make sure `PERCENT_FEMALE_WORKERS` and `DECIMAL_FIELD` are both visible
* Open any facility under the new list
    - [x] Both `PERCENT_FEMALE_WORKERS` and `DECIMAL_FIELD` should appear as `0`

**Test**
* Check out this branch and run `update` and `server`
* Log in as c2@example.com, contribute [XLSX - Testing.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/8561783/XLSX.-.Testing.xlsx) and fully process it.
* In [embed map](http://localhost:6543/facilities?contributors=1&embed=1), open any facility under the new list
    - [x] `PERCENT_FEMALE_WORKERS` should have a percentage value and `DECIMAL_FIELD` should appear as `0`

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/processing.py
+++ b/src/django/api/processing.py
@@ -3,7 +3,8 @@ import csv
 import traceback
 import sys
 
-import xlrd
+from openpyxl import load_workbook
+from openpyxl.utils import get_column_letter
 
 from datetime import datetime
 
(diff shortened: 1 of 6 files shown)


src/django/api/processing.py
```diff
@@ -33,7 +36,53 @@ def _report_error_to_rollbar(file, request):
...
+            for col in percent_col:
+                for cell in ws[col]:
+                    if cell.row != 1:
+                        cell.value = f"{cell.value * 100}%"
+
+        header = ','.join([cell.value for cell in ws[1]])
+
+        rows = ['"{}"'.format(
+            '","'.join([str(cell.value) for cell in ws[idx]]))
+                for idx in range(2, ws.max_row)]
```
jwalgran, line 77: When I processed the `XLSX.-.Testing.xlsx` worksheet I only ended up with 18 of the 18 roes. I suspect this is becasue Python's `range` function does not include the last value

```python
>>> list(range(1,10))
[1, 2, 3, 4, 5, 6, 7, 8, 9]
```
I expect this is a simple fix
```python
for idx in range(2, ws.max_row + 1)]
```

src/django/api/processing.py
```diff
@@ -33,7 +36,53 @@ def _report_error_to_rollbar(file, request):
...
+
+        return header, rows
+    except Exception:
+        _report_error_to_rollbar(file, request)
+        raise ValidationError('Error parsing Excel (.xlsx) file')
+
+
+def get_xls_sheet(file, request):
```
jwalgran, line 85: The OAR team confirmed that we can drop support for `.xls` files.  I like the new `parse_xlsx` name so we don't have to revert back to the old `parse_excel`, just remove the methods that deal with xls. 

In addition to removing the handling here we will also want to remove xls from the extensions list in `ContributeForm`

https://github.com/open-apparel-registry/open-apparel-registry/blob/77717f0046a5b0c6b4d83966df2bfa3d4e2a3bf4/src/app/src/components/ContributeForm.jsx#L187

src/django/api/processing.py
```diff
@@ -33,7 +36,53 @@ def _report_error_to_rollbar(file, request):
...
+
+        for cell in ws[2]:
+            if '%' in cell.number_format:
+                percent_col.append(get_column_letter(cell.column))
+
+        if percent_col:
+            for col in percent_col:
+                for cell in ws[col]:
+                    if cell.row != 1:
+                        cell.value = f"{cell.value * 100}%"
```
jwalgran, line 71: Here is a potential implementation for a formatting helper function 

```python
def format_percent(value):
    if value <= 1.0:
        str_value = str(value * 100)
    else:
        str_value = str(value)
    if str_value[-2:] == '.0':
         str_value = str_value[0:len(str_value)-2]
    return str_value + '%'
```

src/django/requirements.txt
```diff
@@ -27,4 +27,5 @@ thefuzz==0.19.0
 Unidecode==1.0.23
 xlrd==1.2.0
```
jwalgran, line 28: The OAR team confirmed that we can drop support for .xls files so we can remove `xlrd` from our requirements.

jwalgran (commented):
I was able to confirm the correct handling of cells with percent format strings using the new library. I found a small issue and made some suggestions.

emilyhu0106:
Thank you for your detailed review, Justin! 
I have one question related to the formatting. You commented this in the issue:
> If the format string has a % symbol in it, convert the cell value to a string: 0.12 becomes 12% and 0.234 becomes 23.4% (drop trailing zeroes)

Right now all percentages will be formatted as `00.0%` (e.g. `12.0%` and `16.2%`) and I didn't find a [built](https://openpyxl.readthedocs.io/en/stable/_modules/openpyxl/styles/numbers.html)-in way to drop the trailing zeros easily. Do you have any suggestions?

src/django/api/processing.py
emilyhu0106: This is a nice implementation!

jwalgran:
The CI build failure appears to just be lint

```
./api/processing.py:85:34: W291 trailing whitespace
./api/processing.py:99:1: E302 expected 2 blank lines, found 1
```

jwalgran (approved):
Changes look good. I retested this scratch, starting with reproducing the issue on the `develop` branch and the list as a replacement on this branch, verifying that the percent values were now correct. Nice work.

emilyhu0106:
Thank you for reviewing!

Issue #1767, opened 2022-03-22
Reporting for Extended Profiles 


mariel-oar:
As part of tracking the EP release, we would like to have additional reports added that can be run on a date range that we specify. 

Reports: 
* List of all processing_type_facility_type values that do not match a value in facility_type taxonomy 
  * No dates involved.
* Count of facility profiles by additional data points (e.g. count of profiles with number of workers, count with parent company, etc). 
  * For each month (based on `updated_at`), column for each extended field with a facility count
* Count of contributors by additional data points (e.g. count of contributors that submitted number of workers, count submitted parent company, etc).  
  * Similar to the previous report, just counting contrib rather than facility  
* Report that lists all taxonomy values for product_type, processing_type_facility_type, and the raw values that were fuzzy matched to these values 
  * No dates involved

PR #1826, opened 2022-04-29
Add extended field reports

TaiWilkin:
## Overview

Adds new reports for extended fields:

- Contributors With Extended Fields Cumulative: This report lists
the number of contributors who have submitted each field type. It is
organized into months, where each month's count includes all extended
fields with an update date of that month or earlier.
- Facilities With Extended Fields Cumulative: This report lists
the number of facilities that have each field type. It is
organized into months, where each month's count includes all extended
fields with an update date of that month or earlier.
- Processing Type Facility Type Matched Values: This report lists each
taxonomy value that has been assigned to an extended field, accompanied
by a list of the raw values that have been matched to that value (separated
by '|').
- Processing Type Facility Type Unmatched Values: This report lists
each raw value that was submitted as a Processing_Type_Facility_Type
and wasn't matched to a taxonomy value, as well as a count of the number
of times that raw value was submitted.
- Submitted Product Type Values: This report lists each raw value that
was submitted for a product_type extended field, as well as a count of
the number of times that raw value was submitted.

Connects #1767 

## Demo

Screen Shot 2022-04-29 at 2 43 15 PM
Screen Shot 2022-04-29 at 2 43 24 PM
Screen Shot 2022-04-29 at 2 43 35 PM
Screen Shot 2022-04-29 at 2 43 45 PM
Screen Shot 2022-04-29 at 2 43 58 PM

## Notes

The issue requested one report with "Report that lists all taxonomy values for product_type, processing_type_facility_type, and the raw values that were fuzzy matched to these values",
but `product_type` doesn't use any fuzzy matching. It could be that this was meant to be `processing_type`
(which is already included under the `processing_type_facility_type` report), but because the
OAR team has also mentioned wanting to evaluate the `product_type` values being submitted,
I included the `product_type` values as a separate report.

## Testing Instructions

* Run `./scripts/server`
* Visit [contributors_with_extended_fields_cumulative](http://localhost:8081/admin/reports/contributors_with_extended_fields_cumulative)
* Visit [facilities_with_extended_fields_cumulative](http://localhost:8081/admin/reports/facilities_with_extended_fields_cumulative)
* Visit [processing_type_facility_type_matched_values](http://localhost:8081/admin/reports/processing_type_facility_type_matched_values)
* Visit [processing_type_facility_type_unmatched_values](http://localhost:8081/admin/reports/processing_type_facility_type_unmatched_values)
* Visit [submitted_product_type_values](http://localhost:8081/admin/reports/submitted_product_type_values)
* Confirm each report contains the expected values and that the format makes sense

## Checklist

- [ ] CI passes after rebase


src/django/api/reports.py
```diff
@@ -120,6 +121,102 @@ def weekly_geocoding_time_with_queue():
...
+    data = defaultdict(list)
+    months = ExtendedField.objects.dates('updated_at', 'month')
+    for month in months:
+        m = month.strftime(DateFormats.MONTH)
+        data[m] = [m]
+        fields = ExtendedField.objects.filter(
+            updated_at__year__lte=month.year,
+            updated_at__month__lte=month.month)
+
+        for field_name, _ in ExtendedField.FIELD_CHOICES:
```
jwalgran, line 184: Not a problem introduced in this PR, but adding this use of `FIELD_CHOICES` exposed the fact that facility type and processing type are repeated . We will want to remove them, which will likely require creating a migration since is "changes" the choices attached to a model field.

166006960-ad98c044-0077-4889-b8cf-7c31f04cc1a9

https://github.com/open-apparel-registry/open-apparel-registry/blame/a070e8be9dce56112de4ebb757f477c374b0e595/src/django/api/models.py#L2387-L2398

jwalgran, line 184: Not a problem introduced in this PR, but adding this use of `FIELD_CHOICES` exposed the fact that facility type and processing type are repeated . We will want to remove them, which will likely require creating a migration since is "changes" the choices attached to a model field.

166006960-ad98c044-0077-4889-b8cf-7c31f04cc1a9

https://github.com/open-apparel-registry/open-apparel-registry/blame/a070e8be9dce56112de4ebb757f477c374b0e595/src/django/api/models.py#L2387-L2398

jwalgran (commented):
The logic in these Python reports was easy for me to follow. I ran into a few issues in testing.

src/django/api/reports.py
TaiWilkin: I made a model change and a migration in [2ae5f45](https://github.com/open-apparel-registry/open-apparel-registry/pull/1826/commits/2ae5f4525a8119eaf3b172fd9f4c12168fcc1e7a). I checked my local data after migrating to confirm that removing two of the values didn't have any impact on my processing_type and facility_type fields, just in case, and confirmed there were no empty values for the field_name.

src/django/api/reports.py
```diff
@@ -120,6 +121,102 @@ def weekly_geocoding_time_with_queue():
...
+def processing_type_facility_type_matched():
+    temp_data = defaultdict(set)
+
+    for (raw_values, matched_values) in ExtendedField.objects.filter(
+        Q(field_name='processing_type') | Q(field_name='facility_type')
+    ).distinct(
+        'value__raw_values', 'value__matched_values'
+    ).values_list('value__raw_values', 'value__matched_values').iterator():
+        for i, raw_value in enumerate(raw_values.split('|')):
+            if matched_values[i][3] is not None:
```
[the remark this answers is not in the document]
TaiWilkin: After adding some single string and array values to the facility_type and processing_type values in the database, I was able to replicate these errors locally. I updated the report to process the raw_values in the same way as the extended fields script [here](https://github.com/open-apparel-registry/open-apparel-registry/blob/a070e8be9dce56112de4ebb757f477c374b0e595/src/django/api/extended_fields.py#L24). Noting the removal of duplicate values there, I tested submitting a facility_type with repeated values and confirmed that this caused an indexing error. Using the same process of deduplication as the extended_fields processing in report resulted in the duplicate values being handled correctly without error.

TaiWilkin:
Thanks for the thorough review and for testing on staging. This is ready for another look.

jwalgran (commented):
I ran into a crash when testing on staging. My research into this is included below but the short version is that there are 4  processing type extended fields created from claims have a mismatch between the count of raw and matched values. I propose working around this by reporting the error to Rollbar rather than crash and open a follow up issue to resolve it.

https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/1549/

Using Rollbar to look at the local values at the point of the exception there are 7 `deduped_values` and 6 `matched_values`. A 

[
  "Denim Services",
  "Fabric Dyeing",
  "Pleating",
  "Printing",
  "Sewing",
  "Spinning",
  "Trimming/Embellishment"
]
[
  [
    "PROCESSING_TYPE",
    "FUZZY",
    "Printing, Product Dyeing and Laundering",
    "Dyeing"
  ],
  [
    "PROCESSING_TYPE",
    "EXACT",
    "Final Product Assembly",
    "Pleating"
  ],
  [
    "PROCESSING_TYPE",
    "EXACT",
    "Printing, Product Dyeing and Laundering",
    "Printing"
  ],
  [
    "PROCESSING_TYPE",
    "EXACT",
    "Final Product Assembly",
    "Sewing"
  ],
  [
    "PROCESSING_TYPE",
    "EXACT",
    "Raw Material Processing or Production",
    "Spinning"
  ],
  [
    "PROCESSING_TYPE",
    "FUZZY",
    "Textile or Material Production",
    "Embellishment"
  ]
]


This should not happen. The unmatched "Denim Services" should result in a [None, None, None, None] match row. I tested this by POSTing a facility to the API with 

"facility_type_processing_type": [
  "Denim Services",
  "Fabric Dyeing",
  "Pleating",
  "Printing",
  "Sewing",
  "Spinning",
  "Trimming/Embellishment"
]
And it did create the None item in matched_values

I ran this query on staging to try and get more information

```
select * from (
  select facility_id, facility_claim_id, facility_list_item_id,
         CASE WHEN jsonb_typeof(value->'raw_values') = 'array'
           THEN jsonb_array_length(value->'raw_values')
         ELSE -1
         END as raw_count,
         jsonb_array_length(value->'matched_values') as matched_count
    from api_extendedfield
   where field_name = 'processing_type'
  ) s
where raw_count > 0 and raw_count != matched_count;

```

There were 4 extended fields created from claims that had a count mismatch between raw and matched

```
openapparelregistry-> where raw_count > 0 and raw_count != matched_count;
   facility_id   | facility_claim_id | facility_list_item_id | raw_count | matched_count
-----------------+-------------------+-----------------------+-----------+---------------
 IT2019311JZR7ZX |                56 |                       |         7 |             6
 CN201928823FEJ3 |                53 |                       |        12 |            11
 PK2021341XSCC8Y |               333 |                       |        19 |            17
 IN202108462CDJV |               350 |                       |        24 |            22
```

When I ran the query against production I got the same 4 rows.

In staging there are 66 other processing type extended field rows from claims with matching counts.

Here is a partial WIP commit that I added to my staging test branch that adds a try/catch with reporting to Rollbar.  This fixes the report on staging and logs 4 errors to Rollbar, which reinforces the finding that the report code handles the other cases properly.

https://github.com/open-apparel-registry/open-apparel-registry/commit/1352f2bf247d93baf808cc890796346338d2f31f

Screen Shot 2022-05-04 at 12 06 41 AM

TaiWilkin:
@jwalgran I'm pretty sure I know exactly where this error is coming from: this PR https://github.com/open-apparel-registry/open-apparel-registry/pull/1677, when we dropped Denim Services and several other values. We decided at the time to treat them as invalid values vs set them to `None`, but that's causing issues with the logic here.

src/django/api/reports.py
```diff
@@ -120,6 +121,116 @@ def weekly_geocoding_time_with_queue():
...
+            created_at__month__lte=month.month)
+
+        for field_name, _ in ExtendedField.FIELD_CHOICES:
+            field_count = fields.filter(field_name=field_name).distinct(
+                'contributor_id').count()
+            data[m].append(field_count)
+
+    columns = ['month']
+    for field_name, _ in ExtendedField.FIELD_CHOICES:
+        columns.append(field_name)
```
jwalgran, line 205: The OAR team request that we add a "(claim)" suffix to the `country`, `name` and `address` column headers to clearly indicate that they come from the claim process, bot facility data submission.

jwalgran (commented):
Thank you for looking up and referencing that previous PR. Yes. I agree with your plan to filter out these values but also use the try/catch.

src/django/api/reports.py
TaiWilkin: Followup question: We had added `country` as one of the initial fields, but I don't believe that the claim form actually includes `country` as a value for the facility (only for the Office Information section). Do we want to remove this as well as the two duplicate values, since it's not actually a possible field type at the moment?

jwalgran: Good idea. Having the country column filled with all zeros all the time just adds noise to the report.

TaiWilkin:
Updated processing_type reports (I only have this month's data in my system right now):
Screen Shot 2022-05-05 at 8 04 53 AM
Screen Shot 2022-05-05 at 8 04 43 AM

jwalgran (approved):
Looks good, and error free on staging! Thanks for working through the revisions with me.

Issue #1389, opened 2021-06-09
Error opening these two OAR reports (COPY UPDATE)


jwalgran:
@isaacokunato commented on [Wed Jun 09 2021](https://github.com/open-apparel-registry/open-apparel-registry-clients/issues/34)

### Overview
Some unexpected error have been produced when trying to view Reports

### Expected Behavior
Reports are to be on display when opened

### Actual Behavior
It continues to load for a long time before displaying **504 Error, Request could not be satisfied** 

### Suggested remediation

- Rewrite SQL reports as Python

### Error Links
The two links are 
- Percent Facilities With Multiple Matches openapparel.org/admin/reports/percent-facilities-with-multiple-matches/ and
- Average Affiliations openapparel.org/admin/reports/average-affiliations/

Attached is the image of the display error 

![Report error message](https://user-images.githubusercontent.com/82677956/121401510-b23f6d80-c950-11eb-8f44-5fa7526c1107.JPG)

@hlennett @katieashaw 

## Note

This creates outages that are not caused by ECS tasks being cycled. Determine what resource is being consumed and determine how to rewrite the report to avoid it.

Use `makecsvs` management command to build large files.

jwalgran:
Thanks for opening this. I moved it to our public repository because it is not related to a specific contributor.

mariel-oar:
Can the name on the reporting page be updated to say "Do not open" for these list entries? 
(This is assuming that they actual fix for these reports is too involved for right now)

jwalgran:
We actually pull the name of the report from the name of the SQL file in which the query is saved so, for instance, renaming `average_affiliations.sql` to `do_not_open_average_affiliations.sql` is a short-term option.

mariel-oar:
ok, that works. The point is just to make safeguard ourselves from accidentally clicking these (especially as new team members are onboarding)

obrienad:
@mariel-oar - just checkin if you only want us to do the copy update (1 point) OR look at report problem too (3 points).

mariel-oar:
Just the copy update @obrienad

obrienad:
Thanks!

PR #1836, opened 2022-05-05
Mark failing reports with Do Not Open

TaiWilkin:
## Overview

Two reports (Percent Facilities With Multiple Matches, and
Average Affiliations) timeout and cause site outages.

We are updating their names to reflect that they should not be opened
until such time as we have time for a full fix of the reports.

Connects #1389 

## Demo

Screen Shot 2022-05-05 at 9 15 08 AM

## Testing Instructions

* Run `./scripts/server` and login as c1@example.com
* View [reports](http://localhost:8081/admin/reports/) and confirm that the two reports have 'Do Not Open' at the beginning of their names.

## Checklist

- [x] CI passes after rebase


emilyhu0106 (approved):
Looks all good!

TaiWilkin:
Thank you for reviewing!

Issue #1835, opened 2022-05-05
Users unable to upload Excel files


hlennett:
We've had three users now in recent days who have reached out because they were unable to upload files to the OAR. They are receiving the: "An error prevented uploading that file" message on the Contribute screen. @vrwOAR and I have confirmed that behavior for those three lists on staging. When we change the file to a CSV, we are able to upload it. Attaching the three files here. 

[OAR_Contributor_Template_Additional_Data_Points_Excel (5).xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/8629766/OAR_Contributor_Template_Additional_Data_Points_Excel.5.xlsx)

[2.OAR_Contributor_NAKD_April_2022.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/8629767/2.OAR_Contributor_NAKD_April_2022.xlsx)

[OAR_ELK_2022_SUPPLIER_LIST_FINAL.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/8629757/OAR_ELK_2022_SUPPLIER_LIST_FINAL.xlsx)

@vrwOAR also tested with a couple of Excel files that she was able to previously upload to prod and now do not work when trying to upload on staging:

[OSKLEN_Tier3_Raw_Material_Suppliers_jun2020.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/8629800/OSKLEN_Tier3_Raw_Material_Suppliers_jun2020.xlsx)

[Pernambucanas_Domestic_Suppliers_feb2022.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/8629801/Pernambucanas_Domestic_Suppliers_feb2022.xlsx)

---

An exception logged in Rollbar when submitting [OSKLEN_Tier3_Raw_Material_Suppliers_jun2020.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/8629800/OSKLEN_Tier3_Raw_Material_Suppliers_jun2020.xlsx)

https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/1541/

<img width="1093" alt="Screen Shot 2022-05-05 at 6 31 09 AM" src="https://user-images.githubusercontent.com/17363/166933920-d343f0e5-c11a-4af8-8bf5-5788c2be352e.png">

mariel-oar:
@jwalgran, can you investigate how often this is happening? if it's widespread we think it is a priority to have address.

[...]

PR #1837, opened 2022-05-05
Handle blank rows and columns in .xlsx

TaiWilkin:
## Overview

Blank cells have the value 'None', which cause an error when attempting
to .join the cell values for the header. We now format None values to strings
before joining.

After fixing that error, blank rows were then being imported as facility list items and
throwing parse errors. We now skip rows where all the cells are blank.

Connects #1835 

## Demo

Screen Shot 2022-05-05 at 11 42 54 AM

## Testing Instructions

* On `develop`, contribute any list from #1835, and confirm that it throws an error.
* On this branch, contribute any list from #1835, and confirm that it uploads without error.
* On this branch, contribute [OAR_Contributor_Template_Additional_Data_Points_Excel.5.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/8632705/OAR_Contributor_Template_Additional_Data_Points_Excel.5.xlsx) and confirm that it is considered to have only one row. 
* Run `./tools/batch_process {id}` for OAR_Contributor_Template_Additional_Data_Points_Excel.5.xlsx and confirm that it processes correctly and only produces one facility list item. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/processing.py
+++ b/src/django/api/processing.py
@@ -84,11 +84,12 @@ def parse_xlsx(file, request):
                     if cell.row != 1:
                         cell.value = format_percent(cell.value)
 
-        header = ','.join([cell.value for cell in ws[1]])
+        header = ','.join([format_cell_value(cell.value) for cell in ws[1]])
 
         rows = ['"{}"'.format(
             '","'.join([format_cell_value(cell.value) for cell in ws[idx]]))
-                for idx in range(2, ws.max_row + 1)]
+                for idx in range(2, ws.max_row + 1)
+                if any(cell.value is not None for cell in ws[idx])]
 
         return header, rows
     except Exception:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Handle blank rows and columns in .xlsx [#1837](https://github.com/open-apparel-registry/open-apparel-registry/pull/1837)
+
 ### Security
 
 ## [65] - 2022-04-28


jwalgran:
Looking at this now.

src/django/api/processing.py
```diff
@@ -84,11 +84,12 @@ def parse_xlsx(file, request):
...
                         cell.value = format_percent(cell.value)
 
-        header = ','.join([cell.value for cell in ws[1]])
+        header = ','.join([format_cell_value(cell.value) for cell in ws[1]])
 
         rows = ['"{}"'.format(
             '","'.join([format_cell_value(cell.value) for cell in ws[idx]]))
-                for idx in range(2, ws.max_row + 1)]
+                for idx in range(2, ws.max_row + 1)
+                if any(cell.value is not None for cell in ws[idx])]
```
jwalgran, line 92: I was initially torn on this change because dropping rows will lead to there being a misalignment between the row index and the row number in the source worksheet. However, since blank rows will always just fail and a lot of them will slow down the current geocoding process since we  start up a batch job for every row regardless of status, I came around to the idea of discarding them.

jwalgran (approved):
I confirmed that I could upload all all of the XLSX files linked in #1835 and I uploaded another 2720 row workbook to verify that this change was not adding any notable performance overhead.

src/django/api/processing.py
TaiWilkin: @jwalgran As an alternative approach, we could calculate the last row with any values, then stop processing after that row. That would ensure that the actual facility rows had the correct row index, but prevent us from booting up large numbers of rows from the end of the list. We would still start the batch job for some empty rows, if they were in the middle of the list, but I think that would be relatively infrequent.

TaiWilkin:
Thank you for reviewing!

Issue #1643, opened 2022-02-11
Improve searchability of contributor drop down (admin site) 


mariel-oar:
## Overview

In the admin site, there are several pages that have a contributor drop down field (e.g. Source page - [https://openapparel.org/admin/api/source/](https://www.google.com/url?q=https://openapparel.org/admin/api/source/&sa=D&source=docs&ust=1648644038317996&usg=AOvVaw3gBOHEJoxBjHZPeNoJd9u8)). When the items in the dropdown are viewed, there seems to be no order to how the items are displayed. This creates issues in using the tool given the growing number of contributors. 

Proposal: update the display of the items in the dropdown to be either alphabetical or chronological, based on their associated number that identifies the sequence in which they were added to the OAR. 

If it is straightforward, it would be ideal that the dropdown be searchable with partial-matches enabled. 

Demo: 
The drop down in question appears on this page when you click on a specific source: [https://openapparel.org/admin/api/source/341398/change/](https://www.google.com/url?q=https://openapparel.org/admin/api/source/341398/change/&sa=D&source=docs&ust=1648644038318100&usg=AOvVaw1zb63TM9tJ4va_unGKTG05)

mariel-oar:
![Screen Shot 2022-02-11 at 12 53 52 PM](https://user-images.githubusercontent.com/97469356/153662850-9663f581-d754-40e1-ad97-9e5f3b789dde.png)

jwalgran:
I estimated this as a :one: since the Django admin provides a method for sorting the options list

https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.formfield_for_foreignkey

Issue #1206, opened 2021-01-08
Improve the interface for creating API limits


jwalgran:
## Overview

`ApiLimit` objects are currently managed via the Django admin with not custom behavior. Update this management interface, either by replacing it with a React page under /dashboard or modifying the Django admin page so that more context is available when creating `ApiLimit` records.

### Is your feature request related to a problem? Please describe.

In the initial set of 6 limits I created when launching the API limiting feature, 2 were incorrect. In both cases there were multiple contributors with similar names and I created the limit with the wrong contributor.

Some deficiencies in the current Django admin form:

- The contributor dropdown is not sorted alphabetically
- The related admin user for the contributor is not shown
- Aggregated historical API request activity is not shown
 
### Describe the solution you'd like

- Create an updated management UI for API limits that makes it less likely that the wrong contributor account is selected.

PR #1848, opened 2022-05-16
Use autocomplete ordered by name for admin Contributors dropdown

maurizi:
## Overview

Use autocomplete ordered by name for admin Contributors dropdown.

As a side-effect of the method used for sorting the autocomplete list, the Contributor admin list is now also sorted by name - I assume this is fine.

Connects #1643

## Demo

localhost_8081_admin_api_source_15_change_

## Notes

I _think_ despite `contributor` being a FK to more than just 3 models, these are the only 3 models that have Django admin support.

Potentially also addresses some of https://github.com/open-apparel-registry/open-apparel-registry/issues/1206?

## Testing Instructions

* `./scripts/server`
* Go to a Facility Claim, Source, or Api Limit Django admin Change or Add page. instead of an unsorted `<select>` field for contributors, there should be an autocomplete w/ results sorted by name

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 ### Changed
+- Use autocomplete ordered by name for admin contributors dropdown [#1848](https://github.com/open-apparel-registry/open-apparel-registry/pull/1848)
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


jwalgran:
Looking at this now.

jwalgran (commented):
Confirmed. 👍

jwalgran (approved):

Issue #1788, opened 2022-04-04
Priority Errors - Steps to resolve (not necessarily root cause) 


mariel-oar:
##Context 
There are several new error messages that started coming up in the weeks leading up to EP launch. They are not necessarily EP related, but I mention that to give you a sense of the timeframe; we had not seen these errors previously. 

Users often reach out to the OAR team when they run into errors and ask us to help them understand why the error occurred and provide tips on how to resolve. Previously, we have been able to offer guidance on how to correct the issue; however, with these new error messages, we do not know why they occurred and are therefore limited in our ability to help our users. 

We have identified a list of error messages (below) that users have reached out about in the last several weeks. We would like the Azavea team to troubleshoot these error messages and provide any steps taken to resolve the issue. We do not need to identify the root cause as part of this exercise; please focus on remedying the error and list the steps taken to resolve in the sections below. 

##Additional info
- We would like to time box this effort to 4 hours to start and see how many of the errors can be addressed in that time 
- The error list and template for steps taken is here (Please update the sections of the doc) - https://docs.google.com/document/d/1IwHmiBbVOUf9EQ9EHkp312sALjRksHIynJgvzLSJasA/edit?usp=sharing

hlennett:
#1784 Maybe/likely related.

obrienad:
OAR requested we timebox this to 4 hours, thus 1 point.

obrienad:
Moved to backlog since not all errors were addressed within timebox.

PR #1853, opened 2022-05-20
Fix geocoding error message to include status code

maurizi:
## Overview

I was unable to determine the cause of some of geocoding errors being logged here - including the status code should help when this error occurs in the future (and it seems like it was always the intent to include the status code here).

Connects #1788 

## Testing Instructions

* I'm not actually sure how to trigger this error?

## Checklist

- [x] CI passes after rebase


TaiWilkin (approved):
I updated the code manually to add `r.status_code = 404` and then contributed and processed a list. That worked to return the expected error status:
Screen Shot 2022-05-20 at 10 33 56 AM

PR #1854, opened 2022-05-20
Fix geocoding error logging to include status code

maurizi:
## Overview

The `ogr/develop` version of https://github.com/open-apparel-registry/open-apparel-registry/pull/1853

## Testing Instructions

* There aren't any differences between `develop` and `ogr/develop` yet, so this shouldn't be necessary

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Removed
 
 ### Fixed
+- Fix geocode error messages to include the status code [#1853](https://github.com/open-apparel-registry/open-apparel-registry/pull/1853)
 
 ### Security
 
--- a/src/django/api/geocoding.py
+++ b/src/django/api/geocoding.py
@@ -80,7 +80,7 @@ def geocode_address(address, country_code):
     r = requests.get(GEOCODING_URL, params=params)
 
     if r.status_code != 200:
-        raise ValueError("Geocoding request failed with status"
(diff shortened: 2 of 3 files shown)


jwalgran:
Was intrigued by the testing question. I thing adding this to `api.tests.GeocodingTest` will do the trick

```python
    @patch('api.geocoding.requests.get')
    def test_geocode_non_200_response(self, mock_get):
        mock_get.return_value = Mock(ok=True, status_code=400)
        with self.assertRaisesRegexp(ValueError, '400'):
            geocode_address('Noorbagh, Kaliakoir Gazipur Dhaka 1704', 'BD')
```

TaiWilkin:
@jwalgran This is round two review (into OGR) - what would be the steps for a code change at this stage, or should we just bear that as a possible way to test in the future?

maurizi:
> This is round two review (into OGR) - what would be the steps for a code change at this stage, or should we just bear that as a possible way to test in the future?

Probably we should avoid merging PR 1 until after PR 2 is approved to assist in these cases.

In this particular case since it's just a small test case and not a functional change, I'd be in favor of only merging the test to the `ogr/develop` branch - but I'm happy to get it onto `develop` too.

jwalgran:
It would be preferable for us to have our branches diverge as little as possible so I do think it is worth it do have this test merged to both branches. No need for official review ceremony for such a simple change. We can just rubber stamp approve a PR to copy the addition back to `develop`.

jwalgran (commented):
Looks good. The test passed for me locally

jwalgran (approved):

PR #1851, opened 2022-05-18
Document parallel OGR workflow in the README

jwalgran:
## Overview

We explain the motivation behind having separate branches for OAR and OGR development during the beta period and describe how developers should submit PRs.

Connects #1842

## Testing Instructions

* Review the README for correctness

## Checklist

- [x] `fixup!` commits have been squashed
- [x] CI passes after rebase
- [x] CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines

Issue #1842, opened 2022-05-05
Create a repository to test OAR/OGR development and CI workflow


jwalgran:
## Overview

We need to develop, build, and deploy OAR and OGR in parallel until the first official release of OGR at which point it replaces OAR.

We believe adding `ogr-develop` and `ogr-master` branches to the repository will give us the parallel support we need, but we want to test new-to-us workflow in a separate throw away repository before making changes to our actual repository.

### Describe the solution you'd like

- [x] Create a `parallel-dev-toy` repo with the conventional develop and master branches.
- [x] Commit a `./scripts/cibuild` and a `Jenkinsfile`  to the `develop` branch that run a trivial CI task
- [x] Set up Jenkins to build the `parallel-dev-toy` repository
  - Add `parallel-dev-toy` to the `Filter by name` list 
- [x] Ensure that a pull request to `parallel-dev-toy` is built by Jenkins
- [x] Ensure that merging the pull request successfully builds `develop` on Jenkins
- [x] Run through a git flow release and verify that the `release/` branch builds on CI ~and the the `master` branch builds on CI when the release is completed~ We do not do builds from the master branch.
- [x] Update the organization settings in Jenkins to include `ogr/` prefixed branches

```
develop release/* PR-* test/* hotfix/* ogr/develop ogr/release/* ogr/test/* ogr/hotfix/*
```

- [x] Create and push `ogr/develop` and `ogr/master` branches from `develop` and `master` respectively.
- [x] Make a PR targeted at `ogr/develop` with modifications to  the `Jenkinsfile` and `./scripts/cibuild` to  have different behavior
- [x] Ensure that Jenkins correctly builds the PR targeted at `ogr/develop`
- [x] Merge the PR to `ogr/develop`
- [ ] Run through mock development and release workflows with both sets of branches to verify that features can be developed in parallel
- [ ] Update the Jenkins organization to remove `parallel-dev-toy` from the `Filter by name` list

PR #1856, opened 2022-05-20
Document parallel OGR workflow in the README

jwalgran:
## Overview

We explain the motivation behind having separate branches for OAR and OGR development during the beta period and describe how developers should submit PRs.

Connects #1842
Same content as #1851 but targeted at `ogr/develop`

## Checklist

- [x] CI passes after rebase


jwalgran:
Merging without review because it doc-only change using the same commit from the approved #1851

PR #1777, opened 2022-03-26
Update tile load test script and add scripts for hompage load, facility download, and facility POST

jwalgran:
## Overview

Update tile load test script and add scripts for hompage load, facility download, and facility POST

Connects #1776

## Testing Instructions

**NOTE** these scripts have been successfully used on staging so this can be a cursory review.

* Register an account on staging and generate an API token
* Obtain the `CLIENT_KEY` used on staging
* `cd` into the `load-tests` directory.
* Verify that you can run the new load test commands added to the README in the PR

## Checklist

- [x] `fixup!` commits have been squashed
- [x] CI passes after rebase
- [x] CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines

Issue #1776, opened 2022-03-25
Update and add load testing scripts


jwalgran:
There are 4 types of HTTP request activity that we want to be able to be able to simulate
- Loading the home page
- Zooming the map with a contributor filter
- Downloading a search result CSV
- POST a batch facilities API one at a time (mimicking a "bulk" upload from an API partner) 

We need to be able to
- Run simulations of all 4 types of requests simultaneously
- Adjust the number of parallel virtual users making each type of request.

PR #1857, opened 2022-05-20
 Update tile load test script and add scripts for hompage load, facility download, and facility POST

jwalgran:
Update tile load test script and add scripts for hompage load, facility download, and facility POST

Connects #1776
Same content as #1777 but targeted at ogr/develop

## Checklist

- [x] CI passes after rebase


jwalgran:
Merging without a second review because all of the code is the same as #1777 and it is all within the `load-tests` directory which is not used at runtime.

Issue #1838, opened 2022-05-05
Add a new API to return sector choices


obrienad:
## Overview

Add a new API to return sector choices

### Describe the solution you'd like

- Add a `GET /api/sectors` endpoint with a return value have a similar structure to the `product-types` endpoint, a list of distinct strings
- Values should be pulled from `FacilityListItem.sector`
- Add a doc string to populate the generated  swagger documentation

PR #1871, opened 2022-05-27
Sector choices API

maurizi:
## Overview

Adds an API to query for all values in the `FacilityListItem.sector` field.

Connects #1838 

## Demo

image
image

## Notes

This is based on top of #1868 , I'll rebase it on top of `ogr/develop` and change the base branch of the PR once that's merged.

## Testing Instructions

* `scripts/resetdb`
* `scripts/server`
* [http://localhost:8081/api/sectors](http://localhost:8081/api/sectors) should return `['Apparel']`
* Contribute a new CSV with values in the `sector` column
* [http://localhost:8081/api/sectors](http://localhost:8081/api/sectors) should now include those values as well

## Checklist

- [ ] CI passes after rebase
- ~[ ] If this PR applies to both OAR and OGR a companion PR has been created~


src/django/api/views.py
```diff
@@ -632,6 +632,29 @@ def current_tile_cache_key(request):
...
+    ## Sample Response
+
+        [
+            "Agriculture",
+            "Apparel",
+            "Industry",
+        ]
+
+    """
+    sectors = FacilityListItem \
```
jwalgran, line 651: I omitted a detail from the requirements listed in the issue. List items can be inactive or private, in which case we do not want to include the sectors in the public list

Here is a proposed update and associated test

Joining the  

```python
sectors = FacilityListItem \
    .objects \
    .filter(source__is_active=True, source__is_public=True) \
    .annotate(all_sectors=Func(F('sector'), function='unnest')) \
    .values_list('all_sectors', flat=True).distinct()
```


```python
    def test_sector_choices_inactive_and_private(self):
        sources = [
            Source(source_type=Source.SINGLE, contributor=self.source.contributor),
            Source(is_active=False, source_type=Source.SINGLE, contributor=self.source.contributor),
            Source(is_public=False, source_type=Source.SINGLE, contributor=self.source.contributor),
        ]
        Source.objects.bulk_create(sources)

        items = [FacilityListItem(row_index=i, source=sources[i],
                                  status=FacilityListItem.PARSED)
                 for i in range(3)]
        items[0].sector = ['Apparel']
        items[1].sector = ['Agriculture']
        items[2].sector = ['Mining']
        FacilityListItem.objects.bulk_create(items)

        response = self.client.get(reverse('sectors'))
        self.assertEqual(response.status_code, 200)
        self.assertEqual(response.json(), ['Apparel'])
```

jwalgran (commented):
Smart use of ORM features. This is working well. I added a proposed change for an requirement I overlooked in the original issue.

jwalgran (approved):
I was able to verify the change.

PR #1875, opened 2022-05-31
Fix processing_type_facility_type_unmatched

TaiWilkin:
## Overview

The warning we set up in the `processing_type_facility_type_unmatched` 
report to let us know about certain invalid values
in the system was triggering all the time, and not just when it should
be, due to an indentation issue.

Connects https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/1576/?item_page=0&#similar-items

## Testing Instructions

* On `ogr/develop`, add a debugger or a print statement at the Rollbar error in `processing_type_facility_type_unmatched()` and ensure you have at least one unmatched processing_type value. Ensure the unmatched value has been added to the system properly.
* Open the processing_type_facility_type_unmatched report and see that the error is triggered incorrectly.
* Repeat the above steps on this branch and see that the error is not triggered. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+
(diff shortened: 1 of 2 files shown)


jwalgran (approved):
I did not check this out an run it but because it is an isolated change to a report a visual confirmation is sufficient.

Issue #1839, opened 2022-05-05
Return sectors in API responses


obrienad:
## Overview

Include sectors submitted by for each contributor in the facility detail API response

Depends on #1830

### Describe the solution you'd like

- Pull values for each contributor from the most recent active list item
- Add values for each contributor to a "sector" property

```json
{
  "properties": {
    "sector": [
      { 
         "contributor_id": 3,
        "contributor_name": "foo",
        "values": ["Apparel", "Electronics"],      
        "updated_at":  "date string"
       },
      { 
         "contributor_id": 4,
        "contributor_name": "bar",
        "values": ["Apparel"],      
        "updated_at":  "date string"
       }
    ]
  }
```

PR #1884, opened 2022-06-06
Add sector to Facility detail API response

maurizi:
## Overview

Add sector to the Facility detail API response

Connects #1839 

## Demo
localhost_8081_api_facilities_US2022157D7WF1V_

## Testing Instructions

* `scripts/server`
* Interact with the `/api/facilities/<id>` endpoint and verify the `sector` field is working correctly

## Checklist

- [ ] CI passes after rebase


TaiWilkin (approved):
This is working well. I added several facilities with sectors to my database as multiple contributors, and confirmed that the values returned from the API were the most recent. I then deactivated one of the most recent lists and confirmed that those values were not returned.

Issue #1879, opened 2022-06-01
Update logo and base color


mariel-oar:
Summary: update logo, top banner color, map styling and "search" button on side panel (search button update is a nice to have, not necessary if it changes the point value of ticket) 

Logo: [OS-Hub_Logo_Vector.zip](https://github.com/open-apparel-registry/open-apparel-registry/files/8818325/OS-Hub_Logo_Vector.zip)

<img width="207" alt="Screen Shot 2022-06-01 at 7 56 17 AM" src="https://user-images.githubusercontent.com/97469356/171421950-ac92d1bf-94d6-4025-a71b-2910ec1b2b8a.png">

Color: 
![Screen Shot 2022-06-01 at 11 45 21 AM](https://user-images.githubusercontent.com/97469356/171469277-177fcab6-d3c0-4a46-8f9e-129a342279f0.png)

1. Update logo: on top of background color (logo background is transparent). 
2. Update colors on top banner to: green background
3. Update text on top banner to black.
4. Make map greyscale by using the "silver" map style option already supported by the VectorTileFacilityMap component.
Pending time: Update search bar color to green background and black text.

Logo is on 

Green hex: 62CC74 
Black hex: 0D1128

obrienad:
![image](https://user-images.githubusercontent.com/95237721/171508068-b2bd8b71-dde1-4f00-8728-b73ac62fd6b9.png)

obrienad:
![image](https://user-images.githubusercontent.com/95237721/171508162-d51e30b1-804e-4d5d-890f-da5b86c8e924.png)

obrienad:
@mariel-oar as a head's up: making the color changes _might_ require changes to other colors in the application to change for it to look right. We won’t know until we see it.

jwalgran:
I updated the issue description to be more explicit about the map styling

>4. Make map greyscale by using the "silver" map style option already supported by the VectorTileFacilityMap component.

PR #1900, opened 2022-06-10
1st pass of rebrand to Open Supply Hub

maurizi:
## Overview

Makes a few small tweaks to the site branding / logo for the beta period

Connects #1879 

## Demo

localhost_6543_
localhost_6543_ (1)

## Notes

I just made some guesses on how to style the login/logout/translation buttons in the navbar.

My assumption from "Make map greyscale" was that it applied to the facilities layer which had 4 blue colors (and not the basemap).
My first pass was to directly use the grayscale versions of the existing colors, but these were very hard to distinguish, so I settled on the current colors after some simple trial & error, but it's definitely quite basic.

I _could_ have made the primary button color on the entire site green to fulfill the requirements of "Changing the search button color", but based on the limited scope of the issue I left it at just the search button.

## Testing Instructions

* `scripts/server`
* Observe that the main map screen now has a green header with a grayscale map layer, and that the search button is also green

## Checklist

- [x] CI passes after rebase
- ~[ ] If this PR applies to both OAR and OGR a companion PR has been created~


jwalgran (commented):
There is some missing detail in the issue description related to "greyscale map"

We already have support for a greyscale styling of the basemap inside the VectorTileFacilitiesMap component 

https://github.com/open-apparel-registry/open-apparel-registry/blob/3637f8baac2f6fa22de21be04e3fa5f3133eb532/src/app/src/components/VectorTileFacilitiesMap.jsx#L133

This was added so that embedded map users who use an iframe have a more neutral option that may work better with their surrounding brand colors.

We should be able to switch this option on for the main map and then restyle the dots to a green gradient rather than a blue one.

maurizi:
> There is some missing detail in the issue description related to "greyscale map"

Oh this is much simpler. Rebased to remove the old commit, and switched to this in https://github.com/open-apparel-registry/open-apparel-registry/pull/1900/commits/87992bc5568d6fca131dbc2a6d10069ac5805484

maurizi:
localhost_6543_ (2)

src/app/src/components/Navbar/AuthMenu.jsx
```diff
@@ -24,13 +24,13 @@ import { convertFeatureFlagsObjectToListOfActiveFlags } from '../../util/util';
 
 const styles = {
     profileButton: {
-        color: 'white',
```
jwalgran, line 27: We have an existing `COLORS.js`  so where it makes sense we could introduce a constant and make future tweaking easier.

jwalgran (approved):
Confirmed!

src/app/src/components/Navbar/AuthMenu.jsx
maurizi: I think I actually looked for `COLORS.js` on my own once and missed it, since it's currently `COLOURS.js` 🇬🇧

PR #1904, opened 2022-06-14
Avoid exception when parsing arrays in raw data

jwalgran:
## Overview

When submitting a facility via the API the raw data for fields like `product_type` can contain an array value. Attempting to call `fload()` on a list was raiding a `TypeError` rather than a `ValueError`, resulting in an unhandled exception in the `try_parse_int_from_float` helper.

Connects #1903

## Testing Instructions

This assumes you have run `./scripts/resetdb`

* Check out develop
* Log in as c2@example.com
* Browse http://localhost:8081/api/docs/#!/facilities/facilities_create and authenticate with token 1d18b962d6f976b0b7e8fcf9fcc39b56cf278051
* POST this facility detail that matches a dev data facility
```json
{
"name":"1v1 Fight Gear",
"address": "5320 Tower Way, Sanford, 32773, Florida ",
"country":"us",
"product_type": ["api a", "api b"]
}
```
* Browse http://localhost:6543/settings, selected the embed tab, and check the "100%"  box to get the preview to display
* In the embed iframe, view the details of the "1v1 Fight Gear" facility
* Verify that an exception is raised
* Checkout this branch
* Refresh them embed and verify that the exception is resolved

## Checklist

- [x] `fixup!` commits have been squashed
- [x] CI passes after rebase
- [x] CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines
- [x] This PR is targeted at the correct branch (`develop` vs. `ogr/develop`)
- [x] If this PR applies to both OAR and OGR a companion PR has been created

Issue #1903, opened 2022-06-14
Exception raised when viewing an facility on embedded map POSTed to the API with a product type array


jwalgran:
## Overview

Exception raised when viewing an facility on embedded  map POSTed to the API with a product type array

### Steps to Reproduce

* Log in as c2@example.com
* Browse http://localhost:8081/api/docs/#!/facilities/facilities_create and authenticate with token 1d18b962d6f976b0b7e8fcf9fcc39b56cf278051
* POST this facility detail that matches a dev data facility
```json
{
"name":"1v1 Fight Gear",
"address": "5320 Tower Way, Sanford, 32773, Florida ",
"country":"us",
"product_type": ["api a", "api b"]
}
```
* Browse http://localhost:6543/settings, selected the embed tab, and check the "100%"  box to get the preview to display
* In the embed iframe, view the details of the "1v1 Fight Gear" facility
* Verify that an exception is raised
 
### Demo

![2022-06-14 10 00 20](https://user-images.githubusercontent.com/17363/173635331-001f9629-82c6-4e5b-acba-c5b5d0c353ee.gif)

```
ERROR:django.request:Internal Server Error: /api/facilities/US2022161SEBGD5/
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/viewsets.py", line 114, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 505, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 465, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 476, in raise_uncaught_exception
[...]

PR #1910, opened 2022-06-15
Avoid exception when parsing arrays in raw data

jwalgran:
## Overview

This is the same change as #1904 targeted at `ogr/develop`

Connects #1903 

## Testing Instructions

Skip testing because the change is identical in both branches.

## Checklist

- [x] CI passes after rebase


jwalgran:
Merging without review because the change is the same across branches.

Issue #1563, opened 2021-12-20
TypeError: Cannot read properties of undefined (reading 'style')


TaiWilkin:
## Overview

When navigating away from the map while a marker is present in the map, an error is thrown by something attempting to access the `style` property of an undefined element. 

### Expected Behavior

All map elements should be correctly unmounted when navigating away from the map. No error should be thrown. 

### Actual Behavior

This error only appears in development, most likely because the development debug settings produce an alert while the production settings hide it. If the fullscreen React error message is clicked away, the app behaves as expected. 

```
Uncaught TypeError: Cannot read properties of undefined (reading 'style')
    at setTransform (DomUtil.js:216)
    at NewClass.<anonymous> (Map.js:1622)
    at NewClass.fire (Events.js:190)
    at NewClass._moveEnd (Map.js:1259)
    at NewClass._resetView (Map.js:1194)
    at NewClass.setView (Map.js:206)
    at NewClass.setZoom (Map.js:218)
    at NewClass._updateZoomLevels (Layer.js:273)
    at NewClass._removeZoomLimit (Layer.js:245)
    at NewClass.onRemove (GridLayer.js:173)
    at NewClass.removeLayer (Layer.js:185)
    at NewClass.removeFrom (Layer.js:66)
    at NewClass.remove (Layer.js:59)
    at NewClass.remove (Map.js:798)
    at Map.componentWillUnmount (Map.js:270)
    at callComponentWillUnmountWithTimer (react-dom.development.js:19580)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237)
    at invokeGuardedCallback (react-dom.development.js:292)
    at safelyCallComponentWillUnmount (react-dom.development.js:19587)
    at commitUnmount (react-dom.development.js:20109)
    at commitNestedUnmounts (react-dom.development.js:20163)
    at unmountHostComponents (react-dom.development.js:20443)
    at commitDeletion (react-dom.development.js:20500)
    at commitMutationEffects (react-dom.development.js:22782)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237)
    at invokeGuardedCallback (react-dom.development.js:292)
    at commitRootImpl (react-dom.development.js:22509)
    at unstable_runWithPriority (scheduler.development.js:653)
[...]

PR #1919, opened 2022-06-17
Fix VectorGrid bug

TaiWilkin:
## Overview

We use two layers to display facilities in the map - the
VectorTileFacilitiesLayer and the VectorTileFacilityGridLayer.
The VectorTileFacilityGridLayer is the 'dots' on the map, and
is shown when the map is zoomed out to a certain level.

When the map was zoomed in far enough that the grid is hidden,
and the user navigated away from the map, an error was being thrown.
By only rendering the grid layer when it's visible, that error is
prevented.

Connects #1563 

## Testing Instructions

On `develop`:
* Run `./scripts/server` and navigate to a facility details page.
* Navigate to the login page or dashboard (or anywhere in the site away from the map).
* You will see an error message. 

On this branch:
* Repeat the steps above, and you should not see an error message.
* Zoom in and out on the map and confirm that the dots appear and disappear at the appropriate level, and in general works the same as before. 

## Checklist

- [ ] CI passes after rebase


TaiWilkin:
@jwalgran This bug wasn't in this sprint, but I found a fix for it while investigating React upgrades.

jwalgran (approved):
I confirmed the fix and did not see any regression in map behavior. Thanks for fixing this up!

Issue #1887, opened 2022-06-07
Dependabot updates black box


obrienad:
Sequence the updates based on
- Severity
- Production code vs build piipeline

maurizi:
All of the remaining updates fall into one of two categories:
 - They're a transitive dependency of `react-scripts` that will be fixed when we update to the latest version of that in #1914
 - There are 3 alerts for the package `node-fetch`, which is a transitive dependency of 2 Material-UI libraries and React Infinite
   - The Material-UI libraries in question need a major-version upgrade
   - Unfortunately React Infinite seems to be abandoned, so we can't simply upgrade it

PR #1922, opened 2022-06-17
Update xlsx to latest version

maurizi:
Resolves a few dependabot security alerts

Connects #1887 

## Notes

The bundling strategy for the library changed, which changed how we import it, but otherwise there don't seem to be any changes in behavior.

## Testing Instructions

* `scripts/update`
* `scripts/server`
* Download an XLSX file from the Facilities search sidebar - it should continue to work

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/package.json
+++ b/src/app/package.json
@@ -44,7 +44,7 @@
         "redux-persist": "5.10.0",
         "redux-thunk": "2.3.0",
         "validator": "10.11.0",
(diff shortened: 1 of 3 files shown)


TaiWilkin (approved):
Working well, thanks for the fix! I also uploaded and processed an XLSX list successfully, in addition to the listed test.

Issue #1832, opened 2022-05-03
DRAFT: Add a new parameter to the search API for sector that accepts multiple values 


obrienad:
Same OR handling as other multiple value fields

PR #1930, opened 2022-06-22
Filter facilities by sectors

emilyhu0106:
## Overview

Allow users to filter facilities by the `sectors` in Swagger and front-end. We capitalize the first letter of each query value to match the stored value, then any facilities which have any matching sector are returned.

Connects #1832 

## Demo

In Swagger:
image

In front-end:

## Note

`sector` is used as the title in the facility filter and detail sidebar, but here we use `sectors` as the parameter name to match the name in variable name returned from the facility detail API.

## Testing Instructions

* Check out this branch and run `./scripts/server`
* Log in as `c3@example.com, contribute and fully process the list [ExtendedFieldsTestListLimited - Sector.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/8954699/ExtendedFieldsTestListLimited.-.Sector.csv)
* To test in back-end: Go to [Swagger](http://localhost:8081/api/docs/#!/facilities/facilities_list), add `sectors` param (including upper- and lower- cases)
    - [x] Make sure the returned facilities are in valid sectors
* To test in front-end: go to main page, choose sectors based on the choices (new sectors choices should be added automatically as implemented in #1899), and click "Search"
    - [x] Make sure the returned facilities are in valid sectors

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/models.py
+++ b/src/django/api/models.py
@@ -1258,6 +1258,8 @@ class FacilityManager(models.Manager):
             FacilitiesQueryParams.NATIVE_LANGUAGE_NAME, None
         )
 
+        sectors = params.getlist(FacilitiesQueryParams.SECTOR)
+
         facilities_qs = FacilityIndex.objects.all()
 
         if free_text_query is not None:
@@ -1359,6 +1361,11 @@ class FacilityManager(models.Manager):
             unidecode_name = unidecode(native_language_name)
             facilities_qs = facilities_qs.filter(
                 native_language_name__icontains=unidecode_name
+            )
+
+        if len(sectors):
+            facilities_qs = facilities_qs.filter(
+                sector__overlap=sectors
             )
 
         facility_ids = facilities_qs.values_list('id', flat=True)
--- a/src/django/api/tests.py
+++ b/src/django/api/tests.py
@@ -8674,6 +8674,34 @@ class NativeLanguageNameAPITest(FacilityAPITestCaseBase):
         self.assertEquals(data['features'][0]['id'], facility_id)
 
 
+class SectorAPITest(FacilityAPITestCaseBase):
+    def setUp(self):
+        super(SectorAPITest, self).setUp()
+        self.url = reverse('facility-list')
+
+    @patch('api.geocoding.requests.get')
+    def test_search(self, mock_get):
+        mock_get.return_value = Mock(ok=True, status_code=200)
+        mock_get.return_value.json.return_value = geocoding_data
+        self.join_group_and_login()
+        facility_response = self.client.post(self.url, json.dumps({
+            'country': "US",
+            'name': "Azavea",
+            'address': "990 Spring Garden St., Philadelphia PA 19123",
+            'sector': ['Apparel', 'Beauty']
+        }), content_type='application/json')
+        facility_data = json.loads(facility_response.content)
+        facility_id = facility_data['oar_id']
+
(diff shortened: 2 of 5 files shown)


src/django/api/models.py
```diff
@@ -1,3 +1,4 @@
+from cmath import log
```
maurizi, line 1:
```suggestion
```

src/django/api/tests.py
```diff
@@ -8673,6 +8673,33 @@ def test_search(self, mock_get):
         self.assertEquals(data['count'], 1)
         self.assertEquals(data['features'][0]['id'], facility_id)
 
+class SectorAPITest(FacilityAPITestCaseBase):
```
maurizi, line 8676:
```suggestion

class SectorAPITest(FacilityAPITestCaseBase):
```

src/django/api/tests.py
```diff
@@ -8673,6 +8673,33 @@ def test_search(self, mock_get):
...
+    @patch('api.geocoding.requests.get')
+    def test_search(self, mock_get):
+        mock_get.return_value = Mock(ok=True, status_code=200)
+        mock_get.return_value.json.return_value = geocoding_data
+        self.join_group_and_login()
+        facility_response = self.client.post(self.url, json.dumps({
+            'country': "US",
+            'name': "Azavea",
+            'address': "990 Spring Garden St., Philadelphia PA 19123",
+            'sector': ['Apparel','Beauty']
```
maurizi, line 8690:
```suggestion
            'sector': ['Apparel', 'Beauty']
```

maurizi (commented):
Left some suggestions to fix the lint tripping up CI

maurizi:
> `sector` is used as the title in the facility filter and detail sidebar, but here we use `sectors` as the parameter name to match the name in variable name returned from the facility detail API.

All of the other parameter names in the API follow this same pattern, so I think this is fine 😄

maurizi (changes requested):

src/django/api/models.py
```diff
@@ -1361,6 +1364,14 @@ def filter_by_query_params(self, params):
                 native_language_name__icontains=unidecode_name
             )
 
+        if len(sectors):
+            capitalize_sectors = []
+            for sector in sectors:
+                capitalize_sectors.append(sector.capitalize())
+            facilities_qs = facilities_qs.filter(
+                sector__overlap=capitalize_sectors
+            )
```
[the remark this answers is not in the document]
emilyhu0106: I was thinking that capitalizing could help with the swagger users, since there's no options there. @jwalgran What's your suggestions?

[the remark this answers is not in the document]
jwalgran: I think that we should skip capitalization here for two reasons:
- The we call `strip` and `title` on the values that are indexed, so `capitalize` would not match in all cases. It is better to have the value processing happing in a single place, if possible.
- API users can request "valid" search values by making a `GET /api/sectors` request

To help swagger users I would recommend adding a new schema field to `FacilitiesAPIFilterBackend` and in the `description` property explain that "Values must match those returned from the GET /api/sectors endpoint." https://github.com/open-apparel-registry/open-apparel-registry/blob/b171608071e84af321d1f2705b344aac49f9a228/src/django/api/views.py#L692

maurizi (approved):
👍

Issue #89, opened 2019-01-14
Refactor Lists.jsx component


kellyi:
## Overview

Similar to the ControlPanel component described in #81, Lists.jsx blends numerous concerns and it'd be beneficial to refactor it.

In this case:

- data formatting operations should be made into testable utility functions
- some pieces of the logic rendering JSX should be split out into separate components
- some elements which are repeatedly declared in the render method should be generated by mapping over lists

kellyi:
This one should probably be done in tandem with whatever work is done to have facility lists be returned from DRF. Going to mark this as blocked for now.

PR #1931, opened 2022-06-22
Remove hard coded domain from mail.py

nanotubing:
## Overview
When a user uploads a list to be processed, they are informed via email when processing is complete. There is a function in mail.py that builds the URL back to the processed list, and adds it to the body of the email.

Previously, the domains for staging and production were hard-coded based on environment. This PR replaces that by passing the public external domain from terraform through to mail.py via an environment variable.

Connects [#89 in the client repo](https://github.com/open-apparel-registry/open-apparel-registry-clients/issues/89)

## Testing Instructions
Once this branch is deployed to staging:
* Upload a list for processing
* When the email notification is received, confirm that the domain matches the currently configured `r53_public_hosted_zone` for staging, currently https://oshstaging.openapparel.org

## Checklist
- [ ] Temporarily push this PR branch to jenkins for deployment to staging
- [ ] Drop commit that deploys this PR branch to staging
- [ ] CI passes after rebase


src/django/oar/settings.py
```diff
@@ -43,6 +43,11 @@
     raise ImproperlyConfigured(
         'Invalid BATCH_JOB_DEF_NAME provided, must be set')
 
+EXTERNAL_DOMAIN = os.getenv('EXTERNAL_DOMAIN')
+if EXTERNAL_DOMAIN is None:
```
jwalgran, line 47: I think I steered you a bit wrong in our pairing session. To avoid raising when we run the unit tests on staking we want to skip over this check
```
if EXTERNAL_DOMAIN is None and ENVIRONMENT != 'Development':
```

nanotubing: I was just rerunning this test locally, and it did succeed. That seems like a reasonable solution to not overcomplicate things.

nanotubing:
I uploaded a list and received the completion email. The link contained in the email body sent me to the correct page in the staging site. This appears to be working correctly

Screen Shot 2022-06-22 at 16 52 49

jwalgran (commented):
The change does look good. Some cleanup

- Remove the temp commit to build the PR
- Rebase on the latest ogr/develop
- This commit message needs to be fixed to match the code: "Only check for env variable when **not** in development

jwalgran (approved):

Issue #1911, opened 2022-06-16
Add sector to facility claim form and model and assciated facility indexing and serialization


jwalgran:
## Overview

Add sector to facility claim form and model and assciated facility indexing and serialization

### Describe the solution you'd like

- Add `sector` to the `FacilityClaim` model 
```
    sector = postgres.ArrayField(
        models.CharField(max_length=50, null=True, blank=True),
        help_text='The sector(s) for goods made at the facility',
    )

```
- Update the `ApprovedFacilityClaimSerializer` to return the sector values from the `FacilityClaim`
- Add an `InputSection` to  `ClaimedFacilitiesDetails` that allows entering zero or more sector values
	- Fetch the list of sectors choices from from the API and add them as selectable items
	- Use the `isCreatable` prop to allow setting arbitrary values.
- Update claim form save logic to set sectors on the `FacilityClaim`
- Update facility indexing to also read sectors from an approved claim
- Update facility details serialization to also read sectors from approved claim
	- Add `is_from_claim` boolean flag mirroring the flag used on extended profile fields

PR #1934, opened 2022-06-23
Add sector to facility claim

maurizi:
## Overview

Adds the ability to set sector on approved facility claims, and uses facility claims sectors when updating facility indexes.

Connects #1911 

## Demo

submit-next-steps

## Notes

I also added one piece of functionality not in the issue - I updated the sector choice API from #1871 to include sectors from approved facility claims.

## Testing Instructions

* `scripts/server`
* As `c3@example.com` claim a facility
* As `c1@example.com` approve the claim (I used an Incognito window)
* As `c3@example.com` update your claim to include a sector other than "Apparel". You should be able to both use the autocomplete to choose from existing sectors as well as add new ones.
* On the dashboard, search for the sector you just entered. You should see the facility you claimed.

## Checklist

- [x] CI passes after rebase


src/app/src/components/ClaimedFacilitiesDetails.jsx
```diff
@@ -381,6 +392,18 @@ function ClaimedFacilitiesDetails({
...
+                <InputSection
+                    label="Sector"
+                    value={get(data, 'sector', [])}
+                    onChange={updateSector}
+                    disabled={updating}
+                    isSelect
+                    isMultiSelect
+                    isCreatable
+                    selectOptions={sectorOptions}
+                    selectPlaceholder="Use <Enter> or <Tab> to add multiple values"
```
maurizi, line 404: Do we want anything else in the placeholder text here?

maurizi, line 404: Do we want anything else in the placeholder text here?

src/django/api/models.py
```diff
@@ -2661,48 +2661,67 @@ def index_facilities(facility_ids=list):
...
+        .values_list('values', flat=True)
+
+    sector_data = Facility \
+        .objects \
+        .filter(id__in=facility_ids) \
+        .values('id',
+                item_sectors=ArraySubquery(item_sectors, array_type),
+                claim_sectors=ArraySubquery(claim_sectors, array_type))
+    sectors = {d['id']: list({*d['item_sectors'], *d['claim_sectors']})
+               for d in sector_data}
```
maurizi, line 2715: This refactor is more verbose than the previous method, but it's a much more standard use of the ORM - the only weird thing here is `ArraySubquery`, which we'll eventually get to import directly from Django once we upgrade to 4.1+

We can also clean up the query at the same time to remove the need to merge the two arrays in Python - a recent update to Django [allows `.union()` to be used in a subquery](https://code.djangoproject.com/ticket/29338) which would let us combine the two lists in SQL.

src/django/api/serializers.py
```diff
@@ -1042,12 +1043,25 @@ def get_sector(self, facility):
...
                     contributor_id=F('source__contributor_id'),
                     contributor_name=F('source__contributor__name'),
-                    values=F('sector'))
+                    values=F('sector'),
+                    is_from_claim=Value(False, models.BooleanField()))
+        claim_sectors = FacilityClaim \
+            .objects \
+            .filter(facility=facility, status=FacilityClaim.APPROVED) \
+            .order_by('contributor_id', '-updated_at') \
+            .distinct('contributor_id') \
```
maurizi, line 1052: It looks like it isn't intended to be possible to have multiple approved `FacilityClaim` objects for the same `Facility`, but since there aren't any constraints enforcing that at the database-level I wanted to be safe here and assume it might be possible

jwalgran (approved):
Excellent work. The chain of small commits made for easy reviewing. I opened a blocked draft issue #1937 to capture work we want to do as part of a 4.2 upgrade.

Issue #1902, opened 2022-06-14
Replace use of OAR and Open Apparel Registry with OS Hub and Open Supply Hub in on-screen help text and email content


jwalgran:
## Overview

Replace use of OAR and Open Apparel Registry with OS Hub and Open Supply Hub in on-screen help text and email content

PR #1935, opened 2022-06-24
Rename OAR to OS Hub

maurizi:
## Overview

Updates any references to OAR or Open Apparel Registry _visible to a user or API consumer_ to OS Hub and Open Supply Hub respectively. This _includes_ renaming OAR ID _except_ in API responses, where it continues to be `oar_id` for now.

Connects #1902 

## Demo

localhost_6543_ (3)

## Notes

 - After discussion with @jwalgran, we decided to _not_ rename OAR -> OS Hub in any code or code comments for now, and to defer that until after the OS Hub beta period is over. The intent of this is to minimize the number of unnecessary differences between production and the beta site while we may have to target hotfixes to both branches. We will do a follow-on pull request after the beta period is over to remove all final references to OAR.
 - I made the commits renaming "OAR ID" to "OS Hub ID" separate since I wasn't sure which areas if any we might want to keep as OAR ID until we rename the API field itself

## Testing Instructions

* `scripts/server`
* Try to double-check to make sure I didn't miss any user-visible uses of OAR or Open Apparel Registry

## Checklist

- [x] CI passes after rebase


jwalgran (approved):
These changes look good. I made a few small inline suggestions that won't need another review cycle.

While reviewing I noted places where were we will need to also change to the new name/domain in future PRs. Adding a copy of these notes here just as a reference

- "Powered by the OAR logo" on the embedded map footer
- References to the email address inifo@openapparel.org
- References to openapparel.org domain in mail templates and other locations. 
	- PR # #1931 makes the domain available via Django settings)
- Dashboard reports use @openapparel.org email address user filtering and will also need to incorporate the new domain.
- src/app/package.json
- constants.js InfoLink
- settings.py 
	- Allowed hosts
	- license URL

PR #1948, opened 2022-07-01
Add script to copy/anonymize the production database onto staging

maurizi:
## Overview

This adds an interactive script meant to be run on the staging bastion, which automates the process of restoring a production database snapshot, anonymizing it, creating a dump, and restoring that dump on top of staging.

Connects #1813 

## Demo
Screenshot 2022-07-01 111901

Screenshot 2022-07-01 105902

## Notes

I ran this from the bastion.

To run it, I first had to manually install `git` and `python3`: `sudo yum install git python37`. Then I clone `open-apparel-registry` and checked out this branch and ran `./tools/copy_production_db_to_staging`.

For testing purposes, I used a copy of the staging database I spun up from a recent snapshot.

## Testing Instructions

* SSH to the staging bastion
* cd into `open-apparel-registry/` and ensure the latest version of this branch is checked out.
* Run `./tools/copy_production_db_to_staging`. You will be asked to supply staging/production database passwords, as well as confirm data determined using the AWS CLI (I substituted the staging database copy for the staging database endpoint determined by the script).
* At each step, confirm results expected for that step of the script in the AWS console or using `psql` on either the temporary RDS database or the database you will be copying onto (staging by default)

## Checklist

- [x] `fixup!` commits have been squashed
- [x] CI passes after rebase
- [x] CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines
- [x] This PR is targeted at the correct branch (`develop` vs. `ogr/develop`)
- [x] If this PR applies to both OAR and OGR a companion PR has been created

PR #1953, opened 2022-07-06
Add script to copy/anonymize the production database onto staging

maurizi:
## Overview

Companion PR to https://github.com/open-apparel-registry/open-apparel-registry/pull/1948

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/tools/copy_production_db_to_staging
+++ b/tools/copy_production_db_to_staging
@@ -1,0 +1,158 @@
+#!/bin/bash
+
+set -e
+
+if [[ -n "${OAR_DEBUG}" ]]; then
+    set -x
+fi
+
+# Constants
+STAGING_ID="openapparelregistry-enc-stg"
+PRODUCTION_ID="openapparelregistry-enc-prd"
+
+# Derived constants
+PROJECT_ROOT="$(dirname $( dirname -- "$BASH_SOURCE"; ))";
+
+function usage() {
+    echo -n \
+        "Usage: $(basename "$0")
+Clones, anonymizes & restores the production database on-top of the staging database
+"
+}
+
+function required() {
+    if [ -z "$1" ]; then
+        echo "$2 is required"
+        exit 1
+    fi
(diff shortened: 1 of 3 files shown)


tools/copy_production_db_to_staging
```diff
@@ -0,0 +1,158 @@
...
+            --db-instance-identifier $NEW_DATABASE_ID \
+            --skip-final-snapshot"
+        echo "Will run '$DELETE'"
+        read -n 1 -s -r -p "Press any key to delete $NEW_DATABASE_ID"
+        echo
+        $DELETE
+
+        # Run pg_restore
+        export PGPASSWORD="$STAGING_PGPASSWORD" 
+        PG_RESTORE="pg_restore -h $STAGING_DATABASE_ENDPOINT  -U openapparelregistry -d openapparelregistry --clean --if-exists --no-owner $DUMP_FILE"
```
jwalgran, line 148: For OS Hub the username and database name will be `opensupplyhub` so we should change them here for the `ogr/develop` branch version.

maurizi: There's also `STAGING_ID` and `PRODUCTION_ID` at the top of the script, and we reference `openapparelregistry` in a few other places as well - do we want to set this up to expect an OS Hub staging _and_ production database (assuming one will exist soon)?

maurizi: Updated _all_ references of `openapparelregistry` to `opensupplyhub` in https://github.com/open-apparel-registry/open-apparel-registry/pull/1953/commits/df17bd774c75485fb9784efdb8e4f7345e3ef4b2

jwalgran: it does make sense for the ogr/develop version of the script to use opensupplyhub everywhere. In the interim period where we use this script to restore OAR prod to OS Hub staging, we will need to override the variables accordingly. This is fine.

Issue #1959, opened 2022-07-07
Facility search resulting in "whoops" error page 


mariel-oar:
This is a claimed facility that is in the OAR. Uploaded in March 2022. 

When I try searching this facility: /facilities/VN202108625A1KW?q=VN202108625A1KW. I get a Whoops something went wrong error. I was trying to see if claimed facility data was displaying and this is a claimed facility in the OAR. It happens when I try to search it by name, too: /facilities/VN202108625A1KW?q=protrade

Based on the contribution date, it looks liek this facility should have been migrated from OAR. Is there a reason it is erroring in OS Hub? It works in OAR. 

### Proposed Fix

- In `serializers.py` exclude sectors with a null value

jwalgran:
In an example from staging (US2022117EBHAQ2) there is a null value from sectors added to a claim

<img width="544" alt="Screen Shot 2022-07-07 at 1 40 05 PM" src="https://user-images.githubusercontent.com/17363/177867501-45b9a6bf-2044-41af-9fb5-6f06e1c29bfa.png">

A null `values` causes an exception here

<img width="628" alt="Screen Shot 2022-07-07 at 1 43 48 PM" src="https://user-images.githubusercontent.com/17363/177867986-b1fdd9dd-94af-4fb8-bf67-cd00b6737a1e.png">

jwalgran:
A fix for this is deployed to OS Hub staging

PR #1963, opened 2022-07-07
Exclude null claim sector value when serializing facility details

jwalgran:
## Overview

If a claimant has not provided a sector value, we want to skip serializing facility sector array item entirely rather than serializing an item with a null value.

Connects #1959

### Reproduce the error

* Check out `ogr/develop`
* Log in as c2@example.com
* Claim a facility
* In a separate session log in as c1@example.com, browse http://localhost:6543/dashboard/claims, and approve the claim. 
* In the original session refresh the detail page for the claimed facility. You should see an error page

### Test the fix
 
* Check out this PR branch
* Refresh the facility details page and verify that it loads without error.

## Checklist

- [ ] CI passes after rebase
- [ ] ~If this PR applies to both OAR and OGR a companion PR has been created~


maurizi (approved):
testing, working👍

At a broader level beyond fixing this bug - should we reconsider _having_ nullable array fields? Maybe this would be better represented as an empty array instead.

Issue #1958, opened 2022-07-07
Sector field blank on some facilities 


mariel-oar:
All facilities moved over from OAR should have been assigned a sector value of Apparel. All new facilities are blocked if sector is not included. 

There are a number of facilities (look like they are all apparel facilities that are showing up with a blank sector value): 
CN20211821GPP86
MA20202680RFD9Z
TN2020268Y889EV

## Proposed Fix
- Do not filter out inactive sources or matches
- Sort by active sources above inactive sources, then sort by date

jwalgran:
All three of the example facilities have anonymized contributor details. 

The items associated with these facilities are all inactive

I suspect that we are filtering out inactive items when fetching the sectors for a facility

```
opensupplyhub=> select i.id, sector, is_public, is_active from api_facilitylistitem i join api_source s on i.source_id = s.id  where facility
_id = 'MA20202680RFD9Z';
   id   |  sector   | is_public | is_active
--------+-----------+-----------+-----------
 141607 | {Apparel} | t         | f
 142170 | {Apparel} | t         | f
(2 rows)

opensupplyhub=> select i.id, sector, is_public, is_active from api_facilitylistitem i join api_source s on i.source_id = s.id  where facility_id = 'CN20211821GPP86';
   id   |  sector   | is_public | is_active
--------+-----------+-----------+-----------
 536590 | {Apparel} | f         | f
 540863 | {Apparel} | t         | f
(2 rows)

opensupplyhub=> select i.id, sector, is_public, is_active from api_facilitylistitem i join api_source s on i.source_id = s.id  where facility_id = 'TN2020268Y889EV';
   id   |  sector   | is_public | is_active
--------+-----------+-----------+-----------
 141610 | {Apparel} | t         | f
 142173 | {Apparel} | t         | f
(2 rows)
```

Issue #1949, opened 2022-07-05
Discrepancy with total # facilities on the OS Hub staging site 


mariel-oar:
On OS Hub staging there are ~85,000 in total, but when you filter to only apparel sector value it drops to 75752.

Justin response / context: Searching comes from an index of the sector information. This may need to investigate the root cause to update the indexing of that particular item 

cc: @obrienad @jwalgran 

---

The 1-point estimate is to research if reindexing will resolve the issue. If there is a larger underlying problem create a followup issue with a separate estimate.

jwalgran:
#1966 has been deployed and I ran the new `index_sectors` management command. These numbers are much closer now

No filter: 89149
Selecting all available sectors: 89092

I expect the additional change as part of #1964 to address the remainder

jwalgran:
The discrepancy in the number does match the count of facilities that do not have any sector in the index.

```

opensupplyhub=> select count(*) from   api_facilityindex  where sector = '{}' and id in (select id from api_facility);
 count
-------
    57
(1 row)
```

PR #1966, opened 2022-07-08
Return anonymized sectors and index them instead of excluding them when items are private or inactive 

jwalgran:
## Overview

Previously if all sources for a facility were private or inactive then no sector details were being serialized. To fix this we follow the patter of other private or inactive data and anonymize the contributor information during serialization rather than excluding it.

We also extends this change to the facility indexing process, introducing similar filtering and sorting changes to ensure that the index is updated with the correct sectors.

Since we are updating the way the list of sectors is calculated for a facility we need to update all sector values in the index. We add a new `index_sectors` management command do a targeted update of the sector column.

Connects #1949 
Connects #1958

## Demo

Screen Shot 2022-07-08 at 3 20 26 PM

### Setup

* Run `./scripts/resetdb`

### Test serialization

* Log in as c2@example.com
* Contribute [sector-test-1.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/9075351/sector-test-1.xlsx) and run `./tools/batch_process 16` 
* Browse http://localhost:6543/?q=sector%20test and verify the two facilities loaded properly
* Contribute [sector-test-2.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/9075374/sector-test-2.xlsx) **making sure** to select the previous list in the "Select a list to replace" box
* Run `./scripts/manage batch_process -l 17 -a parse`
* Browse http://localhost:6543/?q=sector%20test and verify:
  *  Sector Test Facility One still has the sectors `Automotive, Agrigulture`
  *  Sector Test Facility Two displays a sector but the attribution is anonymized to "a Service Provider"
* Run `./scripts/manage batch_process -l 17 -a geocode` and `./scripts/manage batch_process -l 17 -a match`
* Browse http://localhost:6543/?q=sector%20test and verify:
  *  Sector Test Facility One now has the sector `Foodservice`
* Contribute [sector-test-3.xlsx](https://github.com/open-apparel-registry/open-apparel-registry/files/9075387/sector-test-3.xlsx) but **do not** select a previous list for replacement
* Run `./tools/batch_process 18` 
* Browse http://localhost:6543/?q=sector%20test and verify:
  *  Sector Test Facility One now has the sector `Technology`
* Browse http://localhost:8081/api/docs/#/ and authenticate with `Token 1d18b962d6f976b0b7e8fcf9fcc39b56cf278051` 
* Use the POST /api/facilities endpoint to submit a private facility (**making sure** to set `public` to `false`)
```
{
    "sector": "Transportation",
    "country": "US",
    "name": "Sector Test API",
    "address": "777 main st pittsburgh pa"
}
```
* Browse http://localhost:6543/?q=sector%20test and verify:
  *  Sector Test Facility Three now has the sector `Transportation` with an anonymized contributor name

### Test reindexing

* From `./scripts/manage dbshell` clear the sector sector column with `UPDATE api_facilityindex  SET sector = '{}';`
* Verify a sector search no longer returns facilities http://localhost:6543/facilities?sectors=Apparel
* Run `./scripts/manage index_sectors` and  `./scripts/manage incrementtileversion`
* Verify that sector search is fixed http://localhost:6543/facilities?sectors=Apparel
## Checklist

- [x] CI passes after rebase
- [ ] ~If this PR applies to both OAR and OGR a companion PR has been created~

### DIFF
--- a/src/django/api/models.py
+++ b/src/django/api/models.py
@@ -2659,6 +2659,43 @@ class ArraySubquery(Subquery):
     template = 'ARRAY(%(subquery)s)'
 
 
+def get_sector_dict(facility_ids):
+    item_sectors = FacilityMatch \
+        .objects \
+        .filter(facility_id=OuterRef('pk')) \
+        .filter(Q(is_active=True)
+                & Q(status__in=[FacilityMatch.AUTOMATIC,
+                                FacilityMatch.CONFIRMED,
+                                FacilityMatch.MERGED])) \
+        .order_by('facility_list_item__source__contributor_id',
+                  '-facility_list_item__source__is_active',
+                  '-updated_at') \
+        .annotate(values=Func('facility_list_item__sector',
+                              function='unnest')) \
+        .distinct('facility_list_item__source__contributor_id') \
+        .values_list('values', flat=True)
+
+    array_type = postgres.ArrayField(models.CharField())
+    claim_sectors = FacilityClaim \
+        .objects \
+        .filter(facility_id=OuterRef('pk')) \
+        .filter(status=FacilityClaim.APPROVED) \
+        .annotate(values=Func('sector', function='unnest')) \
+        .distinct() \
+        .values_list('values', flat=True)
+
+    if len(facility_ids) > 0:
+        facility_qs = Facility.objects.filter(id__in=facility_ids)
+    else:
+        facility_qs = Facility.objects.all()
+    sector_data = facility_qs \
+        .values('id',
+                item_sectors=ArraySubquery(item_sectors, array_type),
+                claim_sectors=ArraySubquery(claim_sectors, array_type))
+    return {d['id']: list({*d['item_sectors'], *d['claim_sectors']})
+            for d in sector_data}
+
+
 @transaction.atomic
 def index_facilities(facility_ids=list):
     # If passed an empty array, create or update all existing facilities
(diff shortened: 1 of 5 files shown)


src/django/api/serializers.py
```diff
@@ -1034,36 +1034,62 @@ def get_created_from(self, facility):
...
+
+        claim_sectors = [
+            {
+                'updated_at': c.updated_at,
+                'contributor_id': get_contributor_id(
+                    c.contributor, user_can_see_detail),
+                'contributor_name': get_contributor_name(
+                    c.contributor, user_can_see_detail),
+                'values': c.sector,
+                'is_from_claim': False
```
maurizi, line 1086:
```suggestion
                'is_from_claim': True
```

src/django/api/models.py
```diff
@@ -2659,6 +2659,43 @@ class ArraySubquery(Subquery):
...
+        .filter(Q(is_active=True)
+                & Q(status__in=[FacilityMatch.AUTOMATIC,
+                                FacilityMatch.CONFIRMED,
+                                FacilityMatch.MERGED])) \
+        .order_by('facility_list_item__source__contributor_id',
+                  '-facility_list_item__source__is_active',
+                  '-updated_at') \
+        .annotate(values=Func('facility_list_item__sector',
+                              function='unnest')) \
+        .distinct('facility_list_item__source__contributor_id') \
```
maurizi, line 2675: We didn't previously do this `distinct('...__contributor_id')` in `index_facilities` - instead we did `distinct` over the full unnested list of sectors for each facility. Are we OK dropping any active matches for this contributor that aren't the most recent?

maurizi (approved):
👍 tested & working according to testing instructions

Just one minor typo to fix, and a question that's really more about how the project works in general than a need to change anything.

src/django/api/models.py
jwalgran: Good question. We keep immutable history of what was submitted in our FacilityListItem table so we can always get back to it if wee need it. Indexing by only the most recent values submitted by each contributor allows the contributor to remove a mistakenly submitted value and ensures that mistake is not permanently searchable. If, over the course of the beta period the OS Hub team gets feedback that the old values should remain searchable for some reason, we can rework this.

src/django/api/serializers.py
jwalgran: Addressed in fixup 36c9943

PR #1968, opened 2022-07-12
Ensure inactive matches have anonymized sectors

jwalgran:
## Overview

Prior to this PR the attribution for sectors was not being correctly anonymized if the match record because inactive and those inactive match records were being excluded from the facility indexing process, resulting in
FacilityIndex records with an empty sector value.

Connects #1949

## Demo

TODO

## Testing Instructions

* TODO
* 
## Checklist

- [ ] CI passes after rebase


src/django/api/views.py
```diff
@@ -1547,6 +1548,11 @@ def create(self, request):
             result['oar_id'] = item.facility.id
             if item.facility.created_from == item:
                 result['status'] = FacilityListItem.NEW_FACILITY
+            # and insure that the sector index is updated since the indexed
```
maurizi, line 1551:
```suggestion
            # and ensure that the sector index is updated since the indexed
```

maurizi (approved):
Looks good, tested👍

src/django/api/views.py
jwalgran: Fixed in a rebase. Thanks.

jwalgran:
Thanks for the review.

Issue #1942, opened 2022-06-29
Update colors on OS Hub page


mariel-oar:
**Headline bar and search**
Top bar/headline and search button updated to yellow (Hex code # F C C F 3 F) 

**Map dots** 
Old image
![Screen Shot 2022-06-29 at 10 07 12 AM](https://user-images.githubusercontent.com/97469356/176483769-8e59b38d-576a-4f43-b560-0f0cbcd67daf.png)

Update blue dot colors to the following: 
- Darkest dot: Hex code # 8 4 2 8 FA
- Next darkest dot: Hex code # a868fb
- Next lightest dot: Hex code # cda9fd
- Lightest dot: Hex code # f2e9fe

For reference

<img width="929" alt="Hues " src="https://user-images.githubusercontent.com/95237721/176952142-9e338e3f-5991-428f-a0d5-4d96c95c7a4b.png">

hlennett:
@mariel-oar One thought I just had - to differentiate the OS Hub Beta, do we also want to update the font to Darker Grotesque to match our new brand? 

cc @katieashaw

mariel-oar:
I would rather keep this ticket small since we will be doing an overhaul of the UX later on. I'm not sure if it would be throw away work to change the font now, but that would be my main concern. @jwalgran @obrienad, thoughts on that?

mariel-oar:
ok, UX changes will definitely take font into account so let's hold on that.

obrienad:
Got it.

PR #1938, opened 2022-06-27
Update favicon images with OS Hub logo

jwalgran:
## Overview

We modifeid the exstiting files using the logo and color previously added to the app header. This was a quick copy/paste job to get the beta launched.

Connects #1879

### Before

Screen Shot 2022-06-27 at 1 21 56 PM

### After

Screen Shot 2022-06-27 at 1 21 37 PM

## Testing Instructions

* Verify the new favicon loads

## Checklist

- [x] `fixup!` commits have been squashed
- [x] CI passes after rebase
- [x] CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines
- [x] This PR is targeted at the correct branch (`develop` vs. `ogr/develop`)
- [ ] ~If this PR applies to both OAR and OGR a companion PR has been created~

PR #1974, opened 2022-07-18
Update header and map colors

maurizi:
## Overview

Updates the header color from yellow to green & the map colors from blue to purple

Connects #1942 

## Demo

Before:
before

After:
after

## Testing Instructions

* `scripts/server`
* Go to the main map mapge

## Checklist

- [ ] CI passes after rebase


src/app/src/components/FilterSidebarSearchTab.jsx
```diff
@@ -301,7 +301,7 @@ function FilterSidebarSearchTab({
             className={classes.font}
             style={{
                 boxShadow: 'none',
-                backgroundColor: COLOURS.GREEN,
+                backgroundColor: COLOURS.HEADER,
```
jwalgran, line 304: Good idea to switch this to a "class-style" name rather than a literal color name however we are now using a color called `HEADER` outside of the header. What do you think about a name like `PRIMARY` or `NAVIGATION` instead?

jwalgran (approved):
I confirmed the color changes and made a small naming suggestion.

The favicon images updated in #1938 now have a mismatched background color. If we have time in the orginal 1-point estimate I would suggest repeating the icon update as part of this PR.

Screen Shot 2022-07-18 at 11 19 23 AM

jwalgran (approved):
I pushed up 4eb5950 with updated favicon files 👍

Issue #1796, opened 2022-04-12
Facility with no contributor


hlennett:
A couple of examples of facilities that have no Contributor section: 
https://openapparel.org/facilities/CN202031463D4MB
https://openapparel.org/facilities/DE20220774ZVHM0

The first is the first facility that shows up when you just look at our full database in alphabetical order, so it might be good to address. 

It looks like post-EP launch, anonymized contributors are no longer showing up in the contributor section, so if there are only anonymized contributors on a profile, there's no contributor section. Is this the behavior we want to have? One of these facilities came up in a demo and the stakeholders were a bit confused when they saw it, since our model is showing who is connected to facilities. 

## Technical Implementation

- Use the previous anonymous contributor handling from the old sidebar to the new sidebar.
  - Continue to not not show other contributors in embed mode

mariel-oar:
Thanks @hlennett (model = wireframes? or data model?)

hlennett:
Oh like our mission as an organization.

mariel-oar:
haha yes, got it (thought you might have been saying that there was something public that did show the contributor associations although the profile page wasn't displaying)

mariel-oar:
More examples: 
https://openapparel.org/facilities/IN20190855SBCEK
https://openapparel.org/facilities/CN20211821GPP86

hlennett:
@mariel-oar Was this meant to be closed?

mariel-oar:
no it was not! thank you @hlennett

mariel-oar:
@hlennett , we can see who contributed here: https://openapparel.org/api/docs/#!/facilities/facilities_get_facility_history

looks like both were anonymous and via API. do we want to change this pattern or wait for page changes with OGR?

hlennett:
@mariel-oar I'd love to discuss whether we think it's disconcerting for users or not. As an open data platform, if there's no contributor, is that too jarring? At the very least, we can make an FAQ about it, should we decide not to change.

mariel-oar:
Estimate it for after beta launch, but might need to wait for larger OS Hub pattern updates if it's too big of an estimate (seems like it should be small).

PR #1993, opened 2022-07-26
Fix anonymous contributors section on sidebar

maurizi:
## Overview

Restores how we used to handle anonymous contributors like in https://github.com/open-apparel-registry/open-apparel-registry/blob/cf827be88c158414c075aa0aeab252c4266db305/src/app/src/components/FacilityDetailSidebarInfo.jsx

Connects #1796

## Demo

image

## Notes

I assume we originally only had `id` and `name` and at some point switched to `id` & `contributor_name` _or_ `name`, and that change didn't get reflected in the new sidebar?

## Testing Instructions

* Find a facility, and verify that it has 1 line for the contributors section
* Using the Django shell, set the `FacilityMatch` for that facility to be `in_active`: `FacilityMatch.objects.filter(facility_id='...').update(is_active=False)`
* Refresh page showing the facility detail - on `develop` the contributors section will be hidden, but on this branch it should still show, but w/o linking to a contributors detail page

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Removed
 
 ### Fixed
+- Fix showing anonymous contributors on sidebar [#1993](https://github.com/open-apparel-registry/open-apparel-registry/pull/1993)
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


TaiWilkin (approved):
Great work! Good to go.

Issue #1989, opened 2022-07-25
Add an Event model


jwalgran:
## Overview

This new model will be the foundation of the change notification system and a revised history API

### Describe the solution you'd like

- Read about [Generic relations](https://docs.djangoproject.com/en/2.2/ref/contrib/contenttypes/#generic-relations-1) and verify that the content types framework is appropriate for this use case
  - We don't want direct foreign keys because events may point to model rows that have since been deleted 
- Add an `Event` model and associated migration
  - `content_type`  FK to `ContentType` (see Django docs)
  - `object_id` `PositiveIntegerField` (see Django docs)
  - `event_type` string choice field. Reference the existing history API event types
  - content_object  `GenericForeignKey('content_type', 'object_id')` (see Django docs)
  - event_time UTC datetime
  - event_details `jsonb`
  - created_at
  - updated_at

### Additional Context

Notification feature narrative: https://docs.google.com/document/d/1LHGGKBwP2u7BUkKHf3fr9PaY_05sUMy5Ax0cXWvEdCc

PR #1995, opened 2022-07-26
Add event model

TaiWilkin:
## Overview

Adds the Event model and corresponding migration. This new model will
be the foundation of the change notification system and a revised
history API. It makes use of contenttypes to allow us to use a generic
foreign field and to delete the associated objects without losing
the Event object or the ID of the deleted object.

Connects #1989 

## Demo

## Notes

## Testing Instructions

* `./scripts/manage migrate`
* `./scripts/manage shell_plus`
* In the shell:
```
from datetime import datetime
facility = Facility.objects.first()
event = Event.objects.create(content_object=facility, event_type='OTHER', event_details='testing', event_time=datetime.now())
event
event.content_object
event_2 = Event.objects.create(content_object=FacilityListItem.objects.first(), event_type='OTHER', event_details='testing', event_time=datetime.now())
event_2
event_2.content_object
```
* You should see events printed to the console, and the related Facility and FacilityListItem objects printed as well. Note the ID of the Facility.
* Close the shell. Run `./scripts/server` and login to the site as c1@example.com. Use the [Dashboard Facility Delete](http://localhost:6543/dashboard/deletefacility) to delete the facility.
* Restart `shell_plus`. In the shell:
```
event = Event.objects.get(id=1)
event
event.content_object
event.object_id
```
* You should see the event printed and the ID of the facility printed, but no Facility object should be printed. There should be no error. 

## Checklist

- [x] CI passes after rebase


src/django/api/models.py
```diff
@@ -2490,6 +2492,36 @@ def post_save(sender, **kwargs):
...
+    content_object = GenericForeignKey('content_type', 'object_id')
+    event_type = models.CharField(max_length=50, choices=EVENT_TYPE_CHOICES)
+    event_time = models.DateTimeField()
+    event_details = models.JSONField()
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    def __str__(self):
+        return "{} ({}) - {} ({})".format(
+            self.content_type.name, self.object_id, self.event_type, self.id)
```
maurizi, line 2522: We should add a `Meta` class w/ a multi-column index combining the `content_type` and `object_id` for the generic FK, per the documentation:

> Unlike for the [ForeignKey](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.ForeignKey), a database index is not automatically created on the [GenericForeignKey](https://docs.djangoproject.com/en/4.0/ref/contrib/contenttypes/#django.contrib.contenttypes.fields.GenericForeignKey), so it’s recommended that you use [Meta.indexes](https://docs.djangoproject.com/en/4.0/ref/models/options/#django.db.models.Options.indexes) to add your own multiple column index. This behavior [may change](https://code.djangoproject.com/ticket/23435) in the future.

src/django/api/models.py
```diff
@@ -2490,6 +2492,36 @@ def post_save(sender, **kwargs):
...
+        ('DELETE', 'DELETE'),
+        ('MERGE', 'MERGE'),
+        ('SPLIT', 'SPLIT'),
+        ('MOVE', 'MOVE'),
+        ('OTHER', 'OTHER'),
+        ('ASSOCIATE', 'ASSOCIATE'),
+        ('DISSOCIATE', 'DISSOCIATE'),
+        ('CLAIM', 'CLAIM'),
+        ('CLAIM_UPDATE', 'CLAIM_UPDATE'),
+        ('CLAIM_REVOKE', 'CLAIM_REVOKE'),
```
maurizi, line 2508: Is it just coincidence that these match this constant?
https://github.com/open-apparel-registry/open-apparel-registry/blob/7833460527f082bfe8c77e85582347c03a368e7e/src/django/api/constants.py#L73-L85

If they're unrelated but just have similar choices the way things are is fine, but if they're related to each other (i.e. these are events for those actions) then I'd suggest directly using the same constants here:

```python
    EVENT_TYPE_CHOICES = (
        (FacilityHistoryActions.CREATE, FacilityHistoryActions.CREATE)
        ...,
    )
```

src/django/api/models.py
```diff
@@ -2490,6 +2492,36 @@ def post_save(sender, **kwargs):
...
+    )
+
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.CharField(max_length=32)
+    content_object = GenericForeignKey('content_type', 'object_id')
+    event_type = models.CharField(max_length=50, choices=EVENT_TYPE_CHOICES)
+    event_time = models.DateTimeField()
+    event_details = models.JSONField()
```
maurizi, line 2516: Most other models we specify `help_text` - is there a reason to skip it for this one?

maurizi (changes requested):
This looks mostly good - I looked over the model field values which seem fine to me.

I left a couple of comments for code cleanups, and I'd like to make sure we add a multi-column index for the generic FK to avoid potential performance issues.

src/django/api/models.py
TaiWilkin: Thanks for catching that in the docs. 
I experimented with adding indexes in different ways, and then searching for an Event and using `EXPLAIN` on the SQL query.
```
facility_type = ContentType.objects.get_for_model(Facility)
Event.objects.filter(content_type=facility_type, object_id=facility.id)
```
When using individual indexes, 
```
indexes = [
            models.Index(fields=['content_type', 'object_id'],
                         name='api_event_content_object_idx'),
        ]
 ```
 Postgres searched through the content_type index and then separately filtered the object_id without using the index. 
 
 When using `index_together`, however, Postgres did one index query of the of content_type and object_id together. This seems like it will be more performant for large numbers of events, so I went with the `index_together`.
 As an aside, we currently can't filter by the content_object: `Event.objects.filter(content_object=facility)`. We could add this by using [Reverse Generic Relations](https://docs.djangoproject.com/en/4.0/ref/contrib/contenttypes/#reverse-generic-relations) to the models that we expect to be referenced in Events (such as Facility), however my research suggests this might have performance implications for the related objects (eg, Facilities). In addition, since we want to maintain the reference to the related object even if it has been destroyed, I suspect filtering that way would have the potential to cause errors - disallowing it entirely seems safer unless we determine that we have a strong need for it in the future.

File: src/django/api/models.py (hunk 2 above)
Comment (TaiWilkin) [reply src/django/api/models.py]: The Event model is intended to eventually replace the Facility History API, so it is using the same actions. I've switched to using the constants.

File: src/django/api/models.py (hunk 3 above)
Comment (TaiWilkin) [reply src/django/api/models.py]: Help text added - thanks!

**Comment by TaiWilkin** [comment]
@maurizi Ready for another look. I rebased on `ogr/develop` to get the new changes from there, but the changes to the code content here are still separated out into a fixup.

File: src/django/api/models.py (hunk 1 above)
Comment (maurizi) [reply src/django/api/models.py:2548]: Sometimes accidentally looking at the new version of the docs instead of the one we're on helps out (usually it just makes me wish I could use the fancy new features)👍 

Interesting results from trying things out - I was under the impression that `index_together` was just syntactic sugar for `indexes` option... makes me curious what the difference in SQL generated to create the index was (if there was any difference at all - sometimes SQL experiments like the one you tried above can be heavily influenced by things like how much data is present and statistics from previous runs).

**Review by maurizi — APPROVED** [review]

File: src/django/api/models.py (hunk 1 above)
Comment (TaiWilkin) [reply src/django/api/models.py:2548]: The SQL was the same:
```
openapparelregistry=# explain SELECT "api_event"."id", "api_event"."content_type_id", "api_event"."object_id", "api_event"."event_type", "api_event"."event_time", "api_event"."event_details", "api_event"."created_at", "api_event"."updated_at" FROM "api_event" WHERE ("api_event"."content_type_id" = 19 AND "api_event"."object_id" = 'AR20222082V6R1Z');
```

Here are the EXPLAIN results, actually, for reference:
`indexes`
                                              QUERY PLAN                                  

------------------------------------------------------------------------------------------

------------

 Index Scan using api_event_content_type_id_de4b477a on api_event  (cost=0.15..8.17 rows=1 width=264)

   Index Cond: (content_type_id = 19)

   Filter: ((object_id)::text = 'AR20222082V6R1Z'::text)

`index_together`
                                                     QUERY PLAN                           

------------------------------------------------------------------------------------------

--------------------------

 Index Scan using api_event_content_type_id_object_id_b046420a_idx on api_event  (cost=0.15..8.17 rows=1 width=264)

   Index Cond: ((content_type_id = 19) AND ((object_id)::text = 'AR20222082V6R1Z'::text))

Comment (maurizi) [reply src/django/api/models.py:2548]: > The SQL was the same:

Ah - I meant the SQL for the `CREATE INDEX` statements. One way to find that is by running `./scripts/manage sqlmigrate ... ...`:

This is from your PR:
```shell
src/open-apparel-registry [ ./scripts/manage sqlmigrate api 0098
[+] Running 1/0
 ⠿ Container open-apparel-registry-database-1  Running 0.0s
/usr/local/lib/python3.7/site-packages/thefuzz/fuzz.py:11: UserWarning: Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning
  warnings.warn('Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning')
BEGIN;
--
-- Create model Event
--
CREATE TABLE "api_event" ("id" serial NOT NULL PRIMARY KEY, "object_id" varchar(32) NOT NULL, "event_type" varchar(50) NOT NULL, "event_time" timestamp with time zone NOT NULL, "event_details" jsonb NOT NULL, "created_at" timestamp with time zone NOT NULL, "updated_at" timestamp with time zone NOT NULL, "content_type_id" integer NOT NULL);
ALTER TABLE "api_event" ADD CONSTRAINT "api_event_content_type_id_de4b477a_fk_django_content_type_id" FOREIGN KEY ("content_type_id") REFERENCES "django_content_type" ("id") DEFERRABLE INITIALLY DEFERRED;
CREATE INDEX "api_event_content_type_id_de4b477a" ON "api_event" ("content_type_id");
CREATE INDEX "api_event_content_type_id_object_id_b046420a_idx" ON "api_event" ("content_type_id", "object_id");
COMMIT;
```

This is w/ switching to using `models.Index(fields=["content_type", "object_id"])`

```shell
src/open-apparel-registry [ ./scripts/manage sqlmigrate api 0098
[+] Running 1/0
 ⠿ Container open-apparel-registry-database-1  Running 0.0s
/usr/local/lib/python3.7/site-packages/thefuzz/fuzz.py:11: UserWarning: Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning
  warnings.warn('Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning')
BEGIN;
--
-- Create model Event
--
CREATE TABLE "api_event" ("id" serial NOT NULL PRIMARY KEY, "object_id" varchar(32) NOT NULL, "event_type" varchar(50) NOT NULL, "event_time" timestamp with time zone NOT NULL, "event_details" jsonb NOT NULL, "created_at" timestamp with time zone NOT NULL, "updated_at" timestamp with time zone NOT NULL, "content_type_id" integer NOT NULL);
--
-- Create index api_event_content_598f3a_idx on field(s) content_type, object_id of model event
--
CREATE INDEX "api_event_content_598f3a_idx" ON "api_event" ("content_type_id", "object_id");
ALTER TABLE "api_event" ADD CONSTRAINT "api_event_content_type_id_de4b477a_fk_django_content_type_id" FOREIGN KEY ("content_type_id") REFERENCES "django_content_type" ("id") DEFERRABLE INITIALLY DEFERRED;
CREATE INDEX "api_event_content_type_id_de4b477a" ON "api_event" ("content_type_id");
COMMIT;
```

Which looks like it makes the same compound index but the order the indexes are created is different.

TaiWilkin: Ah, I see! That's very interesting. It's surprising that it handles the actual query differently when it looks like just the order of indexing is different.

Issue #1908, opened 2022-06-15
Upgrade Django


TaiWilkin:
## Overview

Upgrade Django to version 3.2.0

### Describe the solution you'd like

- Upgrade Django to 3.2.0 in the Dockerfile
- Update requirements.txt (note: the versions listed below are based on my investigation, but you might also want to consult [this list of requirements](https://github.com/azavea/iow-boundary-tool/pull/9/commits/9d278c7e84d63f9147b382d23b2ab97df627ed21)):
     - Upgrade django-simple-history to at least 3.0.0 (latest is 3.1.1) https://django-simple-history.readthedocs.io/en/latest/#changes
     - Upgrade django-waffle to at least 2.2.0 (latest is 2.5.0) https://github.com/django-waffle/django-waffle/releases?page=1
     - Upgrade djangorestframework-gis to at least 0.18 (latest is 1.0.0) https://github.com/openwisp/django-rest-framework-gis
     - Upgrade Watchman to 1.1.1
     - Upgrade django-spa to 0.3.5
- Update code to fix warnings and breaking changes:
     - Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'. Configure the DEFAULT_AUTO_FIELD setting.
     - django.contrib.postgres.fields.JSONField is deprecated. Use django.db.models.JSONField instead.
     - The FILE_CHARSET setting is deprecated. Starting with Django 3.1, all files read from disk must be UTF-8 encoded. Remove this setting and make any necessary related changes.
     - Replace .changeReason with ._change_reason when working with history objects

### Additional Context

`django-rest-auth appears` to still work with the upgrades, but hasn't been updated in years. A currently-maintained fork of the project is https://github.com/iMerica/dj-rest-auth and we might want to switch.

PR #1996, opened 2022-07-26
Fix lint / import issues

maurizi:
## Overview

I merged #1965 w/o rebasing it onto the latest from `develop`, so it failed to catch these two issues:
 - An indentation problem that our new linter version now catches
 - A use of `datetime.now()` instead of `timezone.now()` (the latter returns timezone-aware datetime objects)

Connects #1908 

## Testing Instructions

* CI should pass

## Checklist

- [ ] CI passes after rebase


jwalgran (approved):
Looks good (did not checkout locally)

Issue #2007, opened 2022-07-28
List uploads failing on OS Hub staging due to an exception while createting AWS Batch jobs


jwalgran:
https://rollbar.com/OpenApparelRegistry/OpenApparelRegistry/items/1677/

ClientException: An error occurred (ClientException) when calling the SubmitJob operation: Error executing request, Exception : Job name should match valid pattern

<img width="1049" alt="Screen Shot 2022-07-28 at 1 05 59 PM" src="https://user-images.githubusercontent.com/17363/181627506-509edab8-4bda-43c8-bbbc-9afe77565446.png">

PR #2008, opened 2022-07-28
Replace + character in ISO date string when in batch job name

jwalgran:
## Overview

When switching to timezone-aware datetime objects we ended up attempting to create a batch job name with a + character, which AWS Batch does not allow.

Connects #2007
Connects  #1908

## Testing Instructions

I pushed copy of this branch prefixed with `ogr/test` so it could be tested on staging.

After deploying this branch to staging I was able to upload a list

## Checklist

- [x] CI passes after rebase


maurizi (approved):
👍

jwalgran:
Thanks for the review.

Issue #2010, opened 2022-07-29
Decide on a path forward for thefuzz dependency


nanotubing:
Library [thefuzz](https://github.com/seatgeek/thefuzz), by default has a dependency on Sequencemacher, from the difflib module. While using it, this module will print out a warning that this is a slow implementation, and recommends using [python-Levenshtein](https://github.com/ztane/python-Levenshtein) instead.

This warning has been suppressed in #2009.

We should decide on whether a more permanent solution makes sense for this project. These two options seem possible so far:
1. The recommended module python-Levenshtein is an older module written in C, and they are asking for maintainers, so there is doubt surrounding whether or not this project is actively maintained, and how long it will continue to build:
> I am looking for a new maintainer to the project as it is apparent that I haven't had the need for this particular library for well over 7 years now, due to it being a C-only library and its somewhat restrictive original license.

2. There is also an open [Pull Request](https://github.com/seatgeek/thefuzz/pull/10) to replace python-Levenshtein with rapidfuzz, which appears to be an actively maintained, more modern implementation

maxbachmann:
> The recommended module python-Levenshtein is an older module written in C, and they are asking for maintainers, so there is doubt surrounding whether or not this project is actively maintained, and how long it will continue to build:

I actively maintain a fork of the `python-Levenshtein` project: https://github.com/maxbachmann/Levenshtein

> There is also an open https://github.com/seatgeek/thefuzz/pull/10 to replace python-Levenshtein with rapidfuzz, which appears to be an actively maintained, more modern implementation

This is probably the simpler variant, since it just requires you to update the import

Issue #2009, opened 2022-07-29
Suppress "pure-python SequenceMatcher" warning


nanotubing:
when using [thefuzz](https://github.com/seatgeek/thefuzz), this warning is output very frequently from scripts such as `resetdb`, but the performance issue does not cause a serious problem based on our use of this library. We should suppress this warning while we decide on whether a more permanent solution makes sense for this project.

`/usr/local/lib/python3.7/site-packages/thefuzz/fuzz.py:11: UserWarning: Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning`

rajadain:
Possible duplicate of #1733?

PR #2011, opened 2022-07-29
Suppress pure-python SequenceMatcher warning

nanotubing:
Using the slow pure-python module isn't necessarily a problem for our use of this library, and this is a very noisy warning. We're suppressing this warning as we consider the implications of switching to a faster module in #2010 

## Notes

This PR suppresses only the warning:
"Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning"

This warning is thrown when thefuzz module is loaded, which is why the call to `warnings.filterwarnings` is so early in the file.

Connects #2009 

## Testing Instructions

Running `./scripts update` in `ogr/develop` will give output that looks like this, displaying the relevant warning a few times:
```
[+] Running 2/2
 ⠿ Container open-apparel-registry-database-1  Healthy                                                         0.6s
 ⠿ Container open-apparel-registry-django-1    Started                                                         0.9s
[+] Running 1/0
 ⠿ Container open-apparel-registry-database-1  Running                                                         0.0s
/usr/local/lib/python3.7/site-packages/thefuzz/fuzz.py:11: UserWarning: Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning
  warnings.warn('Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning')
Operations to perform:
  Apply all migrations: account, admin, api, auth, authtoken, contenttypes, sessions, sites, socialaccount, waffle
Running migrations:
  No migrations to apply.
```

* Check out this branch
* run `./scripts/update`. The output text will no longer display that warning.
```
[+] Running 2/2
 ⠿ Container open-apparel-registry-database-1  Healthy                                                         1.3s
 ⠿ Container open-apparel-registry-django-1    Started                                                         1.6s
[+] Running 1/0
 ⠿ Container open-apparel-registry-database-1  Running                                                         0.0s
Operations to perform:
  Apply all migrations: account, admin, api, auth, authtoken, contenttypes, sessions, sites, socialaccount, waffle
Running migrations:
  No migrations to apply.
```

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/facility_type_processing_type.py
+++ b/src/django/api/facility_type_processing_type.py
@@ -3,8 +3,8 @@ warnings.filterwarnings(
     "ignore",
     message="Using slow pure-python SequenceMatcher. Install python-Levenshte")
 
-from api.helpers import clean # NOQA
-from thefuzz import process # NOQA
+from api.helpers import clean  # NOQA: E402
+from thefuzz import process  # NOQA: E402
 
 HEADQUARTERS = 'headquarters'
 NO_PROCESSING = 'no processing'
--- a/src/django/api/facility_type_processing_type.py
+++ b/src/django/api/facility_type_processing_type.py
@@ -1,8 +1,10 @@
 import warnings
-warnings.filterwarnings("ignore", message="Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning")
+warnings.filterwarnings(
+    "ignore",
+    message="Using slow pure-python SequenceMatcher. Install python-Levenshte")
 
-from api.helpers import clean
-from thefuzz import process
+from api.helpers import clean # NOQA
+from thefuzz import process # NOQA
 
 HEADQUARTERS = 'headquarters'
 NO_PROCESSING = 'no processing'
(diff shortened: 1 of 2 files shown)


jwalgran:
I suspect you could appease the linter by moving the call to `filterwarnings` below the rest of the import statements.

nanotubing:
> I suspect you could appease the linter by moving the call to `filterwarnings` below the rest of the import statements.

I tried that first, but the warning filter wasn't taking affect until I put it at the top of the file, I think because the warning is thrown when thefuzz is imported.

src/django/api/facility_type_processing_type.py
```diff
@@ -1,5 +1,10 @@
-from api.helpers import clean
-from thefuzz import process
+import warnings
+warnings.filterwarnings(
+    "ignore",
+    message="Using slow pure-python SequenceMatcher. Install python-Levenshte")
+
+from api.helpers import clean # NOQA
+from thefuzz import process # NOQA
```
maurizi, line 7: We can specify which lint rules specifically we want to opt out of, so we don't have to opt-out of lints we want to keep (e.g. "unused import")

```suggestion
from api.helpers import clean  # NOQA: E402
from thefuzz import process  # NOQA: E402
```

maurizi (approved):
This does suppress the warnings 👍 

I added a code suggestion but its pretty trivial, feel free to merge after addressing it w/o another review cycle.

Issue #1945, opened 2022-06-29
Add transfer status to the Adjust Facility Matches interface


obrienad:
- Add whether the facility has been transferred from another facility to the Adjust Facility Matches dashboard page;
  - This only needs to show the immediately prior facility it was transferred from - i.e. do not show history of more than one prior facility.

PR #2014, opened 2022-08-03
Add transfer status to Adjust Facility Matches

jacobscarfmerrell:
## Overview

Add transfer status to Adjust Facility Matches

Connects #1945

## Demo

## Notes

## Testing Instructions

1. Get an authentication token (c2 has one already under settings>tokens) and go to http://localhost:8081/api/docs/#!/facilities/facilities_create and authenticate there
2. Get the name, address, country, and sector for a facility and submit those as the name, address, country and sector for a new facility (under facilities>post)
3. As c1, go to [the Adjust Facility Matches page in the dashboard](http://localhost:6543/dashboard/adjustfacilitymatches)
5. Search using the OAR ID of the match you've just submitted; the new match should not display a "Transferred from" field
6. Transfer the new match to another OAR ID (e.g. US2022214B3QZNE)
7. Search again using the OAR ID of the match you just transferred; the new transfer should display a "Transferred from" field with a value of the prior OAR ID

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/components/DashboardAdjustMatchCard.jsx
+++ b/src/app/src/components/DashboardAdjustMatchCard.jsx
@@ -13,6 +13,7 @@ import get from 'lodash/get';
 import merge from 'lodash/merge';
 import find from 'lodash/find';
 import capitalize from 'lodash/capitalize';
+import isNil from 'lodash/isNil';
 import { Link } from 'react-router-dom';
 import { toast } from 'react-toastify';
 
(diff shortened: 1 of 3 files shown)


src/app/src/components/DashboardAdjustMatchCard.jsx
```diff
@@ -392,6 +392,12 @@ export default function DashboardAdjustMatchCard({
                                 label="Confidence Score"
                                 value={match.confidence}
                             />
+                            {typeof match.transferred_from === 'string' && (
```
jwalgran, line 395: We don't use `typeof` elsewhere in the app and JavaScript types can be tricky. Since we are looking for a non-null value here, consider

```javascript
import isNil from 'lodash/isNil';
```

and 

```javascript
                            {!isNil(match.transferred_from) && (
```

jwalgran (approved):
I verified the tests and this looks good. One optional suggestion.

Issue #2015, opened 2022-08-03
Add status fields to `FacilityList` model


jwalgran:
It is likely that a 2 stage migration will be required change the default value of `status` after

These fields are similar to ones that appear on `FacilityClaim`

Add the following fields to `FacilityList` with appropriate `help_text`

- `status` choice field (not null, not blank)
  - Values: `PENDING`, `APPROVED`, `REJECTED` 
  - Set the initial default value to `APPROVED` so that all existing lists predating the approval feature are considered approved
  - Add a migration step to change the default `status` to `PENDING`
-  `status_change_reason` (not null, allow blank, default empty string)
- `status_change_by` (nullable FK to `User`)
  - ` on_delete=models.PROTECT`

PR #2025, opened 2022-08-05
Add status fields to FacilityList model (#2015)

jacobscarfmerrell:
## Overview

Add status fields to FacilityList model

Connects #2015

## Notes

There are two migrations. In the first, `status` defaults to APPROVED,
so that facility lists predating these migrations won't need to have
their statuses approved or rejected. The second migration then changes
the `status` default to PENDING for future records.

## Checklist

- [x] CI passes after rebase


TaiWilkin (changes requested):
This is working well. The APPROVED -> PENDING default swap is a clever way to handle the status setup. 
I recommend cleaning up the migration order/names a bit - you should be able to avoid a merge migration by renaming the new migrations to `0101_` and `0102_`, then updating their `dependencies` arrays to rely on `0100_` and `0101_` respectively. I also suggest updating the `auto_` name to something descriptive.

TaiWilkin (approved):
Looks good!

Issue #2018, opened 2022-08-03
Allow moderation team members to approve or reject pending facility lists and trigger processing


jwalgran:
Depends on #2016
Depends on #2017 

- Add `approve` and `reject` methods to the `FacilityListViewSet`. 
  - Update the `status`, `status_change_reason`, and `status_change_by` fields.
- Connect list approve and reject UI elements to actions that make POST requests to the new API endpoints
- If a list is rejected, send a rejection email to the contributor incorporating the `status_change_reason` text (reference `send_claim_facility_denial_email`)
- If the the list is approved, call the function created in #2016 to submit the remaining AWS Batch jobs

Issue #2016, opened 2022-08-03
Split the Batch processing for updated lists to incorporate moderation approval


jwalgran:
To date the existing code for handling an uploaded list schedules a sequence of 4 AWS Batch jobs to fully process the list.  We are introducing moderator approval of lists before they are fully processed, requiring us to split the job submission process.

NOTE Batch does not run in development so this will need to be tested in staging

- Refactor `api.aws_batch.submit_jobs` into 2 functions, one that just submits the `parse` job and a second that submits the `geocode`, `match`, and `notify_complete` jobs.

PR #2028, opened 2022-08-09
Split batch steps

TaiWilkin:
## Overview

To date the existing code for handling an uploaded list schedules a
sequence of four AWS Batch jobs to fully process the list. We are
introducing moderator approval of lists before they are fully
processed, requiring us to split the job submission process.

The 'parse' step is completed in the first function (`submit_parse_job`), and the 
'geocode', 'match' and 'notify' steps occur in the second function (`submit_jobs`). 
Currently, both functions are being called in the list upload view,
with the parse job id being provided to the `submit_jobs` function 
so that the other jobs don't begin until after parsing. 
In #2018, the `submit_jobs` function will be removed from this view, 
and instead triggered manually. 

Connects #2016 

## Demo

Screen Shot 2022-08-09 at 11 50 44 AM
Screen Shot 2022-08-09 at 11 51 15 AM

## Testing Instructions

* This branch is currently live on OSHub staging. Upload a list, and confirm that it processes as expected. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/aws_batch.py
+++ b/src/django/api/aws_batch.py
@@ -44,98 +44,88 @@ def fetch_latest_active_batch_job_definition_arn(client):
             'No job queues available. Response {0}'.format(response))
 
 
-def submit_jobs(facility_list, skip_parse=False):
-    """
-    Submit AWS Batch jobs to process each FacilityListItem in a FacilityList
-    through all processing steps.
-    https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/batch.html#Batch.Client.submit_job
-    """
-    client = boto3.client('batch')
-    queue_arn = fetch_batch_queue_arn(client)
-    job_def_arn = fetch_latest_active_batch_job_definition_arn(client)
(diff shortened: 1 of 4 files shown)


src/django/api/aws_batch.py
```diff
@@ -176,6 +175,58 @@ def append_processing_result(result_dict):
...
+    }
+
+    if job_ids is None:
+        job_ids = []
+
+    geocode(job_data, facility_list, job_ids)
+
+    match(job_data, facility_list, job_ids)
+
+    notify(job_data, facility_list, job_ids)
```
jwalgran, line 230: Consider adding a comment that makes it explicit that these calling these functions mutate the `job_ids` parameter.

jwalgran (approved):
I confirmed this worked in staging and verified that the job dependencies were set correctly.

src/django/api/aws_batch.py
TaiWilkin: Good call - updated!

Issue #2017, opened 2022-08-03
Update /dashboard/lists and /lists/{id} pages to incorporate list status


jwalgran:
### /dashboard/lists page

- Add a dropdown to change the `sattus` value (`PENDING`, `APPROVED`, `REJECTED`, `MATCHED`)
  - **NOTE** the `status` field is also uses as part of #1979.

**NOTE**:  The tasks below are also part of #1979. Make sure to coordinate if this is being developed in parallel to avoid duplicating work.

- Add pagination to /api/facility-lists
- Add a `status` query param to the lists API
- Add pagination UI to the /dashboard/lists page

### /lists/{id} page
- Add the `status` and `status_change_reason` fields to the FacilityListSerializer
- On the `/lists/{id}` page
  - Show the list status at the top of the page
  - If the status is `REJECTED` , show a "More Information" link that, when clicked, shows the `status_change_reason` in a dialog.
- If the logged in user is a superuser
  -  If the list status is `PENDING`, show "APPROVE" and "REJECT" buttons at the top of the page 
  - If "REJECT" is clicked shoe a dialog with a text box to collect a `status_change_reason` value 
    - Reference the `Dialog` used in `DashboardClaimDetailsControls.jsx`

Issue #1979, opened 2022-07-20
Move confirm/reject to a backend workflow and introduce permissions for ability to confirm / reject specific lists


obrienad:
## Overview

Move confirm/reject to a backend workflow, with the ability to give specific users access to confirm/reject specific lists. This will improve data quality as the OAR team is better positioned to understand when a facility meets the criteria for being matched. The requirement includes that there must be a confirm/reject queue populated with pending facility items. This does not include flagged queue).

## Technical Implementation

- Add non-null `match_responsibility` field to both `Contributor` and `FacilityList` models.
  - Choice field: `MODERATOR` or `CONTRIBUTOR`
  - The default for both `Contributor` and `FacilityList` should be `MODERATOR` **BUT** any existing lists at the time of migration should have a default value.
    - A suggested way of handling this is to have a 2-stage migration with the the default faclut of `FacilityList.match_responsibility` first set to `CONTRIBUTOR` so that all the existing list values are set appropriately, then modified to be `MODERATOR`.
- Ensure that this `match_responsibility` field can be edited for both `Contributor` and `FailityList` via the Django admin  
- When a `FacilityList` is created, copy the value of `match_responsibility` from the `Contributor` to the `FacilityList`
- When browsing the list details page, show the confirm/reject buttons based on the user and the value of `match_responsibility`. 
  - If the contributor is  viewing the list, show the buttons if `match_responsibility` is `CONTRIBUTOR`
  - If any superuser is logged in, show the buttons if `match_responsibility` is `MODERATOR`
    - Reference how dashboard access is managed in the client to see how super admins are recognized 
 - When browsing the list details page conditionally show an informational messaged value of `match_responsibility`. 
   -  If the contributor is  viewing the list but the `match_responsibility` is `MODERATOR` show a message that the list is awaiting moderation
     - **QUESTION** Get exact copy from the OS Hub team 
- Add pagination to `/api/facility-lists`
- Add a `match_responsibility` and `status` filter parameters to `/api/facility-lists`
  - `status` supports a single value, `MATCHED` and only returns lists where the items are in any of the match status 
- Ensure the sort order of `/api/facility-lists` is oldest list first
- Add pagination UI to the /dashboard/lists page
- Modify the default content of the /dashboard/lists page
  - Fetch lists with `match_responsibility=MODERATOR&status=MATCHED`
- Add dropdown to change the `match_responsibility` value

jwalgran:
[...]

PR #2030, opened 2022-08-10
Have admins confirm/reject potential matches

maurizi:
## Overview

Moves confirm/rejection of potential facility matches for most lists from being done by contributors to being done by OS Hub admins. Select contributors and/or lists can be changed in the Django admin to be moderated by contributors instead, as will any existing lists.

Also updates the admin list dashboard to add a new option to display all contributors (shown by default), as well as a new control for filtering by who is responsible for match moderation and pagination controls.

Connects #1979

## Demo

image

image

image

## Notes

This adds a new default `status` filter of `MATCHED` to the Dashboard - we'll restore the ability to see lists in other status' in https://github.com/open-apparel-registry/open-apparel-registry/issues/2017

## Testing Instructions

* On `develop`
* As a regular contributor, add a new list that will cause some items to be potential matches that need to be confirmed / rejected (I uploaded the same list twice w/ different names)
* On this branch run `scripts/manage migrate`
* You should be able to continue to moderate the lists you already uploaded
* If you upload another list, you should instead see a message that it is awaiting moderation
* Log out and log in as a super user (or log in using a private window). On the Dashboard lists page you should see the list you just uploaded, and be able to confirm/reject the potential match.
 * If you switch the match responsibility filter you should see the first w/ potential matches list you uploaded at the top of the list

## Checklist

- [x] CI passes after rebase


src/app/src/components/FacilityListItems.jsx
```diff
@@ -146,6 +149,13 @@ class FacilityListItems extends Component {
                 </p>
             ) : null;
 
+        const awaitingModerationMessage = readOnly ? (
+            <p style={{ color: 'red' }}>
+                This list has matches awaiting moderation from the{' '}
+                {isAdminUser ? 'contributor' : 'OS Hub admins'}.
```
maurizi, line 155: @obrienad this is the new copy we wanted to run by OAR - I took a stab at it (you see this in the 3rd screenshot on the PR description as well).

src/app/src/components/FacilityListItemsConfirmationTableRow.jsx
```diff
@@ -110,7 +110,9 @@ function FacilityListItemsConfirmationTableRow({
                     variant="head"
                     style={listTableCellStyles.headerCellStyles}
                 >
-                    <b>Actions</b>
+                    <ShowOnly when={!readOnly}>
+                        <b>Actions</b>
+                    </ShowOnly>
```
maurizi, line 115: This is a change from how we displayed the read-only page before for admins, but it seemed a little weird to me to have an empty "Actions" header when we hide the actions themselves

maurizi:
@jwalgran I assigned you as the primary reviewer because I think you have the most context on what is expected here, but @TaiWilkin was thinking of doing a follow-on task for this so I thought it might be helpful for them to just take a look at the code changes.

jwalgran:
Great plan on the double review assignment. Thanks.

src/app/src/components/FacilityListItems.jsx
```diff
@@ -162,7 +172,7 @@ class FacilityListItems extends Component {
             </Button>
         );
 
-        const backRoute = readOnly ? dashboardListsRoute : listsRoute;
+        const backRoute = isAdminUser ? dashboardListsRoute : listsRoute;
```
jwalgran, line 175: When an admin is returning to the list view using the in-app "Back to lists" button the selected contributor is being preserved but the new responsibility selection is not

jwalgran (approved):
Clear implementation. I did not find any issue in the core functionality and only a minor issue with admins losing filter selections when using the back button.

TaiWilkin (approved):
This looks good to me. Thanks for tagging me in to take a look. 
I was able to replicate the list navigation issue Justin found, but found no other issues.

PR #2054, opened 2022-08-16
Upgrade terraform to 1.1.9

nanotubing:
## Overview
A terraform upgrade was necessary for a desired feature surfaced from work in another PR in the parent issue. 

Connects #1998 

## Notes
This was a multi-step upgrade that involved jumping through multiple versions to minimize the amount of code changes necessary at each step. Due to the nature of the upgrade, the terraform code was applied at each step, so code covered by this PR is currently in place on the OSH staging environment.

## Testing Instructions
This code is currently deployed to staging, so testing should be done as a verification of the changes.

* ensure that jenkins completed successfully
* review the console output in jenkins to verify that nothing was changed that affects the functioning of the site
* navigate the staging site to ensure correct functionality
* submit a list to ensure that submitted lists are processing correctly.

## Checklist

- [ ] `fixup!` commits have been squashed
- [x] CI passes after rebase
- [ ] CHANGELOG.md updated with summary of features or fixes, following [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) guidelines
- [x] This PR is targeted at the correct branch (`develop` vs. `ogr/develop`)
- [ ] If this PR applies to both OAR and OGR a companion PR has been created

Issue #1998, opened 2022-07-27
Provision new AWS resources for API notification delivery


jwalgran:
- Provision an AWS Batch queue for notifications
  - Use list processing Batch jobs as a reference
  - Send batch job failure notifications using the same method as list processing failure notifications
- Provision a new AWS Batch compute environment
  - Notification delivery will require modest resources. Configure compute to have access to as many instance types as are appropriate to launch a Django container that just makes db querys and POSTs  to URLs
- ~Provision a CloudWatch log stream for logging sent notifications~ (details on this removal in comments below)
  - ~The "CLI" ECS tasks will need write permissions to the log stream~

jwalgran:
On 2022-08-05 @maurizi  and I chatted about CloudWatch in the context of #2000.  The content CloudWatch of log groups can be filtered and parsed in powerful ways so we can use our existing log group and just filter it to get information about notifications

<img width="960" alt="Screen Shot 2022-08-05 at 1 12 05 PM" src="https://user-images.githubusercontent.com/17363/183158633-35374331-8899-483e-9c3b-de8bc13fbf2f.png">

<img width="613" alt="Screen Shot 2022-08-05 at 1 16 22 PM" src="https://user-images.githubusercontent.com/17363/183158664-06dbfe53-67dc-43c9-b9f4-c3b1107616dc.png">

jwalgran:
While testing my new Batch configuration in staging I learned that logs for Batch jobs end up in the general `/aws/batch/job` log group. 

According to [an issue in the terraform-aws-provider repo](https://github.com/hashicorp/terraform-provider-aws/issues/15560#issuecomment-710077627) changing the log stream for a job requires version 3.1 of the Terraform AWS provider. OS Hub currently used 2.7 (Terraform 0.11)

I found that through Log Insights it is possible to filter on a partial log stream name and the log streams includes the job name.

```
fields @timestamp, @message
| filter @logStream like /jobOpenSupplyHubStagingNotifications/
| sort @timestamp desc
| limit 20
```

This leads me to believe that don't strictly have to upgrade our Terraform and create dedicated log streams to view filtered notification log events, but this log group has a retention set to 1 month, and I anticipate wanting to see notification logs going back further than 1 month. It is easy to change the retention through the console, but:
- This will presumably change the retention for all batch job logs across all deployments.
- I do not know how this log group was created initially so I am concerned about manual edit being overwritten.

[...]

PR #2058, opened 2022-08-17
Fix a terraform deploy bug caused by git security settings

nanotubing:
## Overview

Due to a recent security change implemented by git, we are receiving the following error while the
`infra/plan` is running:
`fatal: unsafe repository ('/usr/local/src' is owned by someone else)`

We're adding a git command to the `infra` script, to set the working
directory as a safe directory, which should resolve this error

Connects #1998
This resolves a bug caused after merging #2054 

## Testing Instructions
* if the `infra` step in jenkins passes on merge, this PR has been successful

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/scripts/infra
+++ b/scripts/infra
@@ -32,6 +32,7 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
 
     if [[ -n "${OAR_SETTINGS_BUCKET}" ]]; then
         pushd "${TERRAFORM_DIR}"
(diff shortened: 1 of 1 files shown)


jwalgran (approved):
This change looks reasonable. I will defer to the comment from @BryanQuigley that we may need to explicitly enumerate directories.

jwalgran (approved):
This looks resonable

PR #2067, opened 2022-08-18
replace thefuzz with rapidfuzz

### LINKED ISSUE #2010: Decide on a path forward for thefuzz dependency

maxbachmann:
This replaces `thefuzz` with `rapidfuzz` to resolve #2010. This is probably the simplest way to resolve the issue, since rapidfuzz provides a compatible API.
This is faster, does not print the warning and is MIT licensed which should be a better fit for this project.

### DIFF
--- a/src/django/api/facility_type_processing_type.py
+++ b/src/django/api/facility_type_processing_type.py
@@ -445,14 +445,15 @@ def get_facility_and_processing_type(facility_or_processing_type):
 
         # Try for fuzzy match
         if not matched_value or matched_value is None:
+            # Match must score 85 or higher to be considered usable.
             matched_value = process.extractOne(
                 cleaned_input,
-                ALL_PROCESSING_TYPES.keys()
+                ALL_PROCESSING_TYPES.keys(),
+                score_cutoff=85
             )
             match_type = FUZZY_MATCH
 
-            # Match must score 85 or higher to be considered usable.
-            if not matched_value or matched_value[1] < 85:
+            if not matched_value:
                 return (None, None, None, None)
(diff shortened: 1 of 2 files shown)


jwalgran:
Thanks for your contribution. I will take a look at this in our upcoming development sprint.

maxbachmann:
@jwalgran could this get a review.

jwalgran:
@KlausGPaul notifying you about this PR that seeks to address #2010. I removed myself that the assigned reviewer.

CC @mariel-oar

maxbachmann:
Since this did not get a review after 8 months I am closing the PR. Feel free to apply the change directly yourself.

Issue #2034, opened 2022-08-10
Add model version checking to `OgrGazetteerMatching` and modify insert conflict resolution


jwalgran:
Depends on #2033

- Add `__init__` to `OgrStaticStaticGazetteer`
  - Set `self.trained_model` to the active `TrainedModel
  - Call superclass passing`self.trained_model.dedupe_model`
- Add `trained_model` checks to top of the methods in `OgrGazetteerMatching`
  - Compare `TrainedModel.objects.get_active_version_id()`  to the `self.trained_model.id` instance variable value. If they are a mismatch, raise a custom `ModelOutOfDate` exception
- Replace `ON CONFLICT...DO NOTHING` with `ON CONFLICT...DO UPDATE...`
  - If we index an existing record we want to update the `record_data`

Before
```
ON CONFLICT (block_key, record_id) DO NOTHING
```
After
```
ON CONFLICT (block_key, record_id)
DO UPDATE SET record_data = EXCLUDED.record_data
```

jwalgran:
After a call with Forest on 2022-08-22 we revised our implementation

### Before

- Add `mocel_version_id` instance variable to `OgrGazetteerMatching`
- Add `model_version_id` checks to top of the methods in `OgrGazetteerMatching`
	- Compare `TrainedModel.objects.get_active_version_id()`  to the `model_version_id` instance variable value. If they are a mismatch, raise a custom `ModelOutOfDate` exception

### After

- Add `__init__` to `OgrStaticStaticGazetteer`
  - Set `self.trained_model` to the active `TrainedModel
  - Call superclass passing`self.trained_model.dedupe_model`
- Add `trained_model` checks to top of the methods in `OgrGazetteerMatching`
	- Compare `TrainedModel.objects.get_active_version_id()`  to the `self.trained_model.id` instance variable value. If they are a mismatch, raise a custom `ModelOutOfDate` exception

Issue #2033, opened 2022-08-10
Build a new branch for the new dedupe version and architecture amd make changes to `TrainedModel`


jwalgran:
- Build a new branch starting from the dedupe 2.x upgrade branch and rebased onto the latest `ogr/develop`
	- Copy over content from the `feature/jcw/dedupe-2` branch without the old WIP cruft
		- Add the latest `gazetteer.py` , `train_model.py` and the `TrainedModel` class in `models.py`
		- Rename `creation_time` to `created_at` for consistency with the other models in the app.
		- Make a new migration fro `TrainedModel` AFTER adding `is_active`
- Add `is_active` boolean to `TrainedModel`.  Default false.
- Add a custom default [manager](https://docs.djangoproject.com/en/4.1/topics/db/managers/#adding-extra-manager-methods) to `TrainedModel` so we can call `TrainedModel.objects.get_active()`. Should return the single instance where `is_active` is true or throw. 
	- When loading a model from the database use the `get_active()` manager method instead of getting latest
	- Add `get_active_version_id()` to return the single value.

PR #2083, opened 2022-08-23
Build a new branch for the new dedupe version and architecture and make changes to TrainedModel

nanotubing:
## Overview
For 2033:
We built a clean branch based off a recent copy of `ogr/develop`, applying commits from the dedupe version upgrade branch
* We made changes to the fields in TrainedModel, and added a constraint that only one model be active
* We've also created a custom default manager attached to that model, and added a migration for the changes. 

For 2034:
* We subclass the GazetteerMatching, StaticGazetteer, and Gazetteer classes, applying changes from the research branch necessary for our implementation of dedupe.
* We update the model training architecture to train a model and save it to the TrainedModel object in the database.
* the index, unindex, and blocks methods check that they're running the current active model, and raise an exception if not.

Connects #2033 
Connects #2034

### 2033
* check out this branch
* run `./scripts/manage train_model` twice to train two new models
* exit that and run `./scripts/dbshell`
* run the query `select id, created_at, is_active from api_trainedmodel;`. you should get two rows output with your two models
* make the first model the active model by running `update api_trainedmodel set is_active=true where id=1;`
* try to make the second model the active model by running `update api_trainedmodel set is_active=true where id=2;`. This command should fail because of the constraint that only one model be active
* exit from the SQL shell and run `./scripts/manage shell_plus`
* run `TrainedModel.objects.get_active()`, which should return the object of the currently active model
* run `TrainedModel.objects.get_active_version_id()`, which should return the ID of the currently active model
## 2034
* run `./scripts/update`
* run `./scripts/server`. If everything works as expected, you will get the error seen below. This happens because a model is trained on startup, but it is not being marked as the active model
```
django_1    |   File "/usr/local/src/api/gazetteer.py", line 31, in index
django_1    |     self.check_model_version()
django_1    |   File "/usr/local/src/api/gazetteer.py", line 17, in check_model_version
django_1    |     raise ModelOutOfDate()
django_1    | api.gazetteer.ModelOutOfDate
```

## Checklist

- [ ] CI passes after rebase


src/django/api/models.py
```diff
@@ -2934,22 +2934,27 @@ def index_facilities(facility_ids=list):
...
+                       fields=['is_active'],
+                       condition=Q(is_active=True))]
     dedupe_model = models.BinaryField()
     created_at = models.DateTimeField(auto_now_add=True)
     is_active = models.BooleanField(
         null=False,
         default=False,
-        help_text=('True if this is the currently active version of the dedupe model')
+        help_text=('True if this is the currently active version of the dedupe \
+        model')
```
jwalgran, line 2957: Using this method of splitting a string over multiple lises should avoid the need for a migration.
```suggestion
        help_text=('True if this is the currently active version of the dedupe '
                   'model')
```

src/django/api/migrations/0107_alter_trainedmodel_is_active.py
```diff
@@ -0,0 +1,18 @@
...
+        ('api', '0106_trainedmodel_unique_is_active'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='trainedmodel',
+            name='is_active',
+            field=models.BooleanField(default=False, help_text='True if this is the currently active version of the dedupe         model'),
+        ),
+    ]
```
jwalgran, line 18: Changing the way the help_text is broken across 2 lines should allow us to delete this migration (see comment in the models file)

src/django/api/matching.py
```diff
@@ -17,8 +18,10 @@
                         FacilityListItem,
                         FacilityMatch,
                         HistoricalFacility,
-                        HistoricalFacilityMatch)
+                        HistoricalFacilityMatch,
+                    TrainedModel)
```
jwalgran: Consider indenting another level here to make this  consistent.

jwalgran (commented):
I added a few inline comments.

I expect using `setUpTestData` to allow us to fix the broken tests by training and activating a model for the test classes that need  it.

jwalgran (commented):
Looks good. I verified the tests. This PR is going to be our "target" for other PRs and we will only want to merge it once the feature is complete.

jwalgran:
Superseded by #2203

Issue #2105, opened 2022-08-29
Convert dedupe_indexed_records to a model


nanotubing:
The `dedupe_indexed_records` table is currently a vanilla postgres table, being accessed directly through psychopg2. To make it easier to manipulate this table using our existing django tools, this table should be converted into a model with a corresponding migration.

PR #2107, opened 2022-08-29
Convert dedupe_indexed_records to a model

nanotubing:
## Overview

`dedupe_indexed_records` was previously a standard SQL table, but we're moving it to a django
model, so we can take advantage of django features such as dumpdata and loaddata:
* This PR removes the code that was previously creating the table and index
* the name of the table has changed slightly, from `dedupe_indexed_records` to `api_dedupe_indexed_records`
* the loop that uses SQL to populate the index table is not changing, because there is no easy way to replicate `ON CONFLICT` doing it the django way

Connects #2105

## Testing Instructions

* check out this branch
* run `./scripts/update`. This will run the migration that creates the `api_dedupe_indexed_records` table
* run `./manage/dbshell`
* run `\d api_dedupe_indexed_records` and verify that the table, constraints, and index were created correctly

## Checklist

- [ ] CI passes after rebase


src/django/api/models.py
```diff
@@ -2959,6 +2959,16 @@ class Meta():
...
 
+class dedupe_indexed_records(models.Model):
+    class Meta:
+        indexes = [
+            models.Index(fields=['block_key', 'record_id'],
+                         name='dedupe_indexed_records_idx')
+        ]
+    block_key = models.TextField(unique=True)
+    record_id = models.CharField(unique=True,
+                                 max_length=32)
```
jwalgran, line 2970: `block_key` and `record_id` are not unique. The `indexes` option Meta option will handle the unique pair.

```suggestion
    block_key = models.TextField()
    record_id = models.CharField(max_length=32)
```

This change will require deleting and recreating the migration with makemigrations.

jwalgran (changes requested):

src/django/api/models.py
nanotubing: I had added a UniqueConstraint meta option before I saw your comment. Is there any reason to leave this along with the index, or is it not necessary to have both?
```
constraints = [UniqueConstraint(fields=['block_key',
                                                'record_id'],
                                        name='unique_blockkey_recordid')]
```

jwalgran: Good catch. We do need to have this unique constraint.

jwalgran (approved):

src/django/api/gazetteer.py
```diff
@@ -198,9 +178,9 @@ def build_index_table(self, canonical):
         table_name = 'dedupe_indexed_records_{}'.format(self.trained_model.id)
 
         with connection.cursor() as cursor:
-            self._create_index_table(cursor)
             cursor.execute(
-                "CREATE TABLE {} (LIKE dedupe_indexed_records INCLUDING ALL)"
+                "CREATE TABLE {} (LIKE dedupe_indexed_records INCLUDING "
+                                  "ALL)"
                 .format(table_name)
```
jwalgran, line 184: The linter is unhappy about this
```
./api/gazetteer.py:183:35: E131 continuation line unaligned for hanging indent
```

 This works 

```suggestion
                "CREATE TABLE {} (LIKE dedupe_indexed_records INCLUDING ALL)"
                .format(table_name)
```

jwalgran (approved):
Looks good. There is just a bid of lint to remove

Issue #2082, opened 2022-08-22
Make default list status filter "pending" on /dashboard/lists


obrienad:
## Overview

The current default is to show lists in the `MATCHED` status. Change this default to `PENDING`.

PR #2109, opened 2022-08-30
Use 'pending' as default status filter on dashboard lists

moradology:
## Overview

This PR sets the default filter for /dashboard/lists to 'pending'.

Connects #2082

## Demo

Note the URL's lack of a `status` query parameter:
image

## Notes

Looks like perhaps the wrong enum was used when setting the default; that should be rectified here.

## Testing Instructions

Navigate to the dashboard's list filter page and verify that `pending` is the default filter. Check the network tab to be sure that the appropriate query parameter appears (`status=PENDING`)

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Use 'pending' as the default status filter on the dashboard list page [#2109](https://github.com/open-apparel-registry/open-apparel-registry/pull/2109)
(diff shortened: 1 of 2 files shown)


jwalgran:
The build failure looks like a transient npm issue. I triggered a new build.

jwalgran (approved):
Tested. Looks good. 

I unchecked the CHANGELOG checklist item. Our standard practice on this project is to update the CHANGELOG.md file as part of the PR after the PR is approved. Since this is a single-commit PR, you can just amend the commit after adding a new line to CHANGELOG.md, then force push (for a multi-commit PR we would add an additional commit titled "Update CHANGELOG" with the change)

Issue #2049, opened 2022-08-15
Search/Filtering/Results UX Updates


mariel-oar:
## Overview

This ticket captures requirements for updating the UX for search and filtering functionality as well as the display of search results. 

### Included Changes

Must-have: 
1. Updated font, logo and color changes (via #2059 )
2. An updated 3 panel design that displays the relevant search filters along with the results and map (column 1 = filter/search, panel 2 = results, panel 3 map view)
3. Facility results displays: name, address, OS ID, and whether the facility is claimed 
4. Integration of new claimed icon 
5. Filters with fixed values (e.g. number of workers) have drops downs for values; filters with non-fixed values work as they do in OAR 
6. User can further refine search results by interacting with the filters in panel 1 

Nice to haves captured in #2063 

### Visual Design 
Screen design - WIP

mariel-oar:
@mariel-oar to clarify item 6 in must have with Adam

PR #2110, opened 2022-08-30
Search/Filtering/Results UX Updates (#2049)

jacobscarfmerrell:
## Overview

The facilities search results UX has changed significantly. Now, the search and results belong in one sidebar for large windows. For mobile screen sizes, we have tabs for the facility results list and map, and the filters are moved into a modal.

Connects #2049 

## Demo

## Notes

Because the two Map components get created and destroyed in the course of resizing the window, leaflet is frequently throwing an irritating error that should only be visible in development.

## Testing Instructions

* How to test this PR
* Prefer bulleted description
* Start after checking out this branch
* Include any setup required, such as bundling scripts, restarting services, etc.
* Include test case, and expected output

## Checklist

- [ ] Review this layout on a screenshare with the OS Hub Team before merging
- [x] CI passes after rebase


jwalgran:
The build failure was caused by lint

```
./api/serializers.py:612:80: E501 line too long (81 > 79 characters)
```

TaiWilkin (changes requested):
This is a great framework for the new layout. Thanks for working through this!
Let's hold off on trying to fix the map error when resizing, based on time constraints and since it doesn't seem to impact functionality. Please make a follow-up ticket for fixing the error when time allowed, though. 
Some styling adjustment requests:
- Can you update the top-of-list buttons (download, copy, filter, etc) to have the updated style, and add some space between the tabs and the filter button? (If updating the tab style turns out to be complicated/time-consuming, please instead make a follow-up card for that.)
- In the same region, can you add the 'Facilities' header and facilities count (count should be available in Redux)? (Don't add the 'sort' dropdown though)
Screen Shot 2022-09-07 at 11 31 44 AM
- Can you make the facility results panel slightly wider (maybe move to 4 columns instead of 3)?
- Can you vertically center the items in the OAR ID/Claimed row so the claim badge is aligned a bit more? 
Screen Shot 2022-09-07 at 10 44 45 AM vs Screen Shot 2022-09-07 at 11 40 25 AM
- Can you update the multiple-selected-contributors style to the new toggle with the 'i' next to the label? (OR if this turns out to be time-consuming, just adjusting the styling so the 'i' button doesn't overflow to a new line)
Screen Shot 2022-09-07 at 11 41 39 AM vs 
Screen Shot 2022-09-07 at 11 41 28 AM
- Can you add a close button and 'Filters' header to the filters drawer? 
Screen Shot 2022-09-07 at 11 44 30 AM

src/app/src/components/MapAndSidebar.jsx
```diff
@@ -1,123 +1,31 @@
...
-                                />
-                            </Switch>
-                        )}
-                        {hasError && <FacilitiesMapErrorMessage />}
-                    </Grid>
-                </Grid>
-            </Fragment>
-        );
-    }
+function mapStateToProps({ ui: { activeFilterSidebarTab } }) {
```
TaiWilkin, line 25: Looks like this prop isn't being used, so this mapStateToProps can probably be removed

src/app/src/components/FilterSidebar.jsx
```diff
@@ -93,85 +79,73 @@ class FilterSidebar extends Component {
...
+                        )}
+                    </Grid>
+                </Hidden>
+                <Hidden mdDown>
+                    <Grid item md={2} style={{ height: '100%' }}>
+                        <FilterSidebarSearchTab />
+                    </Grid>
+                </Hidden>
+                <Hidden mdDown>
+                    <Grid item sm={12} md={3}>
```
TaiWilkin, line 148: Can we increase this to `md={4}`?

jwalgran:
During our weekly check-in with the OS Hub team I showed a preview of this branch and they were concerned about the relative widths of the 3 columns (the search fields section looked a bit narrow). Tai's specific review comments may address this. Thanks.

jacobscarfmerrell:
@TaiWilkin @jwalgran I think all that's left to do is to update the multiple-selected-contributors style to the new toggle with the 'i' next to the label and to try and boost the size of the filter bank on large screens. I can finish these two points tomorrow morning!

TaiWilkin:
@jacobscarfmerrell Great work on the tab and styling! They look great. This is coming along well. 

I'm seeing a couple of small issues still. 

- I think the new styling on the contributors tab caused this - it's now pushed out of alignment on the homepage. 
Screen Shot 2022-09-09 at 1 05 55 PM
- We should include the Facilities header and count (for the results panel) on the desktop / non-mobile view as well
- This one is weird, so I made a gif - when I'm scrolling down the list, the padding changes:
PaddingGlitch
- On the desktop view, the last item on the results list is blocked by the footer

TaiWilkin:
@jwalgran This PR is ready to show to Mariel and team at the demo today.

TaiWilkin (approved):
Thank you for working through this complex issue!

Issue #2074, opened 2022-08-19
Create a dedicated page for facility details


jwalgran:
Previously facility details have been shown in the sidebar on the main page. We now want the /facilities/{os-hub-id} route to be a dedicated page.

This issue covers creating an empty placeholder facility details page component and updating the routing so that navigating to /facilities/{os-hub-id}?some-search=foo renders the new details page and navigating back returns to /facilities?some-search=foo

PR #2115, opened 2022-08-31
Add facility details page

TaiWilkin:
## Overview

Pulls the facility details view into a separate page, only adding the
basic routing functionality and some initial framing.

Connects #2074 

## Demo

Aug-31-2022 16-41-01
Screen Shot 2022-08-31 at 5 02 18 PM

## Notes

The task says to use a blank page for the route. However, I didn't want to merge a site with a 'broken' facility details page into OSHub staging, since it's being used, so I put the existing facility details page into the new page with something of an initial 'frame' as a placeholder.  

## Testing Instructions

* Run `./scripts/server`
* Search for facilities with at least one filter set. 
* Select a facility and confirm that the facility details load in a new page. 
* View the page in mobile and confirm that mobile styling looks reasonable.
* Use the 'Back to search results' button and confirm that the query string is still present on the search page. 
* Login as c3@example.com, and navigate to the embed tab in the user settings page. Set a height and width for the map. Confirm that you are able to navigate back and forth between search and the facility details without errors within the embedded map. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/components/FilterSidebarFacilitiesTab.jsx
+++ b/src/app/src/components/FilterSidebarFacilitiesTab.jsx
@@ -1,6 +1,7 @@
 import React, { Fragment, useState } from 'react';
 import { arrayOf, bool, func, number, string } from 'prop-types';
 import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
 import { Link } from 'react-router-dom';
 import CircularProgress from '@material-ui/core/CircularProgress';
 import LinearProgress from '@material-ui/core/LinearProgress';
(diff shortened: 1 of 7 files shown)


src/app/src/components/FilterSidebarFacilitiesTab.jsx
```diff
@@ -280,12 +284,16 @@ function FilterSidebarFacilitiesTab({
...
                                                 pathname: makeFacilityDetailLink(
                                                     oarID,
+                                                    search,
                                                 ),
                                                 state: {
                                                     panMapToFacilityDetails: true,
                                                 },
                                             }}
-                                            href={makeFacilityDetailLink(oarID)}
+                                            href={makeFacilityDetailLink(
```
TaiWilkin, line 293: This ensures we don't lose the query string when navigating to the facility details

jwalgran (approved):
Worked as described. 👍

TaiWilkin:
Thank you for reviewing!

Issue #2092, opened 2022-08-24
Setup cache backend and DRF throttling settings in Django


jwalgran:
Django cache framework: https://docs.djangoproject.com/en/3.2/topics/cache/
DRF throttling: https://www.django-rest-framework.org/api-guide/throttling/

- The climate-change-api project has an ElastiCache setup based on this module
- Reference the climate project for adding a memcache container to the local dev enfironment

PR #2125, opened 2022-09-07
Add user-customizable API throttling

maurizi:
## Overview

Adds a memcached caching backend, and uses it to add rate throttling using the DRF built-in rate throttling, along with a custom throttle class to enable us to configure throttling at a per-user level in the Django admin.

Connects #2092

## Demo

Error response returning after throttling:
localhost_8081_api_facilities_

Admin fields to configure throttling per user:
localhost_8081_admin_api_user_1_change_ (2)

## Notes

This doesn't add Elasticache to our AWS environment, so this will use the default local memory caching on staging for API throttling, until that is implemented.

I wasn't entirely sure what endpoints to throttle or what default rates to use - I'd appreciate feedback here, preferably based on OAR usage statistics if we have them handy, I started with what seemed like conservative defaults and only unthrottled GET endpoints that seemed fairly simplistic - we could probably be more aggressive in what we allow.

## Testing Instructions

* `scripts/update`
* Try to trigger rate limiting
* Set your user's rate limits in the admin and verify they've taken effect

## Checklist

- [x] CI passes after rebase


jwalgran:
I will need to spend some more time reviewing locally but I wanted to reply with at least a partial response to your notes sooner rather than later

The endpoint that we have the most performance struggle with is POSTing to /api/facilities. I think that we need to incorporate into our `allow_request` function a conditional check on path and verb so that `POST /api/facilities/` by default cannot exceed `30/minue` but allows us to increase that POST limit per user. 

I based that 1 POST every 2 seconds limit on the activity level that has caused prod outages in the past and by doing some analysis of our `RequestLog`

We keep track of all of the requests made with an API token in a database table.  I counted them up grouping by user, method, path prefix, and hour.

```
SELECT count(*) AS count, user_id, method, SUBSTRING(path, 1, 16), date_trunc('hour', created_at)
         AS hour
  FROM api_requestlog
 WHERE response_code = 200
 GROUP BY user_id, method, path, hour
 ORDER BY count DESC
```

I saved the results of this query to our project share
https://drive.google.com/file/d/1VdCGxXu8HGiwlJt-DB-pgywEhF6s90-u/view?usp=sharing


We may be able to craft an Athena query to do some similar analysis of the CloudFront logs grouped by IP address instead of user. That would give us a picture of API calls made via the app.

maurizi:
@jwalgran I added a separate throttle / rate field for data uploads, and applied it to _only_ the facilities POST endpoint.

jwalgran (approved):
This implementation is clear and worked as described in my testing. I agree with the choices for APIs which we are leaving as not throttled.

I verified that memcached was being used by changing `-v` to `-vv` in `docker-compose.yml`  and seeing the cache activity in the log when I made API requests

```
memcached_1  | <24 new auto-negotiating client connection
memcached_1  | 24: Client using the ascii protocol
memcached_1  | <24 get :1:throttle_burst_4
memcached_1  | >24 END
memcached_1  | <24 set :1:throttle_burst_4 1 60 24
memcached_1  | >24 STORED
memcached_1  | <24 get :1:throttle_sustained_4
memcached_1  | >24 END
memcached_1  | <24 set :1:throttle_sustained_4 1 86400 24
memcached_1  | >24 STORED
```

I tested the throttles using the Swagger docs and confirmed they work, and that changes to throttle rates made in the database took effect right away.

Screen Shot 2022-09-08 at 12 05 43 PM

jwalgran:
Some minor lint to clean up to get the build passing.

Issue #2038, opened 2022-08-10
Update the implementation of `GazetteerCache`


jwalgran:
Depends on #2037

- Update the implementation of `GazetteerCache`
	- We no longer need to interact with history tables or keep `_facility_version` or `_match_version`
	- The purpose now is to catch `ModelOutOfDate` exceptions
	- Add static `index` and `unindex` and `search`  methods that have the same parameters as the dedupe `Gazetteer` methods
		- For all three methods, if `_gazetteer ` is None, use `TrainedModel.get_active()`  then set `_gazetteer` to an `OgrStaticGazetteer` created from the saved model
			- For `index` and `unindex` , call the respective methods on `_gazetteer` in a try block.  
				- Catch `ModelOutOfDate`  , load the latest saved Gazetteer, then try calling `index` or `unindex` again
			- `seach` calls `blocks` so `blocks` could raise in the same wai as index. Load the latest saved Gazetteer, then try calling `search`

PR #2134, opened 2022-09-12
Rewrite the GazetteerCache to work with database-backed gazetteer

jwalgran:
## Overview
Our new gazetteer subclasses use the database for sharing trained models and index data so the code within the GazetteerCache used to keep the in-memory model in sync with database changes is no longer needed.

Instead, the GazetteerCache now provides class methods that wrap the gazetteer instance methods and handle refreshing the gazetteer only if a new one has been trained and activated.

Updating the Dedupe index used to be handled by processing historical FacilityMatch rows in the GazetteerCache. Now that we have replaced historical record processing with a shared Dedupe index table we need to update that index whenever a FacilityMatch is saved.

We also remove the dedupe_indexed_records.json fixture because our data has new OAR IDs each time we run `resetdb`. The newly added signal handler takes care of updating the dedupe_indexed_records table via the GazetteerCache.

The updated TrainedModel fixture has a new active model trained after this new coded was added.

Connects #2038

## Demo

```
django_1    | [2022-09-12 19:19:15 +0000] [117] [DEBUG] POST /api/facilities/
django_1    | LOADING THE LATEST TRAINED MODEL
django_1    | Predicate set:
django_1    | SimplePredicate: (sameFiveCharStartPredicate, address)
django_1    | SimplePredicate: (firstTwoTokensPredicate, name)
django_1    | SimplePredicate: (twoGramFingerprint, name)
django_1    | INDEXING  880sp:0 US2022255WZQ50H
django_1    | INDEXING  esstte:2 US2022255WZQ50H
```
### Setup

The test instructions use `jq` to make JSON curl responses more readable. You can install this on macOS with `brew install jq`

### Testing

* Run `./scripts/resetdb` and verify that it completes without error
* Run `./scripts/manage train_model` 
* Open a Django shell with `./scripts/manage shell_plus`
* In the Django shell activate the new model with `TrainedModel.objects.latest('id').activate()`
* Submit a facility. Verify that you see log messages about the new model being loaded and the new facility being added to the dedupe index
```
curl -X POST "http://localhost:8081/api/facilities/" -H  "accept: application/json" -H  "Authorization: Token 1d18b962d6f976b0b7e8fcf9fcc39b56cf278051" -H  "Content-Type: application/json" -d "{\"name\":\"Azavea\",\"address\": \"990 Spring Garden St.,Philadelphia PA 19123\",\"country\": \"US\",\"sector\": \"Apparel\"}" | jq
```
* Take the OAR ID from the console message and use the Django shell it to check that facility was added to the dedupe index
```
DedupeIndexedRecords.objects.filter(record_id='{THE-OAR-ID}').count()
```
```
2
```
* Log in as c1@example.com (admin) and browse http://localhost:6543/lists/8
* Confirm the first pending match. Verify that `INDEXING` is written to the console. Note the FacilityMatch ID that is indexed
* Browse http://localhost:8081/admin/api/facilitymatch/{MATCH-ID}/change/, uncheck the Is Active box, and click "Save and continue editing". Verity that `UNINDEXING` is written to the log. Verify that the records were removed using the Django shell. Replace the `record_id` in the line below with the one written to the console
```
DedupeIndexedRecords.objects.filter(record_id='CN2022253QCFCYX_MATCH-301').count()
```
```
0
```

## Checklist

- [x] `fixup!` commits have been squashed. **DROP** commit has been removed
- [ ] CI passes after rebase


src/django/api/gazetteer.py
```diff
@@ -47,6 +47,7 @@ def index(self, data: Data) -> None:  # pragma: no cover
 
         with connection.cursor() as cursor:
             for item in self.fingerprinter(data.items(), target=True):
+                print('INDEXING ', item[0], item[1])
```
jwalgran, line 50: This is a temporary addition for PR testing that will be removed before morging

src/django/api/gazetteer.py
```diff
@@ -80,12 +81,13 @@ def unindex(self, data: Data) -> None:  # pragma: no cover
         # TODO: Delete in batches?
         with connection.cursor() as cursor:
             for k in data.keys():
+                print('UNINDEXING ', k)
```
jwalgran, line 84: This is a temporary addition for PR testing that will be removed before morging

src/django/api/matching.py
```diff
@@ -538,234 +537,46 @@ class GazetteerCache:
...
-                Facility
-                .objects
-                .filter(id__in=changed_facility_ids_qs)
-                .extra(select={'country': 'country_code'})
-                .values('id', 'country', 'name', 'address')
-            }
-
-        return facility_changes, latest_facility_dedupe_records
+    def load_latest_gazetteer(cls):
+        logger.info('LOADING THE LATEST TRAINED MODEL')
```
jwalgran, line 550: This is a temporary addition for PR testing that will be removed before morging

jwalgran:
The lint failing the build is present on the `ogr/feature/cs/dedupe_upgrade_take2` branch and will be fixed separately

nanotubing (approved):
I walked through everything, and aside from it taking me a minute to figure out which part of the string the MATCH_ID is everything was very straightforward. Does exactly what it says on the tin

jwalgran:
Thanks for the review.

jwalgran:
CI failed but it is due to existing test failures on ogr/feature/cs/dedupe_upgrade_take2

Issue #2133, opened 2022-09-12
Navigating from the facility detail page to the search results page should maintain state


jwalgran:
Navigating back to the search results should not run the search and should maintain scroll position.

Here is a screencast of the current interaction in which the search is rerun and state is lost.

![2022-09-12 08 07 38](https://user-images.githubusercontent.com/17363/189689895-7efeb29a-a1cd-4c7b-806e-7d5aa1846f42.gif)

TaiWilkin:
Feedback on #2075 to be applied in this ticket:
Comments from the OAR team:
- can we make the font for the OAR ID itself size 18 instead of 14
- is it possible to pin the "Description" data point as the last data point and allow it to be the width of the 2 columns, instead of wrapping into the one column. E.g. see this facility's description (note that description only shows up for claimed facilities if the owner has entered a description)

Issue #2127, opened 2022-09-09
Map doesn't show up in mobile HomePage


jacobscarfmerrell:
## Overview

In PR #2110 Map doesn't show up in mobile HomePage

### Expected Behavior

When you navigate to the home page (`/` route), you should be able to scroll down and see the map.

### Actual Behavior

The map doesn't show up.

PR #2141, opened 2022-09-14
Fixup recent UX changes

TaiWilkin:
## Overview

- Increases the font size of the OSHub id on the details page to 18px
- Makes the claim description full-width and moves it to the last spot
- Prevents reloading the facility search when navigating to details and back
- Restores location in list when returning to search from details
     - List location is reset to the top when a new search is completed
     - If the user refreshes the page while on the details page, the search will be rerun and the list location will be reset to the top
- Restores the map in mobile view on the homepage

Connects #2133, #2127

## Demo

Sep-14-2022 10-32-57

Screen Shot 2022-09-14 at 10 33 51 AM

## Notes

The scroll-to-location consistently returns the user to approximately the correct location, but sometimes is a little bit off. I've spent some time attempting to tweak it, but I think this is as accurate as we'll get without spending significantly more time on it. 

I based this on Jacob's PR to handle merge conflicts.

## Testing Instructions

* Run `./scripts/server
* View the map on the homepage on mobile
* Run a search
* Scroll the list
* Navigate to a facility, then back using the back button on screen
* Navigate to a facility, then back using the browser back button
* Navigate to a facility with a claim description

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix list URL construction when running as a Batch job [#2066](https://github.com/open-apparel-registry/open-apparel-registry/pull/2066)
 - Fix getLangFromTranslateElement function [#2097](https://github.com/open-apparel-registry/open-apparel-registry/pull/2097)
 - Update the Python and Rollber versions used for lamba functions [#2120](https://github.com/open-apparel-registry/open-apparel-registry/pull/2120)
+- Fixup recent UX changes [#2141](https://github.com/open-apparel-registry/open-apparel-registry/pull/2141)
 
 ### Security
 
--- a/src/app/src/Routes.jsx
+++ b/src/app/src/Routes.jsx
@@ -17,7 +17,6 @@ import LoginForm from './components/LoginForm';
 import UserProfile from './components/UserProfile';
 import Contribute from './components/Contribute';
 import Homepage from './components/Homepage';
-import MapAndSidebar from './components/MapAndSidebar';
 import FacilityLists from './components/FacilityLists';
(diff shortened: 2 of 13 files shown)


jwalgran (commented):
When returning from the details page to the search results, the results lists is scrolled up to the top.

We are storing the scroll position in Redux (under `ui.scrollTop`) and this Codepen example linked to from https://github.com/seatgeek/react-infinite/issues/267#issuecomment-395940059 makes think that we could restore the scroll position.

https://codesandbox.io/s/v1vlz0j293?file=/src/index.js

jwalgran (commented):
Actually, I missed that we already have this block

```
    const [hasScrolled, setHasScrolled] = useState(false);
    const scrollElements = Array.from(
        document.getElementsByClassName('infinite-scroll'),
    );
    if (!hasScrolled && scrollElements.length) {
        setHasScrolled(true);
        // The timeout allows persisted data to be restored before scroll is attemped
        setTimeout(() => {
            scrollElements.forEach(x => {
                x.scrollTo(0, scrollTop);
            });
        }, 100);
    }
````

I do now see it trying to set the scroll position, but it is not returning to the correct spot for me.

jwalgran (approved):
I confirmed all of the changes except for the scroll restoration. I don't think we should block merging the PR but we will need further work on the scroll restoration portion. I observed 2 issues:. 

- The further down the list I scrolled the less accurate the scroll restoration appeared to be.
- Repeated scroll-click-back sequences would sometimes result in the scroll restoration breaking. Returning to the details page would result in the list returning to the top.

TaiWilkin:
I've created a follow-up issue for the scroll restoration: https://github.com/open-apparel-registry/open-apparel-registry/issues/2150

Issue #2124, opened 2022-09-07
Deleted facilities not really deleting


mariel-oar:
We are seeing an issue where facilities that have multiple matches (or having many API calls with create=true and match status = pending on the backend) are not completely deleted from the platform and are instead aliased as a different OAR ID (which is created when the original OAR ID is deleted). A user can continue to search for the original OAR ID via the URL and it will be redirected to the newly created OAR ID's URL. 

As an example if you search for this facility via the url IN20220746KTGGV you will notice that it redirects to this facility's page: IN2022250HGNBNT. This will continue to happen (we think) until all of the entries that are in match status = pending on the backend are exhausted. 

E.g. - 
Deleted facility:  IN20220746KTGGV 
Aliased as: IN2022250HGNBNT 

Related to #2019 

## Technical Notes

The delete endpoint currently uses the following logic:
- Delete the FacilityListItem which the Facility was created from, and its linked FacilityMatches
- Look for other automatically matched and confirmed match items associated with the Facility, and if a high-quality alternate match exists, create a new Facility from that match. Move all of the matches and list items to the new Facility.
- Perform various cleanup tasks
- Delete the original Facility

The goal of this behavior is to retain potential high-quality FacilityListItems when deleting a Facility. However, in cases where a single contributor submits a facility dozens or hundreds of times, this results in moderators needing to delete each of those submissions one at a time. 

We would like to adjust the logic of the delete endpoint to the following:
- Delete the FacilityListItem which the Facility was created from, and its linked FacilityMatches
- **Delete any linked FacilityListItems and FacilityMatches linked to the Facility and submitted by the same contributor as submitted FacilityListItem which the Facility was created from.**
- Look for other automatically matched and confirmed match items associated with the Facility **but not submitted by the `created_from` contributor**, and if a high-quality alternate match exists, create a new Facility from that match. Move all of the matches and list items to the new Facility.
- Perform various cleanup tasks
- Delete the original Facility

mariel-oar:
Discussed in weekly as a 2pt issue based on the implementation listed above (delete any pending facilities that are from the same contributor)

PR #2153, opened 2022-09-18
OSH: Delete matched facilities from created contributer

mstone121:
## Overview

If a facility has more than one match, those other matches could be promoted when the original facility is deleted. This means that a contributer might need to delete a facility many times to fully delete the facility.

In order to simplify the delete process, facility list items from the same contributor as the original facility should be deleted along with the original facility. This includes the matches associated with these facility list items.

Connects #2124

## Notes

This PR has some code with strange looking destructuring with the python double astrisks operator:
```python
                for match in other_matches.filter(**{
                    "facility_list_item__source__contributor":
                    created_by_contributor
                }):
```
I did this to make the linter happy. I couldn't figure out a way to separate keyword arguments on more than one line (and when it's on one line it's too long for the linter). I'm open to suggestions on better ways to format this. 

## Testing Instructions

* `./scripts/manage test api.tests.FacilityDeleteTest`

## Checklist

- [x] CI passes after rebase


jwalgran (approved):
Excellent work. I tested this manually as well and it works as expected.

jwalgran:
@mstone121 Sorry for introducing some confusion when we chatted about `develop` vs `ogr/develop` before. We need this specific update to our delete behavior on our production OAR branch (`develop`) so that we can use it to clean up production data before we transfer it from OAR to OS Hub. The target for this PR can be moved back to `develop`.

CC @mariel-oar @vrwOAR to double check me on this.

jwalgran:
OS Hub confirmed that we want this behavior in both branches. We have done this for other PRs by opening a second PR and referencing the first PR in the description.

jwalgran (approved):

Issue #2119, opened 2022-09-02
Store raw data file 


mariel-oar:
## Overview
It's very helpful for the team to have access to the raw data file to be able to troubleshoot issues that arise when lists are uploaded. In that spirit, there should be a mechanism to store and retrieve that file on the admin site. 

### Describe the solution you'd like

1. Make the raw file available for download from the list detail page (individual list page) 
2. If easier (i.e. the difference between a 1 pt issue and something bigger), the raw file could instead be accessed from the Django [Facility Lists](https://openapparel.org/admin/api/facilitylist/) page

### Additional Context
There have been two lists recently where it's unclear if the data is of poor quality because of a parsing issue in the system or because of how the file is formatted. The OS Hub team would be able to troubleshoot these lists with more direct access to the raw data file.

PR #2154, opened 2022-09-20
Store raw data

TaiWilkin:
## Overview

When there is an error in list processing, it's sometimes unclear whether that error was caused by an improperly formatted list file or by a bug in the code. Debugging is much more efficient and effective when we have access to the originally uploaded file. We now store the raw data file in S3 when it's uploaded, and enable downloading via both the list details and the Django admin. 

Connects #2119 

## Demo

Screen Shot 2022-09-21 at 9 08 36 AM
Screen Shot 2022-09-21 at 9 09 15 AM

## Testing Instructions

* Contribute a file. It should save the the local file system. (Note: The download buttons won't work.)
* In `docker-compose.yml`, set the `AWS_STORAGE_BUCKET_NAME` to `opensupplyhub-testing-files-eu-west-1`.
* In `settings.py`, change the conditional `if not DEBUG` on line 462 to be True (to cause Django to use S3 instead of the local filesystem)
* Contribute a file. Upload should succeed. 
* Visit the [opensupplyhub-testing-files-eu-west-1](https://s3.console.aws.amazon.com/s3/buckets/opensupplyhub-testing-files-eu-west-1?region=eu-west-1&tab=objects) bucket on S3. The file you uploaded should be there. 
* On the list page, use the 'download submitted csv' button to download the CSV.
* On the Django admin, use the download button in the FacilityList's details to download the CSV. 
* Contribute a file with the same name. View the bucket on S3 and confirm that the original file wasn't overwritten. 
* You should be able to download the second CSV via the list details and Django admin. 
* Visit a list details page that was uploaded on a different branch (ie where the list file hasn't been stored). The list should load without error, and the button for the original file should be hidden. 

## Checklist

- [ ] CI passes after rebase


TaiWilkin:
@jwalgran I haven't built a test branch for this. Do we want to apply this to OAR as well as OSHub? If so, should we do an OAR test branch, so we get it running on a less user-facing environment before we merge to ogr/develop and develop?

jwalgran:
We don't need to make these changes to OAR. I recommend that we coordinate with the OS Hub team when we will be testing this on OS Hub staging

deployment/terraform/iam.tf
```diff
@@ -68,6 +68,31 @@ resource "aws_iam_role_policy" "batch_describe_and_submit" {
...
+      "s3:GetObject",
+      "s3:ListBucket",
+      "s3:PutObject"
+    ]
+  }
+}
+
+resource "aws_iam_role_policy" "s3_read_write_files_bucket" {
+  name   = "S3ReadWriteFiles"
+  role   = aws_iam_role.ecs_task_role.name
```
jwalgran, line 92: When I ran `scripts/infra plan` in a terraform container I got an error on this line. Looking at the other policies in the file, I think we need to use `app_task_role`

```suggestion
  role   = aws_iam_role.app_task_role.name
```

jwalgran (changes requested):
Nice, clear implementation. Thanks for the easy-to-follow commit sequence. I was able to successfully test this locally.

I commented on one terraform error I uncovered when trying to run `plan` and I pushed up  aae665a5 with suggested terraform updates to ensure that we set the env var for all Django process. If you like that implementation you can cherry-pick the commit onto the PR branch.

TaiWilkin:
@jwalgran Ready for another look!

src/app/src/components/FacilityListItems.jsx
```diff
@@ -170,10 +180,22 @@ class FacilityListItems extends Component {
                 onClick={downloadCSV}
                 disabled={downloadingCSV}
             >
-                Download CSV
+                Download Formatted CSV
```
jwalgran, line 183: Not directly related to the new functionality, but since we accept and encourage using Excel files we should consider changing "CSV" here to "File"

jwalgran (approved):
Retested locally. Terraform `plan` runs without error. Looks good. I made one small text-change  suggestion that I noticed in my re-review.

PR #2159, opened 2022-09-21
Update the facility list content when there are no results

jwalgran:
## Overview

Previously a "0" was being rendered because React will still render zero even though it is falsy

We remove the "Search for facilties" button because it is a left over when when filters and results always appeared in separate tabs.

Connects #2049

### Before
Screen Shot 2022-09-21 at 11 33 18 AM

### After

Screen Shot 2022-09-21 at 11 37 26 AM

## Testing Instructions

* Browse http://localhost:6543/facilities?q=DOESNOTMATCHANYTHING and verify the results content

## Checklist

- [x] CI passes after rebase


jwalgran:
Fixing the typo I just noticed

jwalgran:
Fixed the typo

TaiWilkin (approved):
Looks good!

jwalgran:
Thanks for the review.

Issue #2132, opened 2022-09-12
Use the new marker on the OS Hub facility details satellite map


jwalgran:
Currently the legacy marker is being used

<img width="703" alt="Screen Shot 2022-09-12 at 7 53 17 AM" src="https://user-images.githubusercontent.com/17363/189687627-2b7c09ad-50b3-40dc-9c92-fb7dad4f5f72.png">

A screenshot of the new marker in the designs

<img width="684" alt="Screen Shot 2022-09-12 at 8 00 34 AM" src="https://user-images.githubusercontent.com/17363/189687903-3237d6b5-8468-4b2f-881c-4c09df505c81.png">

The SVG for the marker was already added to project, but the Google satellite map may require a PNG version. The marker is in the Map Components section of the Figma https://www.figma.com/file/FGVrJYElirWCQyR7nBehF0/OAR

TaiWilkin:
Noting that the icon doesn't work in development, because Google requests the marker separately, and then can't access our local-host. This ticket isn't expected to fix that issue, just update the icon.

PR #2169, opened 2022-09-26
Update facility static map marker

mstone121:
## Overview

This PR replaces the legacy static selected marker with one that is styled to match the rest of OS Hub.

Connects #2132 

## Demo

Screen Shot 2022-09-26 at 2 58 23 PM

## Notes

This can't be tested locally because google will make a request to fetch the marker icon which fails because it cannot access localhost where the image is hosted.

## Testing Instructions

* No way to test locally.
* In staging, go to a facility details page.
* Check that static map marker is updated

## Checklist

- [ ] CI passes after rebase


jwalgran (approved):
Looks good. I don't think it's worth the effort to do a separate test deploy to staging before merging a simple image swap.

Issue #2081, opened 2022-08-22
Very long download times 


mariel-oar:
# Bug Description 
Download of all US facilities (using country filter) took 8 min. It's about 9k facilities. 
Download started at :29 and finished at :37. File is attached. 

# Expectation 
It should take no more than ~2 minute to download the entire database. 

We should disable this feature until it is possible to download ~20k records within 30s, full db in 2 min. We risk users trying to do multiple downloads, thinking that the initial download didn't work / froze. Is people try to do multiple downloads, it more cost for us. 

As an interim solution, we are enabling download for up to 10,000 facilities; beyond that the download button is disabled.
 
This info should be displayed via tool tip or popup when a search results in >10k facilities: 
"Downloads are supported only for searches resulting in 10,000 facilities or less." 

# Technical Implementation

- Add a waffle switch called `allow_large_downloads` to conditionally control visibility of the download button on the search results
  - Reworked search results in PR #2110 
- If the the switch is disabled, only show the download button if the total number of matching facilities is < 10,000

mariel-oar:
[facilities2_8min.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/9396572/facilities2_8min.csv)

KlausGPaul:
Download of complete facilities list, no filter, on a Sunday, took roughly from 07:28 to 15:26 for 91382 records. 
Using the API, downloading all data incl. extended fields, took from 19:12:57 to 23:42:21, thats 16163 seconds for 91379 records. All times CEDT. API calls were made with a 2...3 second pause between calls, no timeouts were encountered.

obrienad:
tagged as CAN because OAR needs to give green light for bugs.

mariel-oar:
is there a point estimate for this @obrienad ?

mariel-oar:
@jwalgran , is this 1 pt for investigation?

jwalgran:
@mariel-oar The 1-point is to add a feature flag switch so that we can hide the download button until we are able to more quickly generate the download.

mariel-oar:
ok, thanks @jwalgran.

Is it possible to only disable download for searches that return more than 10k facilities?

jwalgran:
Yes. I will adjust the feature description to only disable the download button at >10k

mariel-oar:
thank you @jwalgran! last question on this one: 
is it possible to add helper text (i) next to the download button that says "Downloads are only supported for searches of 10,000 facilities or less."

[...]

PR #2175, opened 2022-09-27
Create allow large downloads flag (#2081)

jacobscarfmerrell:
## Overview

Creates an allow large downloads flag, which if not set to true limits downloads to result sets smaller than 10000 facilities.

Connects #2081

## Testing Instructions
* Run migrations so you can get the new switch
* Because test fixtures seed 901 facilities, you will need to lower the limit here, e.g. to 100:
  * `src/app/src/components/FilterSidebarFacilitiesTab.jsx:230`
* Searching without filters will lead to the download button being disabled and showing a tooltip
* Searching with a filter like "socks" which filters the results down below the threshold should enable the download button again
* Log in to the django admin console and go to http://localhost:8081/admin/waffle/switch/ and activate the switch to allow large downloads
* The download button should be enabled and the tooltip should be missing on every search now

## Checklist

- [ ] CI passes after rebase


src/app/src/components/FilterSidebarFacilitiesTab.jsx
```diff
@@ -222,11 +223,23 @@ function FilterSidebarFacilitiesTab({
...
                         </div>
                     ) : (
-                        <DownloadFacilitiesButton
-                            setLoginRequiredDialogIsOpen={
-                                setLoginRequiredDialogIsOpen
+                        <FeatureFlag
+                            flag={ALLOW_LARGE_DOWNLOADS}
+                            alternative={
+                                <DownloadFacilitiesButton
+                                    disabled={facilitiesCount >= 10000}
```
TaiWilkin, line 230: Can this check be made into a utility function and/or use a constant for the 10,000 (for clarity/DRYness)?

src/app/src/components/NonVectorTileFilterSidebarFacilitiesTab.jsx
```diff
@@ -215,11 +220,23 @@ function NonVectorTileFilterSidebarFacilitiesTab({
...
                 <div style={facilitiesTabStyles.titleRowStyles}>
                     {headerDisplayString}
-                    <DownloadFacilitiesButton
-                        setLoginRequiredDialogIsOpen={
-                            setLoginRequiredDialogIsOpen
+                    <FeatureFlag
+                        flag={ALLOW_LARGE_DOWNLOADS}
+                        alternative={
+                            <DownloadFacilitiesButton
+                                disabled={facilitiesCount >= 10000}
```
TaiWilkin, line 227: See other comment

src/app/src/components/DownloadFacilitiesButton.jsx
```diff
@@ -59,56 +61,73 @@ function DownloadFacilitiesButton({
...
-                onClick={handleClick}
-                style={{
-                    backgroundColor: '#FFCF3F',
-                    fontSize: '16px',
-                    fontWeight: 900,
-                    lineHeight: '20px',
-                }}
-            >
-                <div
+        <Tooltip
```
TaiWilkin, line 64: I'm seeing the tooltip all the time, regardless of whether the download button is enabled or the large downloads flag is enabled. Can it be hidden except for when the button in disabled?

TaiWilkin (approved):
This is working well. I've added a couple of suggestions for tweaks. I don't think they'll require a re-review, but feel free to tag me in if you want me to take another look anyway.

src/app/src/components/DownloadFacilitiesButton.jsx
jacobscarfmerrell: @TaiWilkin that's what I did originally, but then I saw [this](https://github.com/open-apparel-registry/open-apparel-registry/issues/2081#issuecomment-1252689299). I agree, though, and I think that it definitely makes sense to hide when the flag is set to true. Should I only hide it when the flag is true, or whenever the button is disabled?

TaiWilkin: Ah, I missed that. Good catch. I think it's okay to leave the tooltip regardless of whether the button is disabled, then. So only hide it when the flag is true.

Issue #2080, opened 2022-08-22
Updated Sector Taxonomy & Logic 


mariel-oar:
## Overview

This ticket includes requirements for updated processing logic for the sector field. This 

### Included Requirements 

1. Sector should be a combined column with product_type in the upload template (e.g. column name = sector_product_type) 
2. This will be a required field. i.e. the list must be uploaded with sector, product type or both. 
3. This column will be processed in a manner similar to how the processing_type_facility_type column is processed in the OAR 
4. There will be a set list of identified sector values that should be pulled from this new sector_product_type column; the values identified as sectors are then displayed under the sector field on facility profiles; values NOT identified as sectors are displayed as product types. 
5. All OAR data (when migrated) should have "Apparel" added to the sector field. 
6. Sector drop down filter should display only the sector values that are on the identified list of supported sectors AND have been submitted as list data (i.e. do not automatically show the entire list of supported sectors - they may not all be submitted data on the site) 

### Technical Implementation

- Add a `Sector` model with a single unique name field with a max of 200 characters.
  - Update `admin.py` so that `Sector` is editable via the Django admin
-  Create a fixture to load the initial list [sectors.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/9576363/sectors.csv)
  - Make sure the default value of `Unspecified` is also included
- Add `sector_product_type` as a field read from Excel/CSV or API requests. 
  - (OS Hub Team Task) Remove `sector` and `product_type` from the templates
  - Avoid breaking existing API integrations by allowing `product_type` to continue to be used as a field name alias for `sector_product_type`
  - Update API documentation 
- For each value submitted to `sector_product_type` try a case-insensitive match to the list of allowed sectors. If there is a match, add the value to the sector array and exclude it from values that will go into the `product_type` extended field. 
- As discussed with the OS Hub team it is OK to keep the sector as a required field. If none of the values in the `sector_product_type` match the list of allowable sectors, set a default value of `Unspecified`

jwalgran:
Labeled client action needed because we are awaiting a list of Sectors from the OS Hub team

obrienad:
Blocked as Klaus is working on defining the sectors. Timing re:

mariel-oar:
Azavea needs to know the name / code of the list that they need to subscribe people to 
as well as the API token that they need to use 
cc: @vrwOAR @hlennett

KlausGPaul:
[...]

PR #2191, opened 2022-09-28
Don't create an empty product_types extended field when parsing CSV

jwalgran:
## Overview

This `len` check alreay existed in the the facility POST API handler but was missing from the CSV parsing.

Connects #2080

## Testing Instructions

* Verify the logic and that the new unit tests pass CI

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/processing.py
+++ b/src/django/api/processing.py
@@ -187,11 +187,12 @@ def parse_facility_list_item(item):
 
         parser = CsvRowSectorProductTypeParser(fields, values)
         item.sector = parser.sectors
-        if 'product_type' in fields:
-            values[fields.index('product_type')] = parser.product_types
-        else:
-            fields.append('product_type')
-            values.append(parser.product_types)
+        if len(parser.product_types) > 0:
+            if 'product_type' in fields:
+                values[fields.index('product_type')] = parser.product_types
+            else:
+                fields.append('product_type')
+                values.append(parser.product_types)
 
         if CsvHeaderField.COUNTRY in fields:
             item.country_code = get_country_code(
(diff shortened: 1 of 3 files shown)


src/django/api/processing.py
```diff
@@ -187,11 +187,12 @@ def parse_facility_list_item(item):
...
-            values[fields.index('product_type')] = parser.product_types
-        else:
-            fields.append('product_type')
-            values.append(parser.product_types)
+        if len(parser.product_types) > 0:
+            if 'product_type' in fields:
+                values[fields.index('product_type')] = parser.product_types
+            else:
+                fields.append('product_type')
+                values.append(parser.product_types)
```
mstone121, line 195: What about when a user enters only valid sectors under `product_types`? In this case we need to delete the product type field so those values are not also added as product types. Here's a test:

```python
    def test_only_valid_sectors_as_product_type(self):
        facility_list = FacilityList.objects.create(
            header="sector_product_type,address,country,name"
        )
        source = Source.objects.create(
            source_type=Source.LIST, facility_list=facility_list,
            contributor=self.contributor
        )
        item = FacilityListItem.objects.create(
            row_index=1,
            sector=[],
            raw_data="Apparel,1234 main st,ChInA,Shirts!",
            source=source
        )
        parse_facility_list_item(item)
        self.assert_successful_parse_results(item)
        self.assertEqual(['Apparel'], item.sector)
        self.assertEqual(0, ExtendedField.objects.all().count())
```

mstone121 (changes requested):

mstone121 (approved):
Looks good! Thanks for the explanatory comment.

jwalgran:
Thanks for the review.

Issue #2173, opened 2022-09-27
Links from the OS Hub contributor details page to the search page sometimes loose their query string argumes


jwalgran:
## Overview

Links from the OS Hub contributor details page to the search page sometimes loose their query string argumes.

### Expected Behavior

The "View map of facilities" link and the links to the individual lists on the contributor details page navigate to the search page and run the appropriate search

### Actual Behavior

The links have the correct query parameters but the query parameters disappear when the search results load.

NOTE: Browsing a contributor search link directly in a new browser window (like `/facilities?contributors=1710`) does work correctly, but browsing a list search link directly in a new browser window (like `/facilities?contributors=1710&lists=1355`) results in the list query string argument disappearing. 

### Steps to Reproduce

* Browse a facility details page
* Open the contributors drawer
* Click a contributor link
* From the contributor details page click the "View map of facilities" button

### Contributor search

![2022-09-27 09 11 43](https://user-images.githubusercontent.com/17363/192579956-a5a24154-780a-4071-bf75-ae981e9b22bc.gif)

### List search

![2022-09-27 09 25 44](https://user-images.githubusercontent.com/17363/192582160-4b41851d-c593-4169-8cb3-ad3691a18b54.gif)

mariel-oar:
this happens 100% of the time?

PR #2204, opened 2022-10-02
Fix user profile facility map link

mstone121:
## Overview

Previously the user ID was used in the facility map link. This PR updates the link to use contributor_ids from the user's facility lists.

Connects #2173 

## Notes

I'm not sure if it's possible for a user to be associated with lists from multiple contributor IDs. If not, then the user-profile route should be updated to return the contributor_id at the top level (rather than in the nested facility lists).

## Testing Instructions

* Go to a user profile page
* Click "View map of facilities" button
* The facilities page should load with the appropriate contributor filter applied.
* The correct contributor ID should be in the URL params

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/components/UserProfile.jsx
+++ b/src/app/src/components/UserProfile.jsx
@@ -216,7 +216,7 @@ class UserProfile extends Component {
                 {title}
                 <div>
                     <a
-                        href={`/facilities?contributors=${profile.id}`}
+                        href={`/facilities?contributors=${profile.contributorId}`}
                         rel="noopener noreferrer"
                         style={{
                             backgroundColor: '#FFCF3F',
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -110,6 +110,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Update the Python and Rollber versions used for lamba functions [#2120](https://github.com/open-apparel-registry/open-apparel-registry/pull/2120)
 - Fixup recent UX changes [#2141](https://github.com/open-apparel-registry/open-apparel-registry/pull/2141)
 - Fixed parsing country strings with newline characters [#2195](https://github.com/open-apparel-registry/open-apparel-registry/pull/2195)
+- Fixed user profile facilites map link [#2204](https://github.com/open-apparel-registry/open-apparel-registry/pull/2204)
 
 ### Security
 
--- a/src/app/src/reducers/ProfileReducer.js
+++ b/src/app/src/reducers/ProfileReducer.js
@@ -160,6 +160,7 @@ export default createReducer(
                                 profileSummaryFieldsEnum.facilityLists
                             ],
                     },
+                    contributorId: { $set: payload.contributor_id || null },
                 },
                 fetching: { $set: false },
                 error: { $set: null },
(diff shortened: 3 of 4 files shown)


src/app/src/components/UserProfile.jsx
```diff
@@ -205,6 +205,10 @@ class UserProfile extends Component {
             </div>
         );
 
+        const contributorIdsString = profile.facilityLists
+            .map(list => list.contributor_id)
+            .join('&contributors=');
```
jwalgran, line 210: This is a resourceful way of dealing with this issue without updating the API, however if a contributor submits facilities exclusively through the API, they will not have any lists, but they will have associated facilities.

To set up this situation I:

- Registered an confirmed a new account
- Generated an API token on the settings page
-  In a separate browser session logging in as c1@example.com, browsed the new user account in the Django admin and added all the permission groups
Screen Shot 2022-10-03 at 1 35 05 PM
- POSTed a facility

```bash
curl -X POST "http://localhost:8081/api/facilities/" -H  "accept: application/json" -H  "Authorization: Token {TOKEN_VALUE_HERE}" -H  "Content-Type: application/json" -d "{    \"sector_product_type\": \"Apparel\",    \"country\": \"China\",    \"name\": \"Nantong Jackbeanie Headwear Garment Co. Ltd.\",    \"address\": \"No.808,the third industry park,Guoyuan Town,Nantong 226500.\"}"
```

The change below appears to work in this API-only case and in my regression testing I was still able to register a new account and edit the profile of a logged in account.

```diff
diff --git a/src/app/src/components/UserProfile.jsx b/src/app/src/components/UserProfile.jsx
index ff5f18ba..b563edfd 100644
--- a/src/app/src/components/UserProfile.jsx
+++ b/src/app/src/components/UserProfile.jsx
@@ -216,7 +216,7 @@ class UserProfile extends Component {
                 {title}
                 <div>
                     <a
-                        href={`/facilities?contributors=${profile.id}`}
+                        href={`/facilities?contributors=${profile.contributorId}`}
                         rel="noopener noreferrer"
                         style={{
                             backgroundColor: '#FFCF3F',
diff --git a/src/app/src/reducers/ProfileReducer.js b/src/app/src/reducers/ProfileReducer.js
index 71b98a47..57b636d4 100644
--- a/src/app/src/reducers/ProfileReducer.js
+++ b/src/app/src/reducers/ProfileReducer.js
@@ -160,6 +160,7 @@ export default createReducer(
                                 profileSummaryFieldsEnum.facilityLists
                             ],
                     },
+                    contributorId: { $set: payload.contributor_id || null },
                 },
                 fetching: { $set: false },
                 error: { $set: null },
diff --git a/src/django/api/serializers.py b/src/django/api/serializers.py
index 41ceb894..cf8c705e 100644
--- a/src/django/api/serializers.py
+++ b/src/django/api/serializers.py
@@ -267,12 +267,14 @@ class UserProfileSerializer(ModelSerializer):
     facility_lists = SerializerMethodField()
     is_verified = SerializerMethodField()
     embed_config = SerializerMethodField()
+    contributor_id = SerializerMethodField()
 
     class Meta:
         model = User
         fields = ('id', 'name', 'description', 'website', 'contributor_type',
                   'other_contributor_type', 'facility_lists', 'is_verified',
-                  'embed_config')
+                  'embed_config', 'contributor_id')
+        read_only_fields = ('contributor_id',)
 
     def get_name(self, user):
         try:
```

jwalgran (changes requested):

src/app/src/components/UserProfile.jsx
mstone121: I didn't realize the contributor ID was associated directly with the user. Thanks for the suggestion! Updated in e16c22e

jwalgran (approved):

PR #2205, opened 2022-10-02
Don't render components before query param hydrate

mstone121:
## Overview

Some components fetch data from the API. If these components render before the query params have been hydrated into state, they'll reference incomplete data when making API requests.

This commit delays render of components that use the `withQueryStringSync` "hook" until after state has been hydrated from the query parameters. A circular progress is rendered in its place.

Connects #2173 

## Demo

https://user-images.githubusercontent.com/8356789/193476652-bb3bdcff-f0c6-400d-a473-517e892d9644.mov

## Notes

I moved the on-mount facility fetch out of the hydrate function. This allows setting the component to its finished state (and thus rendering the wrapped component) before the facilities are done fetching.

## Testing Instructions

* Go to a user profile page with at least one facility list
* Click "View facilities" on a facility list
* Both the contributor and the facility list should be selected in the filters

## Checklist

- [x] CI passes after rebase


jwalgran (approved):
Excellent job reworking the event handling. I did not find any regressions in behavior.

PR #2207, opened 2022-10-03
Fix parse errors (OSHub)

TaiWilkin:
## Overview

Duplicate pointed at OS Hub for [#2206](https://github.com/open-apparel-registry/open-apparel-registry/pull/2206).

In a recently uploaded list on OAR staging, several items failed to parse. 

When a facility type or processing type is submitted with a pipe as
the final character, it creates an empty string as the final
value in the array. When we attempted to process the empty string,
it was cleaned to a None value, which threw an error. Empty strings are
now processed as 'No match' or (None, None, None, None).

Also, when a list of items is provided as an extended field value, if all
items in the array are empty strings, don't make the extended field.
This prevents a completely blank extended field from being created. 
These were being created for `product_type` for some items in the
failed list. 

Connects #[104](https://github.com/open-apparel-registry/open-apparel-registry-clients/issues/104)

## Demo
Screen Shot 2022-10-03 at 11 44 06 AM

## Notes

I marked the empty string as "No match" instead of filtering it out to keep the number of items consistent between the raw and processed data. 

## Testing Instructions

* Run `./scripts/server` and sign in as c2@example.com. 
* Contribute [parse_failure_items.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/9699922/items.csv). Upload should succeed.
* Run `./tools/batch_process {list_id}`. The list should parse successfully. 
* In the items from the list, there should be no blank `product_type` field, and the `processing_type`/`facility_type` fields should show the correct values from the CSV. 

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixup recent UX changes [#2141](https://github.com/open-apparel-registry/open-apparel-registry/pull/2141)
 - Fixed parsing country strings with newline characters [#2195](https://github.com/open-apparel-registry/open-apparel-registry/pull/2195)
 - Fixed user profile facilites map link [#2204](https://github.com/open-apparel-registry/open-apparel-registry/pull/2204)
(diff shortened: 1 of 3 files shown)


maurizi (approved):
tested, working👍

TaiWilkin:
Thank you for reviewing!

Issue #2183, opened 2022-09-28
Update info.openapparel.org links for OSHub


TaiWilkin:
#### Info site URLs
All URLs that link to info sites at info.openapparel.org need to be updated to: info.opensupplyhub.org: 

1. On the [Contribute Page](https://9f692df0338dcbc9848646c6.openapparel.org/contribute), the URL that links to step by step instructions needs to be updated: 

- FROM https://info.openapparel.org/stories-resources/how-to-contribute-data-to-the-oar 
- TO https://info.opensupplyhub.org/stories-resources/how-to-contribute-data 

*Note that the URL above ends at "contribute-data" and that the "to-the-oar" part is dropped

2. On the Footer: links for Contact Us, Work With Us, FAQs, Privacy Policy, and Terms of Service change from a root url of https://info.openapparel.org/[page] TO https://info.opensupplyhub.org/[page]
- Contact Us: https://info.opensupplyhub.org/contact-us
- Work With Us: https://info.opensupplyhub.org/work-with-us
- FAQs: https://info.opensupplyhub.org/faqs
- Privacy Policy: https://info.opensupplyhub.org/privacy-policy
- Terms of Service: https://info.opensupplyhub.org/terms-of-service
- Donate: http://info.opensupplyhub.org/donate

3. On the Header: 
- Donate: http://info.opensupplyhub.org/donate

TaiWilkin:
#2143 covers this

jwalgran:
@TaiWilkin Based on your comment it looks like #2143 covers. Please check that there isn't anything noting in this issue that has not already been implemented.

TaiWilkin:
@jwalgran We re-opened this because we'll need to do a final switch-over to the new root URL for the info site, and a check of all info site links, to confirm everything is working once the info site is live.

PR #2214, opened 2022-10-04
Update info site URL

mstone121:
## Overview

The info site is changing from [info.openapparel.org](https://info.openapparel.org) to [info.opensupplyhub.org](https://info.opensupplyhub.org).

Connects #2183 

## Testing Instructions

* Test all header info links
* Test all footer info links
* Test email template links work
* Test API schema view terms link works

While info.opensupplyhub.org is not set up, none of these links will "work", but their URLs can be verified.

## Checklist

- [x] CI passes after rebase


src/django/api/fixtures/contributors.json
```diff
@@ -1,1377 +1,1374 @@
 [
     {
         "model": "api.contributor",
-        "pk": 1,
```
jwalgran, line 4: In 1073cd3 we intentionally make it so that the `pk` of the contributor and `admin_id` were not the same value, which makes it easier to catch bogs related to using the wrong ID.

We will want to make sure our `pk`s continue to start from 1

Because the updated URL does not show up in any other fixtures, rather than committing new fixtures we could instead just find-and-replace `https://info.openapparel.org"` with `https://www.example.com`

src/django/api/fixtures/contributors.json
```diff
@@ -1,1377 +1,1374 @@
...
+        "pk": 2,
         "fields": {
             "admin": 2,
-            "name": "Service Provider A",
-            "description": "A service provider dedicated to transparency in apparel supply chains",
-            "website": "https://info.openapparel.org",
```
jwalgran, line 9: Because the updated URL does not show up in any other fixtures, rather than committing new fixtures we could instead just find-and-replace `https://info.openapparel.org"` with `https://www.example.com`

jwalgran (changes requested):
I made a suggestion to fix the donate link and simply the fixture changes.

src/django/api/tests.py
```diff
@@ -8496,11 +8496,11 @@ class ContributorManagerTest(TestCase):
     fixtures = ['users', 'contributors']
 
     def test_filter_by_name(self):
-        matches = Contributor.objects.filter_by_name('factory a')
+        matches = Contributor.objects.filter_by_name('facility a')
         self.assertGreater(matches.count(), 0)
 
         # No result should be returned for exact match
-        matches = Contributor.objects.filter_by_name('factory')
+        matches = Contributor.objects.filter_by_name('facility')
```
jwalgran, line 8503: Now that the changes to the contributors fixture have been simplified we need to revert this test change.

jwalgran (approved):
Changes look good. CI is failing on a test and I believe all we need to do is revert the change now that the contributors fixture updates have been simplified.

src/django/api/tests.py
mstone121: Right, I forgot about this. I rebased on `ogr/develop` and removed the commit with these changes.

Issue #2188, opened 2022-09-28
Update Google Analytics Key for OS Hub


obrienad:
Update Analytics Key for OS Hub.

Consider updating to [Google Analytics 4](https://support.google.com/analytics/answer/10089681?hl=en) since Google Google Analytics Universal Analytics tag is going to be deprecated next year.

hlennett:
We do have one set up for OS Hub. Property ID: 320730191. Is there any specific code or info you need?

nanotubing:
@hlennett the value we need might be called Tracking ID in google analytics? it starts with `UA-`, followed by a string of numbers. Thanks!

hlennett:
How this: G-PFQY2JR2KH. I think with GA-4 the format may have changed.

nanotubing:
Thanks, I'll give it a try!

mariel-oar:
@TaiWilkin here are the measurement IDs for each site: 

OS Hub info: G-3N9248MR8P
OS Hub prod: G-GSLW5DLZQY
OS Hub staging: G-GS2XMYGRN9

PR #2221, opened 2022-10-05
Upgrade Google Analytics to gtag.js

TaiWilkin:
## Overview

`analytics.js`, an older version of Google Analytics, can't be used with Google Analytics 3. In July 2023, Universal Analytics, which we previously used, will no longer be available for use. In order to enable GA4, we switch to using `gtag.js`.

Connects #2188 

## Demo

Screen Shot 2022-10-05 at 3 23 47 PM

## Notes

Universal Analytics keys will still work with `gtag.js` until July 2023. We currently have a UA key on OSHub staging but a GA4 key on production. 

## Testing Instructions

* Update your `.env` to set `REACT_APP_GOOGLE_ANALYTICS_KEY` equal to the new G-XXXXX (GA4 formatted) id. 
* In `public/index.html` replace `%REACT_APP_GOOGLE_ANALYTICS_KEY%` with the GA4 id 
     - This should happen automatically in production. See [docs](https://create-react-app.dev/docs/adding-custom-environment-variables/#referencing-environment-variables-in-the-html).
* Change `util.ga.js` line 35 to `if (!environment) {` (to allow analytics to run in development).
* Run `./scripts/server`
* Visit http://localhost:6543/settings and check if your analytics permissions are on or off. Turn them on (approve them) if they are off. 
* Open the network tab in the developer tools and refresh the page. 
* Search for requests containing the GA4 key. You should see several. In one of the payloads, you should see "User consented to GA tracking".
* Click around the site and you will see additional requests. 
* Go back to settings and reject analytics. Refresh the page. 
* Now in the network tab, you should only see two GET requests and no POSTs. Clicking around should send no POST requests with user data. 

## Checklist

- [ ] CI passes after rebase


src/app/src/util/util.ga.js
```diff
@@ -170,10 +153,6 @@ export const startGATrackingIfUserHasAcceptedNotification = () => {
             anonymizeIp: true,
         });
 
-        window.ga('send', 'pageview', {
```
TaiWilkin, line 173: A pageview event is sent automatically as part of `config`.

nanotubing (approved):
nice job with the testing instructions, they were very easy to follow.

Firefox blocks a lot on its own, so I ended up switching to Chrome, which was happy to let the request through. I see no problems with this PR

TaiWilkin:
For people looking at this as reference in the future, we did a followup fix here: https://github.com/open-apparel-registry/open-apparel-registry/pull/2229

PR #2223, opened 2022-10-06
Show components with no query parameters

mstone121:
## Overview

Recent changes to the `withQueryStringSync` function prevented components that use it from rendering if the query parameters have not hydrated. If a component does not have query parameters, it would be stuck in the loading state.

Connects #2173 

## Testing Instructions

* Go to https://localhost:6543/facilities directly
* Filter components/page should load

## Checklist

- [x] CI passes after rebase


TaiWilkin (approved):
Thanks for this quick fix. I visited the facilities page via multiple paths and with a variety of filters set and was unable to cause the infinite spinner with the fix.

Issue #2220, opened 2022-10-05
Superuser unable to remove list items from pending lists 


mariel-oar:
### Overview 
There is a button to remove a facility in the list detail view. This button does not currently work. 

### Desired Behavior 
Required: 
1) PENDING LIST - Clicking remove on a facility when a list is in pending status, results in the facility being removed from the list and not geocoded (in the event that list is approved) 

Nice to have: 

2) MESSAGE - Similar to when a list is rejected, there should be a popup when a facility is removed from a pending list. This is free form text box for OS Hub team to message what the issue is. This then triggers an email with OS Hub team's feedback to go to the contributor. 

### Additional Context: 
![Screen Shot 2022-10-05 at 11.34.20 AM.png](https://images.zenhubusercontent.com/61f9cafd1f9027ce23332739/f17482f1-e27b-40a6-bd66-15fdbb26002b)!

### Technical Implementation

The API is returning 404

<img width="1214" alt="Screen Shot 2022-10-07 at 9 05 35 AM" src="https://user-images.githubusercontent.com/17363/194598445-50e83052-3b0e-4b3d-acfe-fbfab6c4cda7.png">

This is most likely because the `remove_item` view function includes

```python
            facility_list = FacilityList \
                .objects \
                .filter(source__contributor=request.user.contributor) \
                .get(pk=pk)
```

and this associated except block

```python
        except FacilityList.DoesNotExist:
            raise NotFound()
```

Before we updated the workflow to allow the moderation team to handle approve/reject on lists we enforced that only the owner of the list could take action on it

- Update `remove_item` to allow superusers to remove items from a list

jwalgran:
- In which  environment did you run into this error, OAR production, OAR staging, or OS Hub staging? 
- From which list did you create that screenshot?

I was able to remove an item from a test list in OS Hub staging, so that points to there being a problem with a specific data situation, not the entire feature.

<img width="1219" alt="Screen Shot 2022-10-05 at 11 09 18 AM" src="https://user-images.githubusercontent.com/17363/194132001-10c84474-502c-4097-bbfa-c40a211a5f15.png">

mariel-oar:
OS Hub staging

mariel-oar:
@obrienad , do you know if the nice to have on this is being implemented?: 

Nice to have:

[...]

PR #2228, opened 2022-10-07
Update facility list item removal

mstone121:
## Overview

The PR does two things:
- Allow superusers to remove facility list items
- Allow facility list items to be removed from pending lists

Connects #2220 

## Testing Instructions

* `./scripts/manage test api.tests.ConfirmRejectAndRemoveAndDissociateFacilityMatchTest`
  * `test_superuser_can_remove_list_item`
* Login in as superuser
* Go to a pending list
* Remove a list item

Also,

* Login as a user with a pending list, e.g. c3@example.com
* Go to the pending list and remove a list item

## Checklist

- [x] CI passes after rebase


TaiWilkin:
@mstone121 I was able to remove list items with statuses 'NEW_FACILITY', 'CONFIRMED_MATCH', 'MATCHED' on a pending list, but I wasn't able to remove pending matches because the 'remove' button didn't display for them. Is that the correct process for testing, or am I missing a step?

jwalgran:
I focused on the wrong problem when writing the technical implementation section for #2220. The issue description describes wanting to have moderators be able to remove items from a list before it has been approved and run through the geocoding matching steps (when items have a list status of `PARSED`). This is a new feature. 

The current "REMOVE" button has POSTs to a view that deactivates match rows. To "deactivate" an item before it has been matched will require adding a new status value for list items. `REMOVED` already exists as a "meta status." One potential solution is to add an `ITEM_REMOVED` status to `FacilityListItem` and then modify the `special_case_q_statements`  to classify an item as `REMOVED` if it has either the new `ITEM_REMOVED` status or an inactive match 

https://github.com/open-apparel-registry/open-apparel-registry/blob/7f9edb659a96374a4dd6f48190ba782bd8994caf/src/django/api/views.py#L3167

We can then change the implementation of `remove_item` to either the existing logic if there are one or more `matches_to_deactivate`, otherwise it can set the status of the item to `ITEM_REMOVED` and add a processing result in a similar way to when a facility is deleted

https://github.com/open-apparel-registry/open-apparel-registry/blob/7f9edb659a96374a4dd6f48190ba782bd8994caf/src/django/api/views.py#L1719-L1727

mstone121:
@TaiWilkin @jwalgran This is ready for review again.

Issue #2235, opened 2022-10-12
(Final) Terminology Updates


mariel-oar:
### Description 
- This ticket includes all final copy changes / tweaks for OS Hub. 

### Requirements: 

1. Footer update - REPLACE: "Open Apparel Registry, Inc., doing business as Open Supply Hub, is established as a 501(c)(3)..." WITH: "Open Supply Hub is established as a 501(c)(3)" 
![Screen Shot 2022-10-11 at 9.10.06 PM.png](https://images.zenhubusercontent.com/61f9cafd1f9027ce23332739/1a469751-23cb-4182-b145-affc5d8b12d8)
2. Update copy for reject notification email: https://docs.google.com/document/d/16bXhAUnvYkjeTibNGJSUafTb7rtsL6XyT2ajIpPYR8I/edit?usp=sharing
3. Donate link on Footer update to: https://givebutter.com/opensupplyhub2022
4. Reporting link link on the Footer should be: https://opensupplyhub.allvoices.co/
4. The additional fields helper text popup that displays when you hover: remove "the" before Open Supply Hub 
![Screen Shot 2022-10-12 at 2.03.39 PM.png](https://images.zenhubusercontent.com/61f9cafd1f9027ce23332739/60326f45-7251-4ca3-86ed-26dece645db1)

hlennett:
In the Embed tab in settings, "Contribute your supplier data" links to an OAR link and the email address is the OAR email address. Can we please also use the following copy (small tweaks made throughout):

To begin, [upload your supplier data](https://openapparel.org/contribute) (via upload or API) to OS Hub with all of the data fields you wish to have displayed on your map using the template supplied to you by the OS Hub Team. This list must include any additional data points you would like to display on your customized map, such as facility type, number of workers etc. The Embedded Map will display all facilities included in your lists.

Adjust the below settings to your liking, such as the map color and size. Once complete, copy the Embed Code to add to your website.

Have questions? Check out the FAQs on our [Embedded Map info page](https://info.opensupplyhub.org/embedded-map) or email [info@openapparel.org](mailto:info@openapparel.org).

hlennett:
The Embed page also refers to the OS Hub ID. Should it be OS ID?

hlennett:
On the upload data page:

can we update this sample text:
"example: 'This is the Alpha Brand list of suppliers for their apparel products valid from June 2021 to Sept 2021'

to:
[...]

PR #2237, opened 2022-10-12
OS Hub Terminology Updates

mstone121:
## Overview

This PR updates copy in a few places on the website.

A few uniform changes:
- `info@openapparel.org` -> `info@opensupplyhub.org`
- `the OS Hub` -> `OS Hub` (where it makes sense, i.e. not when followed by "team")

Connects #2235 

## Notes

This PR updates the header donate link even though #2235 specifies only that the footer link be updated. I assumed they should point to the same location.

## Testing Instructions

* Login
* Footer copy is updated/correct
* Contribute page copy is updated/correct
* Embedded map settings page copy is updated/correct
* No references to the openapparel URL
* No references to "the OS Hub"

## Checklist

- [ ] CI passes after rebase


jwalgran (approved):
These changes look good. 

I did not find any remaining instances of info@openapparel.org
It does make sense to update all of the "Donate" links.

I did find one suggested additional change from "the OS Hub" to "OS Hub" in two email templates:

https://github.com/open-apparel-registry/open-apparel-registry/blob/15d3b969420896c607937a3d249878580f14c221/src/django/api/templates/mail/facility_list_complete_body.html#L19

https://github.com/open-apparel-registry/open-apparel-registry/blob/15d3b969420896c607937a3d249878580f14c221/src/django/api/templates/mail/facility_list_complete_body.txt#L10

Issue #2231, opened 2022-10-11
On initial load of the OS Hub homepage the compact map and sidebar are small before resizing


jwalgran:
**NOTE** The fix for this should be time boxed to 1 hour 

This results in 
- A redraw with a lot of page motion once the content is ready
- Potential map rendering issues

This was most easily repeatable in an incognito window. I was able to repeat it in both Firefox and Chrome

![2022-10-11 11 29 31](https://user-images.githubusercontent.com/17363/195172174-55124bdf-31bd-4c01-a736-b183c9f1434f.gif)

### Technical implementation

- Set height based on window minus header
- Center spinner

mariel-oar:
@obrienad @jwalgran, can you provide more details about what part of this is repeatable? or what conditions are required for repeatability? I am not able to see this behavior in chrome. 

also a point estimate would be helpful in deciding how much of a priority this is

jwalgran:
We will be having a planning meeting shortly so we can provide more detail and an estimate within the hour.

obrienad:
It is a 1 (point, but also hour). Ability to reproduce it might be affected by data being loaded or the map is changing - those things might make it easier to see the problem.

PR #2238, opened 2022-10-12
Fix initial page load layout issue

mstone121:
## Overview

Loading the homepage for the first time can result in some odd looking page layouts. This PR sets a fixed height and width on the loading sidebar so the page renders closer to how it will eventually look. 

Connects #2231 

## Notes

There is still a flash of a loading circle in the beginning. I believe this is do to the loading of feature flags from the database. Also, since the map is not rendered before this, these changes do fix the map loading glitch.

This PR also sets the overflow on the sidebar to auto.

## Testing Instructions

* Open a private window (no cache)
* Go to http://localhost:6543/
* See that page load looks good

## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/app/src/components/HomepageSidebarSearch.jsx
+++ b/src/app/src/components/HomepageSidebarSearch.jsx
@@ -142,7 +142,16 @@ function FilterSidebarSearchTab({
 
     if (fetchingOptions) {
         return (
-            <div className="control-panel__content">
+            <div
+                className="control-panel__content"
+                style={{
+                    minWidth: '50vw',
+                    minHeight: '90vh',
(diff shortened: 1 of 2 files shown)


jwalgran (approved):
I added throttling to slow down the browser a bit and compared the hard refresh behavior of `ogr./develop` vs this branch. The new loading behavior looks better.

@TaiWilkin do you have any concerts with the new style?

TaiWilkin (approved):
LGTM. I loaded the facilities page and an embedded map as well as the homepage in desktop and mobile view with no issues.

Issue #2246, opened 2022-10-17
Splitting a facility on OS hub results in KeyError: 'previous_facility_os_id' 


jwalgran:
## Overview

Attempting to split a facility results in an unhanded exception trying to reference `previous_facility_os_id`, likely due to the fact that this was renamed from `previous_facility_oar_id`

https://rollbar.com/OpenApparelRegistry/OpenSupplyHub/items/28/

### Additional context
<img width="566" alt="Screen Shot 2022-10-17 at 8 12 23 AM" src="https://user-images.githubusercontent.com/17363/196215195-b8d93fa3-88be-48fd-a41c-2133b5d76936.png">

PR #2251, opened 2022-10-18
Add management command to replace oar_id in JSON fields

jwalgran:
## Overview

The `FacilityListItem` and `FacilityMatch` models both have JSON fields which can contain keys that reference oar_id rather than os_id. Splitting a facility was failing with an error due to the code expecting a key name containing os_id.

Despite the fact that it is a one time operation, we wrote this as a management command rather than a data migration to facilitate testing with a "dry run" option.

Connects #2246
Connects #2252

### Setup

In `./scripts/manage shell_plus` add some test data and leave the shell open

```
i = FacilityListItem.objects.all().first()
i.processing_results.append({'action': 'merge_facility', 'merged_oar_id': 'ZXY987'})
i.processing_results.append({'action': 'move_facility', 'previous_facility_oar_id': 'ABC123'})
i.processing_results.append({'action': 'split_facility', 'previous_facility_oar_id': 'DDD434'})
i.save()

m = FacilityMatch.objects.all().first()
m.status['split_from_oar_id'] = 'HHH888'
m.save()
```

### Test

- Run ` ./scripts/manage replace_oar_id --dry-run` and verify that it prints a single item and a singly match to the console  with "dry" in the message
- In the Django shell verify that the changes to the key names were not saved
```
i.i.refresh_from_db()
i.processing_results

m.refresh_from_db()
m.results
```
- Run ` ./scripts/manage replace_oar_id` and verify that it prints a single item and a singly match to the console with "save" in the message
- Run ` ./scripts/manage replace_oar_id` and verify  that it does not print any items or matches
- In the Django shell verify that the key names have been changed
```
i.i.refresh_from_db()
i.processing_results

m.refresh_from_db()
m.results
```

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/management/commands/replace_oar_id.py
+++ b/src/django/api/management/commands/replace_oar_id.py
@@ -1,0 +1,61 @@
+from django.core.management.base import BaseCommand
+from django.db.models import Q
+
+from api.models import FacilityListItem, FacilityMatch
+
+
+class Command(BaseCommand):
+    help = ('Replace references to oar_id in item processing_results and '
+            'match results')
+
+    def add_arguments(self, parser):
+        parser.add_argument('-d', '--dry-run', action='store_true',
+                            default=False, help='Do not save cahnges')
+
+    def handle(self, *args, **options):
+        dry_run = options['dry_run']
+
+        if dry_run:
+            print('== DRY RUN - SKIPPING ALL SAVES')
+
+        print('-- PROCESSING ITEMS')
+
+        filters = (
+            Q(processing_results__contains=[{"action": "move_facility"}])
+            | Q(processing_results__contains=[{"action": "merge_facility"}])
+            | Q(processing_results__contains=[{"action": "split_facility"}])
+        )
+
+        for item in FacilityListItem.objects.filter(filters).iterator():
+            did_update_key = False
+            for result in item.processing_results:
+                if 'previous_facility_oar_id' in result:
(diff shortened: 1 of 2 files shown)


src/django/api/management/commands/replace_oar_id.py
```diff
@@ -0,0 +1,61 @@
...
+        parser.add_argument('-d', '--dry-run', action='store_true',
+                            default=False, help='Do not save cahnges')
+
+    def handle(self, *args, **options):
+        dry_run = options['dry_run']
+
+        if dry_run:
+            print('== DRY RUN - SKIPPING ALL SAVES')
+
+            print('-- PROCESSING ITEMS')
```
TaiWilkin: Suggest decreasing indent so this prints outside of dry runs

TaiWilkin (approved):
Noting that on successive runs, the previously saved items are found but skipped. 

Working well. I think having a dry run option for this command was the right choice, since working with production data in these JsonB arrays can be a bit unpredictable.

src/django/api/management/commands/replace_oar_id.py
jwalgran: Good catch. Thank you.

jwalgran: Fixed in the rebase.

jwalgran:
Thanks for the review.

PR #2257, opened 2022-10-19
Allow default batch job memory and vCPU count to be set by vars

jwalgran:
## Overview

Allows changing resources without a code deployment and using different resource allocations in different environments.

Connects https://github.com/open-apparel-registry/open-apparel-registry-clients/issues/107

## Demo

Screen Shot 2022-10-19 at 12 01 33 PM

## Testing Instructions

* Edit the default values `batch_default_job_vcpus` and `batch_default_job_memory` values in `variables.tf` so they are different
* `docker-compose -f docker-compose.ci.yml run --rm terraform`
* `bash-5.1# ./scripts/infra plan`
* In the `plan` output find the `aws_batch_job_definition.default` resource and verify that the `memory` and `vcpus` values match what was set in the variables file.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
+- Allow default batch job memory and vCPU count to be set by vars [#2257](https://github.com/open-apparel-registry/open-apparel-registry/pull/2257)
+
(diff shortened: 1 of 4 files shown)


nanotubing (approved):
I updated the memory and cpu, and it tried to update the batch_job_definition with the correct value

jwalgran:
Thanks for the review.

PR #2259, opened 2022-10-20
Fix logging when checking for existing Hubspot contact

jwalgran:
## Overview

The boolean comparison was mistakenly inverted. If we get a 409 when attempting to create a contact then that contact already exists and we want to continue with subscibing them to the list.

Connects https://github.com/open-apparel-registry/open-apparel-registry-clients/issues/105

## Testing Instructions

* Verify that my interpretation of the logic is correct

## Checklist

- [x] CI passes after rebase


TaiWilkin (approved):
Yes, this was mistakenly inverted. Good catch.

Issue #2233, opened 2022-10-11
The sector field for 2 production facilities does not index correctly 


jwalgran:
## Overview

After testing the migration of OAR production data to OS Hub 2 facilities were missing a value a sector value in the index

```
PK2022244AQE2SN
PK202123038E6DH
```

Determine why the query used to find the sectors for these facilities returns `{}` rather than `{Apparel}`

### Demo

```
opensupplyhub=> select id, sector from api_facilityindex where sector = '{}' and id in (select id from api_facility);
       id        | sector
-----------------+--------
 PK202123038E6DH | {}
 PK2022244AQE2SN | {}
(2 rows)
```

### Additional context

Dumping out the query created by `get_sector_dict` results in the following

```
SELECT "api_facility"."id",
       array
       (
                       select distinct
                       ON (
                                                       "api_source"."contributor_id") unnest(u2."sector") AS "values"
                       FROM            "api_facilitymatch" u0
                       INNER JOIN      "api_facilitylistitem" u2
                       ON              (
                                                       u0."facility_list_item_id" = u2."id")
                       INNER JOIN      "api_source"
                       ON              (
                                                       u2."source_id" = "api_source"."id")
                       WHERE           (
                                                       u0."facility_id" = "api_facility"."id"
                                       AND             u2."status" IN ('MATCHED',
                                                                       'CONFIRMED_MATCH'))
                       ORDER BY        "api_source"."contributor_id" ASC,
                                       "api_source"."is_active" DESC,
                                       u0."updated_at" DESC) AS "item_sectors",
       array
       (
                       SELECT DISTINCT unnest(u0."sector") AS "values"
                       FROM            "api_facilityclaim" u0
                       WHERE           (
                                                       u0."facility_id" = "api_facility"."id"
                                       AND             u0."status" = 'APPROVED')) AS "claim_sectors"
FROM   "api_facility"
 WHERE  "api_facility"."id" IN ('PK2022244AQE2SN', 'PK202123038E6DH')

       id        | item_sectors | claim_sectors
-----------------+--------------+---------------
 PK202123038E6DH | {}           | {}
 PK2022244AQE2SN | {}           | {}
```

obrienad:
This is in the priority backlog because either the data needs to be fixed for these two items or the query to index sectors needs to be adjusted.

PR #2261, opened 2022-10-20
Exclude rejected match items from merge transfer

TaiWilkin:
## Overview

List items can produce multiple matches, each linked to a different facility. When these matches are PENDING or REJECTED, the list item will point to a different facility than the match, as the list item will have additional valid match(es) linked to that different facility.

When merging facilities, we were previously incorrectly updating PENDING and REJECTED matches' list items to point at the target facility, when instead they should have been left alone, as the list items don't belong to the merged facility.

Also adds a management command to run once to fix list items which were impacted by this bug.

Connects #2233 

## Testing Instructions

On `ogr/develop`:
* Go to http://localhost:6543/dashboard/lists and find a list with a pending match. 
* Note the suggested matched facility's ID. (We will now refer to this as Facility A.)
* Reject the pending match. (If there are multiple pending matches for the list item, reject them all.)
* A new facility will be created from the list item. (We will now refer to this as Facility B). Open this in a new tab. You should see that it has an assigned sector.
* In the original tab, go to http://localhost:6543/dashboard/mergefacilities
* Get another facility ID (Facility C) and set this as the target facility. Set the merge facility to Facility A. Complete the merge. 
* Visit Facility C and confirm that Facility A has been merged into it. 
* Visit Facility B and note that it no longer has a sector. 

On this branch:
* Repeat the above steps with new facilities (Facility A-2, B-2, C-2). Facility B-2 should NOT lose its sector(s).
* Visit Facility B and note that it still has no sector.
* Run `./scripts/manage fix_merged_rejections --dry-run`. You should see that 1 item will be adjusted. 
* Run `./scripts/manage fix_merged_rejections`. One item should be adjusted. 
* Visit Facility B and note that it has its sector back. 

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/management/commands/fix_merged_rejections.py
+++ b/src/django/api/management/commands/fix_merged_rejections.py
@@ -1,0 +1,79 @@
+from django.core.management.base import BaseCommand
+from django.utils import timezone
+
+from api.models import FacilityMatch, HistoricalFacilityMatch
+from api.constants import ProcessingAction
+
+
+class Command(BaseCommand):
+    help = ('Reset incorrectly merged facility list items.')
+
+    def add_arguments(self, parser):
+        parser.add_argument('-d', '--dry-run', action='store_true',
+                            default=False, help='Do not save cahnges')
+
+    def handle(self, *args, **options):
+        dry_run = options['dry_run']
+
+        if dry_run:
+            print('== DRY RUN - SKIPPING ALL SAVES')
+
+        count = 0
+        now = str(timezone.now())
+        for hist_match in HistoricalFacilityMatch.objects.filter(
+            status='MERGED'
+        ).iterator():
+            try:
+                prev_record = hist_match.prev_record
+                if prev_record and prev_record.status in ['REJECTED',
+                                                          'PENDING']:
+                    list_item = prev_record.facility_list_item
+                    valid_matches = FacilityMatch \
+                        .objects \
+                        .filter(facility_list_item_id=list_item.id,
+                                status__in=['AUTOMATIC', 'CONFIRMED'])
+                    if len(valid_matches) > 1:
+                        print(
+                            'Multiple valid matches for list item {}'.format(
+                                list_item.id
+                            ))
+                    elif len(valid_matches) < 1:
+                        print(
+                            'No valid matches found for list item {}'.format(
+                                list_item.id
+                            ))
+                    else:
+                        valid_match = valid_matches[0]
+                        if list_item:
+                            list_item.facility_id = valid_match.facility_id
+                            list_item.processing_results.append({
+                                'action': ProcessingAction.UNDO_MERGE,
+                                'started_at': now,
+                                'error': False,
+                                'finished_at': now,
+                            })
+                            if not dry_run:
+                                list_item.save()
+                                print(f'item,save,{list_item.pk}')
+                            else:
+                                print(f'item,dry,{list_item.pk}')
+
+                        try:
+                            match = FacilityMatch.objects.get(pk=hist_match.id)
+                            if not dry_run:
+                                match.status = prev_record.status
+                                match.changeReason = 'Fixing merge rejection'
+                                match.save()
+                                print(f'match,save,{match.pk}')
+                            else:
+                                print(f'match,dry,{match.pk}')
+                        except FacilityMatch.DoesNotExist:
+                            print(f'match,missing,{hist_match.id}')
+                        count = count + 1
+
+            except Exception as e:
+                # In case of an error, we should review/fix the match manually,
+                # but not break the loop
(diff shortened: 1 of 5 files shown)


src/django/api/management/commands/fix_merged_rejections.py
```diff
@@ -0,0 +1,57 @@
...
+    def handle(self, *args, **options):
+        dry_run = options['dry_run']
+
+        if dry_run:
+            print('== DRY RUN - SKIPPING ALL SAVES')
+
+        count = 0
+        for hist_match in HistoricalFacilityMatch.objects.filter(
+            status='MERGED'
+        ):
```
jwalgran: I used `iterator` in this similar management command to avoid caching the rows.  https://github.com/open-apparel-registry/open-apparel-registry/pull/2251/files#diff-28fef09bc57f7679a024157eea04bd5dad854c0de633ff8b058fd8f9e4853f17R29

In prod we have about 90k records so we may not explicitly need to use this 

```
opensupplyhub=> select count(*) from api_historicalfacilitymatch where status = 'MERGED';
 count
-------
 96149
(1 row)
```

Did we explicitly opt not to use iterator? Some guidance on when to use iterator https://stackoverflow.com/a/12688222

src/django/api/management/commands/fix_merged_rejections.py
```diff
@@ -0,0 +1,57 @@
...
+            prev_record = hist_match.prev_record
+            if prev_record.status in ['REJECTED', 'PENDING']:
```
jwalgran: `prev_record` will be `None` if this is the first history record. I don't think we ever create a record with a status of `MERGED`, so we likely don't have to worry about it, though we could consider 

```python
if prev_record and prev_record.status in ['REJECTED', 'PENDING']:
```

src/django/api/management/commands/fix_merged_rejections.py
```diff
@@ -0,0 +1,57 @@
...
+                list_item = prev_record.facility_list_item
```
jwalgran: Historical records in general can link to no longer existing model objects but in this specific case I don't think we ever delete line items.

src/django/api/management/commands/fix_merged_rejections.py
```diff
@@ -0,0 +1,57 @@
...
+                        print(f'item,save,{list_item.pk}')
+                    else:
+                        print(f'item,dry,{list_item.pk}')
+
+                    hist_match.status = prev_record.status
+                    if not dry_run:
+                        hist_match.save()
+                        print(f'match,save,{hist_match.pk}')
+                    else:
+                        print(f'match,dry,{hist_match.pk}')
```
jwalgran: Do we need to, and is it wise to, "rewrite history?" I'm a bit nervous about directly manipulating `HistoricalFacilityMatch` objects since they are just a reflection of the various statues that a `FacilityMatch` has gone through.

Should we instead consider updating the status on the actual `FacilityMatch` and specifying a `changeReason` that explains we are fixing the data?


```suggestion
                    match = FacilityMatch.objects.get(pk=hist_match.id)
                    if not dry_run:
                        match.status = prev_record.status
                        match.changeReason = 'Fixing merge rejection'
                        match.save()
                        print(f'match,save,{match.pk}')
                    else:
                        print(f'match,dry,{match.pk}')
```

jwalgran (commented):
Thanks for writing up such a clear description of the issue and such detail steps for a complicated test process.

I was able to successfully run through the test steps in development, but have a few questions regarding the implementation.

src/django/api/management/commands/fix_merged_rejections.py
TaiWilkin: @jwalgran In line with this, should we add a processing result to the list item with a new action (something like, UNDO_MERGE)?

src/django/api/management/commands/fix_merged_rejections.py
TaiWilkin: I think the dataset is large enough to use iterator, especially as we only use the objects in the loop.

src/django/api/management/commands/fix_merged_rejections.py
TaiWilkin: I'm going to add a check for the item to be safe

src/django/api/management/commands/fix_merged_rejections.py
TaiWilkin: I've pushed a fix up that also adds a processing result to the list item.

TaiWilkin:
Ready for another look

jwalgran (approved):
Changes look good. Good idea updating the processing_results for the fixed items I was able to successfully run through the tests.

TaiWilkin:
Thank you for reviewing!

Issue #2140, opened 2022-09-13
Change references to oar_id in API responses to os_id


jwalgran:
## Overview

Change references to oar_id in API responses to os_id

- Serialized JSON data responses
- API documentation references
- UI references to OAR ID should be OS ID

obrienad:
Mariel confirmed it's OS ID, removed CAN tag.

PR #2276, opened 2022-10-26
Update og:url meta tag to opensupplyhub.org

jwalgran:
## Overview

Update og:url meta tag to opensupplyhub.org

Connects #2140


jwalgran:
Merging this small text change without review to expedite a release.

Issue #2288, opened 2022-10-31
Fix broken info site links in header and footer


jwalgran:
## Overview

Fix broken info site links in header and footer

### Expected Behavior

All links to the info site go to live pages 

### Actual Behavior

These links go to 404 pages

NOTE this list may not be complete

- Header
  - Technology Overview
- Footer
  - Terms of service

### Steps to Reproduce

* Click all info site links in the header and footer.

PR #2289, opened 2022-10-31
Fix header and footer links

mstone121:
## Overview

A few header/footer links are broken. This PR fixes them:
- Technology
- Terms of Service
- Donate

Connects #2288 
Connects #2290 

## Notes

The submenu button-type link was switched from react-router `<Link>`. If a menu button is added that should be an internal link, this will need to be updated, A possible solution is pass down an `external` boolean like how the main header buttons work.

## Testing Instructions

* localhost:6543
* Header > How It Works > The Technology > Overview
* Footer > Terms of Service
* Header & Footer > Donate

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/util/constants.jsx
+++ b/src/app/src/util/constants.jsx
@@ -53,7 +53,7 @@ export const InfoPaths = {
     resources: 'resources',
 
     // Footer
-    termsOfService: 'terms-of-service',
+    termsOfService: 'terms-of-use',
     mediaHub: 'media-hub',
 };
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Removed
 
 ### Fixed
+- Fix broken header and footer links [#2289](https://github.com/open-apparel-registry/open-apparel-registry/pull/2289)
 
 ### Security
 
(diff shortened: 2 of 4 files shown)


jwalgran:
I see the donate links currently working in OS Hub staging and production. If that was the only motivation for changing from Link to a then we may not need that change.

mstone121:
> I see the donate links currently working in OS Hub staging and production. If that was the only motivation for changing from Link to a then we may not need that change.

See #2290

src/app/src/util/constants.jsx
```diff
@@ -32,7 +32,7 @@ export const InfoPaths = {
     researchers: 'researchers',
     serviceProviders: 'service-providers',
     sectors: 'sectors',
-    technology: 'technology',
+    technology: 'technology-overview',
```
TaiWilkin: When I follow this link (/technology-overview), it leads to a 404 page. Oddly, the info site's header currently leads to /technology, which works. I suspect they've changed it since the issue was made earlier.

TaiWilkin (changes requested):
See above note about the 'technology' link - otherwise, this is working well!

mstone121:
> See above note about the 'technology' link - otherwise, this is working well!

[b786fad](https://github.com/open-apparel-registry/open-apparel-registry/pull/2289/commits/b786fad755125cc7c5cf9d2255ad39b71fbe2454)

Also, I noticed that sometimes the social media links wouldn't be clickable. It turns out this was from the footer shape blocking them. [89f3cd1](https://github.com/open-apparel-registry/open-apparel-registry/pull/2289/commits/89f3cd10891366f0508130303f33364295c24d64) sets `pointer-events: none;` on the footer shape.

TaiWilkin (approved):
Perfect. Great catch on the social media buttons.

PR #2298, opened 2022-11-03
Fix embedded map styling

TaiWilkin:
## Overview

The embedded map custom styles weren't applied to all relevant elements when creating the new layout. Applies the custom color/font where applicable, and also tweaks a few small UI style issues. 

Connects Client 115

## Demo

Screen Shot 2022-11-03 at 12 09 11 PM

Screen Shot 2022-11-03 at 12 34 24 PM

With Contrast Text:
Screen Shot 2022-11-03 at 2 29 09 PM
Screen Shot 2022-11-03 at 2 29 22 PM

## Notes

I preferred using the `theme` over pulling in props for custom color/font where possible, to simplify style adjustments later if needed (since the `theme` object is in one place, vs many mapStateToProps). 

Currently, we have three main colors in the app: purple, pink, and yellow. The purple was previously set to be the primary color, so I've left that, and added the pink as secondary, and the yellow as 'action' (since it mostly appears in actionable items, like the search button). The custom type ('action') doesn't have the same power as the built-in primary/secondary color schemes, but is still useful as a variable that can be accessed via theming throughout the app. 

We only allow users to select one color in embedded map config, so in embed mode I've set all the colors to the config color.  However, if we decide to allow users to set multiple colors in the future, it should be fairly simple to apply that via the theme object. 

## Testing Instructions

* Run `./scripts/server`
* View [the homepage](http://localhost:6543/) at full width and mobile. Select two contributors, and view the tooltip for the contributors filter. Note font, color, and alignment throughout.
* View [the facilities page](http://localhost:6543/facilities) at full with and mobile. On mobile, open the filters drawer. Select two contributors, and view the tooltip for the contributors filter. Note font, color, and alignment throughout.
* Go to user embed settings. Note that text should read, 'OS ID' and not 'OS Hub ID'. 
* Set up an embedded map with a very noticeable color and font. Set a field to be searchable, and update the search text. 
* View [the embedded facilities page](http://localhost:6543/facilities?contributors=2&embed=1) at full width and mobile. On mobile, open the filters drawer. Note the alignment of the 'powered by' button. Note font, color, and alignment throughout.
* Test the embedded map again with a user with lowest embed permissions, and the map color set to 'black'.

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/app/src/components/DownloadFacilitiesButton.jsx
+++ b/src/app/src/components/DownloadFacilitiesButton.jsx
@@ -5,6 +5,7 @@ import Button from '@material-ui/core/Button';
 import Menu from '@material-ui/core/Menu';
 import MenuItem from '@material-ui/core/MenuItem';
 import Tooltip from '@material-ui/core/Tooltip';
+import { withStyles } from '@material-ui/core/styles';
 
 import { toast } from 'react-toastify';
 
@@ -12,16 +13,33 @@ import downloadFacilities from '../actions/downloadFacilities';
 import DownloadIcon from './DownloadIcon';
 import ArrowDropDownIcon from './ArrowDropDownIcon';
 
-const downloadFacilitiesStyles = Object.freeze({
-    listHeaderButtonStyles: Object.freeze({
-        height: '45px',
-        margin: '5px 0',
-    }),
-    buttonText: Object.freeze({
-        marginLeft: '0.2rem',
-        marginRight: '0.6rem',
-    }),
-});
+const downloadFacilitiesStyles = theme =>
+    Object.freeze({
+        listHeaderButtonStyles: Object.freeze({
+            height: '45px',
+            margin: '5px 0',
+            backgroundColor: theme.palette.action.main,
+            fontSize: '16px',
+            fontWeight: 900,
+            lineHeight: '20px',
+            '&:hover': {
+                backgroundColor: theme.palette.action.dark,
+            },
+        }),
+        buttonText: Object.freeze({
+            marginLeft: '0.2rem',
+            marginRight: '0.6rem',
+        }),
+        downloadTooltip: {
+            fontSize: '0.875rem',
+            fontFamily: theme.typography.fontFamily,
+        },
+        buttonContent: {
+            display: 'flex',
+            alignItems: 'center',
+            textTransform: 'none',
+        },
+    });
 
 function DownloadFacilitiesButton({
     /* from state */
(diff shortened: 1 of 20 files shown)


TaiWilkin:
Hold off on review briefly; missed a font color issue.

TaiWilkin:
With very dark colors, the text wasn't legible. I've updated to use dynamic text colors on the items with a changeable background color.

src/app/src/components/DownloadFacilitiesButton.jsx
```diff
@@ -19,6 +19,7 @@ const downloadFacilitiesStyles = theme =>
             height: '45px',
             margin: '5px 0',
             backgroundColor: theme.palette.action.main,
+            color: theme.palette.getContrastText(theme.palette.action.main),
```
TaiWilkin, line 22: We have the calculate the contrast text for 'action'; as a custom text set, it doesn't get calculated automatically

src/app/src/components/FilterSidebar.jsx
```diff
@@ -74,16 +75,18 @@ const filterSidebarStyles = theme =>
         },
         filterButton: {
             backgroundColor: theme.palette.action.main,
-            color: '#000',
+            '&:hover': {
+                backgroundColor: theme.palette.action.dark,
+            },
+            color: theme.palette.secondary.contrastText,
```
TaiWilkin, line 81: We don't have to calculate the secondary text color; MUI assigns it to this field automatically, since 'secondary' is a known text field

src/app/src/components/FilterSidebar.jsx
```diff
@@ -127,7 +130,12 @@ class FilterSidebar extends Component {
     }
 
     render() {
-        const { fetchingFeatureFlags, classes, secondaryColor } = this.props;
+        const { fetchingFeatureFlags, classes, theme } = this.props;
+
+        const actionContrastText = theme.palette.getContrastText(
```
TaiWilkin, line 135: The contrast text is needed outside of the styles object because the icons need the color passed in directly, not via a class

jwalgran (commented):
Excellent work on a large styling refactor. I am seeing custom styling options on embedded map working as they dis in OAR.

While testing found 2 things that we will want to fix.

First a quick one. Now is a good time to change this to "Green"

Screen Shot 2022-11-04 at 2 44 28 PM

When testing on an actual touch device (via BrowserStack) I could not get the multiple contributor info popup to appear when I tap it.  This screencast shows browsing this brnach gs production using Safari on an iPhone 8 running iOS 15

https://user-images.githubusercontent.com/17363/200080093-b1751dc8-9700-4d33-bfcb-10ae68477008.mp4

jwalgran (approved):
The changes look good. 

For our usage it would be preferable for the tooltip to stay visible until tapping some other element but it does not look like the API supports that. 5 seconds is a reasonable guess at the delay time and we don't have the budget to try another solution.

TaiWilkin:
Thank you for reviewing!

PR #2299, opened 2022-11-03
Handle additional merge errors

TaiWilkin:
## Overview

Updates the merge error correction script to handle additional faulty merge cases, such as instances where all matches for an overwritten item were pending.

I ran this via a test branch on staging and it succeeded with a minimal number of items to be manually fixed (there's no way to get around needing to manually update items where there are multiple possible valid matches, and a few items no valid matches don't seem to fall under either of the additional cases and should be investigated/adjusted manually if needed). 

Connects #2233

## Testing Instructions

* A test branch should be currently live on staging, or can be rerun to make it live `ogr/test/handle-more-merge-errors`
* Run `./scripts/manage ecsmanage fix_merged_rejections --dry-run` to see the script run on staging
* Confirm that the results seem reasonable

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/management/commands/fix_merged_rejections.py
+++ b/src/django/api/management/commands/fix_merged_rejections.py
@@ -1,7 +1,7 @@
 from django.core.management.base import BaseCommand
 from django.utils import timezone
 
-from api.models import FacilityMatch, HistoricalFacilityMatch
+from api.models import FacilityMatch, HistoricalFacilityMatch, FacilityListItem
 from api.constants import ProcessingAction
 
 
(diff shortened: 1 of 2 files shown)


src/django/api/management/commands/fix_merged_rejections.py
```diff
@@ -18,8 +18,39 @@ def handle(self, *args, **options):
         if dry_run:
             print('== DRY RUN - SKIPPING ALL SAVES')
 
-        count = 0
         now = str(timezone.now())
+
+        def update_list_item(item, facility_id, note=''):
```
jwalgran: I don't see `update_list_item` called with a value for the `note` parameter. Consider removing it.

src/django/api/management/commands/fix_merged_rejections.py
```diff
@@ -33,47 +64,85 @@ def handle(self, *args, **options):
...
                 # but not break the loop
                 print(f'match,error,{hist_match.id},{e}')
 
+        print(f'handling items with no valid matches,{no_valid_matches}')
+        for item_id in no_valid_matches:
+            try:
+                hist_matches = HistoricalFacilityMatch.objects.filter(
+                     facility_list_item_id=item_id,
+                     status='MERGED'
+                )
```
jwalgran: I assume we wand to process these HistoricalFacilityMatch records from oldest to newest. Are we sure that they are sorted that way by default?

jwalgran (approved):
This process looks reasonable to me although I will admin that is is complex and I do not fully understand the entire process in detail

I used [`awslogs`](https://github.com/jorgebastida/awslogs) to download the full log stream from my test run on staging

```
awslogs get -s 1d logOpenSupplyHubStagingAppCLI management/django/ae889777822e45e19e0431d3a9c4901a > ae889777822e45e19e0431d3a9c4901a.txt
```

[ae889777822e45e19e0431d3a9c4901a.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/9955645/ae889777822e45e19e0431d3a9c4901a.txt)

Some counts of of various string matches in the log output
```
> cat ae889777822e45e19e0431d3a9c4901a.txt | grep "No valid matches for" | wc -l
     215
> cat ae889777822e45e19e0431d3a9c4901a.txt | grep "match,dry" | wc -l
   17468
> cat ae889777822e45e19e0431d3a9c4901a.txt | grep "undo_merged_with_valid" | wc -l
    2963
> cat ae889777822e45e19e0431d3a9c4901a.txt | grep "undo_all_invalid_merged" | wc -l
    5257
> cat ae889777822e45e19e0431d3a9c4901a.txt | grep "missing" | wc -l
       0
```

I also saw 524 items in  the "Multiple valid matches for" set at the end of the run  and did not find any lines

TaiWilkin:
Thank you for reviewing!

Issue #2335, opened 2022-12-14
Exclude unapproved lists from the contributor-lists API


jwalgran:
## Overview

The contributor-lists API is used to populate the list filter dropdown. If the list is not approved, it will not have been processed and will therefore never return results. Unapproved lists should be excluded from the API response.

PR #2336, opened 2022-12-14
Only include approved lists in the facility-lists API 

jwalgran:
## Overview

Lists that are not approved will have active facility matches so we want toexclude non-approved lists from the the API used to populate the list filter dropdown.

We also update the `processfixtures` management command. The command is used for loading development data and all the lists processed by this command should have their status set to APPROVED to more closely match the real production process.

Connects #2335 

## Testing Instructions

* Run `./scripts/resetdb` and verify that it completes without error
* Log in as c3@example.com
* Browse http://localhost:6543/contribute
* In "Select a list to replace" choose "Summer 2019 Compliance List" and upload 
[azavea.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/10231597/azavea.csv)
* Browse http://localhost:6543/, select Factory A from the "Data Contributor" filter and verify that the "Contributor List" dropdown does not contain the new "azavea" list
* Run the parse step on the new list
* In a separate browser session log in as c1@example.com
  * Browse http://localhost:6543/dashboard/lists/ and select the pending "azavea" list
  * Approve the list
* Back in the original browser session refresh http://localhost:6543/, select Factory A from the "Data Contributor" filter and verify that the "Contributor List" dropdown now contains both lists.
* Finish processing the new list
* Back in the original browser session refresh http://localhost:6543/, select Factory A from the "Data Contributor" filter and verify that the "Contributor List" dropdown now contains just the "azavea" list

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Copy edits [#2323](https://github.com/open-apparel-registry/open-apparel-registry/pull/2323)
 - Keep list active until replacement is fully matched [#2333](https://github.com/open-apparel-registry/open-apparel-registry/pull/2333)
+- Only include approved lists in the facility-lists API [#2336](https://github.com/open-apparel-registry/open-apparel-registry/pull/2336)
 
 ### Deprecated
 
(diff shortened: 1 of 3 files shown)


lydiascarf (approved):
lgtm, runs as expected!

jwalgran:
Thanks for the review.

Issue #2326, opened 2022-12-05
List approval workflow status updates are needed. Rejected lists to become "inactive". Replaced lists to become "rejected".


vrwOAR:
## Overview

In the new upload workflow, contributor lists are reviewed by OS Hub team. When [rejected](https://opensupplyhub.org/lists/2035), the list will stop progressing through the workflow and a rejection message is delivered to the contributor. Though rejected, the lists are maintaining their active status in the [list source dashboard](https://opensupplyhub.org/admin/api/source/406807/change/?_changelist_filters=contributor__id__exact%3D194) and thus remain on the [front end as an available list](https://opensupplyhub.org/profile/195?q=Akh%20Eco%20Apparels%20Ltd) for said contributor. The list details (list name and description) are available, however, the [facilities and all corresponding facility data points](https://opensupplyhub.org/facilities/?contributors=194&lists=2035) are not available; as the list was rejected prior to geocoding and matching.

Also, as a part of the upload process, a user has historically had the option to replace a previously uploaded list. Lists uploaded to OS Hub that are replaced with a new upload are being made inactive. However, their status within the Django facility list details remains as “Pending”. Therefore the list remains visible within both the OS Hub Dashboard & the user’s dashboard as a pending list awaiting inputs from the OS Hub admin. 

### Expected Behavior

- When a list is rejected during the list approval process the list should also be made inactive in the list source dashboard.
- When a list uploaded to OS Hub is replaced by the user, the status should be updated to “inactive” and “rejected”. It is not necessary to send the rejected notification to the user for this use case, as they have already replaced the list with another upload for processing. 

### Actual Behavior

- Though rejected, the lists are maintaining their "active" status in the [list source dashboard](https://opensupplyhub.org/admin/api/source/406807/change/?_changelist_filters=contributor__id__exact%3D194) and thus remain on the [front end as an available list](https://opensupplyhub.org/profile/195?q=Akh%20Eco%20Apparels%20Ltd) for said contributor.
- Replaced lists are retaining the "pending" status in the change facility list dashboard. Example, [Mammut Sports Group - Mammut Sports Group AG Supplier List 2022](https://opensupplyhub.org/admin/api/facilitylist/2215/change/)

### Steps to Reproduce

Rejected List items:
- Log-in to [OS hub admin dashboard](https://opensupplyhub.org/dashboard)
- navigate to desired list from [rejected list](https://opensupplyhub.org/dashboard/lists/?status=REJECTED)
[...]

PR #2362, opened 2023-02-06
Approval workflow updates

mstone121:
## Overview

This PR addresses two isses with the approval workflow:

1. When a facility list is rejected, that facility list's source remains active.
2. Facility lists that are replaced by another facility list remain in the Pending status. A new meta-status, Replaced, was created for these lists.

Connects #2326 

## Notes

Instead of creating a new status for Replaced lists, we had the option to create a "real" status. This would have been simpler to implement, but could lead to data inconsistency (e.g. a non-replaced list with the Replaced status). In addition, if a replacing list is rejected and we want to return the original list, there would be no way to know its original status. 

## Testing Instructions

- `./scripts/test`
---
- Upload a facility list
- `./scripts/manage batch_process --list-id <uploaded list ID> --action parse`
- Login as an admin of the uploaded list's contributor (must be a different user)
- Reject uploaded, pending list
  - [x] Ensure list is Rejected status
  - [x] Ensure list's source is not active

- Upload a facility list
- Upload a replacement list
  - [x] Ensure status of new list is Replaced
  - [x] Ensure it doesn't show up in the UI facility list lists
    - [x] Except for the admin dashboard list when 'Rejected' is selected

## Checklist

- [x] CI passes after rebase


src/django/api/tests.py
```diff
@@ -300,10 +300,21 @@ def test_replaces(self):
         self.assertEqual(FacilityList.objects.all().count(),
                          previous_list_count + 2)
         self.assertEqual(new_list.replaces.id, original_list.id)
+        self.assertEqual(new_list.status, FacilityList.PENDING)
+
+        # API should return REPLACED meta-status
+        response = self.client.get(f'/api/facility-lists/{original_list.id}/')
```
mstone121: I couldn't figure out how to reverse this url. Tried
```python
reverse('facility-list-retrieve', args=[original_list.id])
```
and
```python
FacilityListViewSet.reverse_action('retrieve', args=[original_list.id])
```

src/django/api/constants.py
```diff
@@ -81,6 +81,7 @@ class FeatureGroups:
 
 class FacilityListStatus:
```
mstone121: This class isn't referred to by anything except the `AdminFacilityListView` `get_queryset` method. Perhaps it can be removed in favor of `FacilityList` statuses.

src/django/api/views.py
```diff
@@ -3141,7 +3141,7 @@ def reject(self, request, pk=None):
             }
         """
         try:
-            facility_list = self.queryset.get(id=pk)
+            facility_list = self.queryset.select_related('source').get(id=pk)
```
jwalgran, line 3149: Confirming my understanding of this addition. Because we are always going to be accessing `facility_list.source` as part of rejects we go ahead and fetch the related model.

src/django/api/constants.py
jwalgran: This was introduced in c1cdb8b

I do not see an obvious reason why this was introduced instead of using `FacilityList.MATCHED`. I would +1 removing `FacilityListStatus`

jwalgran (changes requested):
I agree with the design choice to not add a `REPLACED` status. The other status values are related to approval and mixing in this value unrelated to approval would complicate things in the ways that you described. I did see a few behaviors that I suggest changing


I started these test step from `./scripts/resetdb`

- Log in as c3@example.com
- Contribute embed-test-1.csv
- Run `./scripts/manage batch_process --list-id 16 --action parse`
- In a separate browser session log in as the admin c1@example.com
	- Browse http://localhost:6543/dashboard/lists/
	- Approve the list
- Run  ./tools/batch_process 16
- From the regular user session, contribute embed-test-2.csv and choose the embed-test-1 as the replacement
- Run `./scripts/manage batch_process --list-id 17 --action parse`

At this point if you browse http://localhost:6543/ and choose `Factory A` as the contributor, the Contributor List dropdown still correctly shows `embed-test-1` because the replacement list has not yet been processed however browsing http://localhost:6543/lists does not show the `embed-test-1` list. 

**Requested change 1:** Replaced lists should always show up in the user's `/lists` page but the status on the details page should be `REPLACED` rather than `APPROVED`

The steps to reproduce section of #2326 mentions viewing a list in the Django Admin and that there is no indication that a list has been replaced.

**Requested change 2:**  Customize the Django admin page via `admin.py`. If the list has been replaced, show the ID and name of the list that replaced it and ideally, if it is not a large amount of extra work, hyperlink to that list in the Django admin.

src/django/api/views.py
mstone121: Yep. It saves a trip to the DB.

src/django/api/constants.py
mstone121: [07b1549](https://github.com/open-apparel-registry/open-apparel-registry/pull/2362/commits/07b15492fa12345e790333ee19f372fc8e798b96)

mstone121:
**Requested change one:** [e6da69d](https://github.com/open-apparel-registry/open-apparel-registry/pull/2362/commits/e6da69dce047dedae8d4c0e0db58e860ed12b713)

**Requested change two:** [5018a25](https://github.com/open-apparel-registry/open-apparel-registry/pull/2362/commits/5018a25322a19afb013f2c8edb27176221c40a4b)

src/django/api/admin.py
```diff
@@ -61,6 +61,18 @@ class FacilityHistoryAdmin(SimpleHistoryAdmin):
     readonly_fields = ('created_from',)
 
 
+class FacilityListAdmin(admin.ModelAdmin):
+    readonly_fields = ('replaced_by_link',)
```
mstone121, line 66: If this is just `replaced_by`, Django will automatically create a link to the replacing facility list without the method below. However, it displays a label of `Facility list:` and I couldn't figure out how to override that.

src/django/api/tests.py
jwalgran: I used 

```
./scripts/manage show_urls | grep facility-lists
```

To find the URL we are looking for and get the the name

```
...
/api/facility-lists/<pk>/	api.views.FacilityListViewSet	facility-list-detail
....
```

Using the name `facility-list-detail` appears to get us what we need

```python
>>> from django.urls import reverse
>>> reverse('facility-list-detail', args=[1])
'/api/facility-lists/1/'
```

src/django/api/admin.py
jwalgran: I could not find a way of changing the label either. This is a reasonable workaround 👍

src/django/api/admin.py
```diff
@@ -61,6 +61,18 @@ class FacilityHistoryAdmin(SimpleHistoryAdmin):
...
+
+    @admin.display(description='Replaced by')
+    def replaced_by_link(self, obj):
+        return mark_safe(
+            '<a href="/admin/api/facilitylist/{}/change/">{}</a>'.format(
```
jwalgran: We could use `reverse` here, but I'm not convinced that the change is worth it (this code was based on https://stackoverflow.com/a/1720961)
```python
        url = reverse('admin:{}_{}_change'.format(obj._meta.app_label,
                                                  obj._meta.model_name),
                      args=[obj.replaced_by.id])
        return mark_safe(
            '<a href="{}">{}</a>'.format(url, obj.replaced_by)
        )
```

jwalgran (approved):
Thanks for integrating the requested revisions. This is a clear implementation and ran through the tests without issue.

jwalgran:
I just noticed that the changes to the models file require a migration

```
Running migrations:
  No migrations to apply.
  Your models in app(s): 'api' have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
[+] Running 2/0
 ⠿ Container open-apparel-registry-memcached-1  Running                                                                                 0.0s
 ⠿ Container open-apparel-registry-database-1   Running                                                                                 0.0s

205 static files copied to '/usr/local/src/static', 670 post-processed.
~/Projects/open-apparel-registry (pr-2362)
>
~/Projects/open-apparel-registry (pr-2362)
~/Projects/open-apparel-registry (pr-2362)
> ./scripts/manage makemigrations
[+] Running 2/0
 ⠿ Container open-apparel-registry-database-1   Running                                                                                 0.0s
 ⠿ Container open-apparel-registry-memcached-1  Running                                                                                 0.0s
Migrations for 'api':
  api/migrations/0112_alter_facilitylist_replaces.py
    - Alter field replaces on facilitylist
~/Projects/open-apparel-registry (pr-2362)
>

```

mstone121:
Add migration here: https://github.com/open-apparel-registry/open-apparel-registry/pull/2362/commits/5bcca17860b5763b30f54994c05806742e24a020

src/django/api/tests.py
mstone121: Ah, I forgot about `show_urls`. Thanks! https://github.com/open-apparel-registry/open-apparel-registry/pull/2362/commits/dcfb1e0d68fa3d0b096ce3599e9f1b8a47c3ccbe

src/django/api/admin.py
mstone121: What about https://github.com/open-apparel-registry/open-apparel-registry/pull/2362/commits/64329465c616feadd89443c648cb3f3c17f1cec4?

jwalgran: Looks good 👍

jwalgran:
I didn't retest but the changes all look good to me. Thanks.

jwalgran (approved):

Issue #1847, opened 2022-05-16
Merged OAR ID's redirect when searched


hlennett:
## Overview

Currently, after an OAR ID is merged into another one, if a user tries to access that old facility profile via an old URL, it will redirect to the new facility profile. However, if you search by the old OAR ID, no results are returned. It would be great for the OAR ID facility search to redirect as well (or to have some kind of behavior that shows that the facility still exists and has just been merged into a new OAR ID). 

An example: https://openapparel.org/facilities/BD2020009B5R296 redirects to the new facility profile but https://openapparel.org/?q=BD2020009B5R296 does not.

hlennett:
@mariel-oar Noting that this is coming up more and more in our conversations with facilities. Wonder if it's worth a conversation of whether this can make it into the PRD for OS Hub?

vrwOAR:
@mariel-oar - Two recent examples of merges that are not searchable on FE
- BA20221713TPRZW
- TR2022249BKAX66

an example from a 2021 merge
- VN2019176GPFHDH

mariel-oar:
@obrienad @jwalgran , flagging a new ticket for PB. Can we get an estimate on this one? thank you!

jwalgran:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Civic-Apps-5c8fabc1177a1c1f6672f901/issues/open-apparel-registry/open-apparel-registry/1847?planning-poker) with Zenhub @TaiWilkin @mstone121

obrienad:
@mariel-oar - estimate is confirmed. Tagged CAN pending your greenlight.

PR #2366, opened 2023-02-08
Check facility query string for aliased OS ID

mstone121:
## Overview

This PR adds checks to the facility search logic. If the query string matches the regex patter of an OS ID, the facility alias table is searched. If a match is found, the `id` parameter is overwritten with the aliased ID.

Connects #1847 

## Notes

Additionally, if the query string matches the OS ID regex pattern, the query string parameter is cleared so as not to conflict with the ID queryset filter. This means that if someone inteneded to search for a string that matches the pattern, but is not actually an OS ID, they would not see the expected results. Since the OS ID regex pattern is pretty specific, this seems unlikely to occur.

## Testing Instructions

- `./scripts/manage test api.tests.FacilitySearchTest`
---

* Create a facility alias in the `api_facilityalias` table
* Search for that facility alias id
  * [ ] Ensure the aliases facility shows up in the list

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/src/django/api/os_id.py
+++ b/src/django/api/os_id.py
@@ -1,5 +1,6 @@
 import base32_crockford
 import random
+import re
 
 from django.utils import timezone
 
@@ -77,3 +78,10 @@ def validate_os_id(raw_id, raise_on_invalid=True):
         else:
             return False
     return True
+
+
+os_id_regex = re.compile('[A-Z]{2}[0-9]{7}[A-Z0-9]{6}')
+
+
+def string_matches_os_id_format(string):
+    return os_id_regex.match(string or '') is not None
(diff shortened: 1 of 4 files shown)


src/django/api/os_id.py
```diff
@@ -77,3 +78,10 @@ def validate_os_id(raw_id, raise_on_invalid=True):
         else:
             return False
     return True
+
+
+os_id_regex = re.compile('[A-Z]{2}[0-9]{7}[A-Z0-9]{6}')
+
+
+def string_matches_os_id_format(string):
+    return os_id_regex.match(string or '') is not None
```
jwalgran, line 87: Good design to locate this in the `os_id` module 👍

jwalgran (approved):
I tested this by merging a facility via the admin dashboard to replicate the condition described in #1847 and the new search logic worked correctly.

mstone121:
Thanks for the review!

Issue #2376, opened 2023-03-01
Take a final RDS snapshot and destroy OAR production resources


mariel-oar:
This will require a different set of steps then the staging destroy (#2351) since we need to leave the DNS/CloudFront/certificate setup in place to handle the redirect until we can migrate those resources to the OSH deployment.
- [x] Remove all resources from .tf files other than those required for DNS/CloudFront/certificate
- [x] Run a production `./scripts/infra plan` and verify the output is expected and that the resources needed for the redirect remain in place
- [x] Create an RDS snapshot
- [x] Enable deletion protection on both OS Hub RDS instances via the console (for extra confidence)
- [x] Use `./scripts/infra apply` to destroy the resources

jwalgran:
I updated the description since, unlike staging, there are a few non-compute resources we want to leave in place to handle the old domain name and redirection.

PR #2378, opened 2023-03-03
Remove most production resources

nanotubing:
## Overview

we're updating the terraform to remove all resources from the OAR production environment, except for cloudfront and its related DNS and certificate resources.

Connects #2376

## Demo
attached is the output from running `scripts/infra plan` in this branch
[2376_plan_output.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/10885656/2376_plan_output.txt)

## Testing Instructions

* Run a `scripts/infra plan` in this branch and review the output to verify that nothing necessary is being removed

## Checklist

- [ ] CI passes after rebase


nanotubing:
I updated my commit, which is now also removing the infra stage from the Jenkinsfile to avoid re-creating resources in the staging environment.

It is not expected that the output of `scripts/infra plan` would have changed based on this update, but the current plan output is attached anyway.

[2376_plan_output_b.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/10902242/2376_plan_output_b.txt)

jwalgran (approved):
I was able to run plan against production successfully. This looks correct. There are two resources that look like we may also be able to remove.

- The SNS topic in `alarms.tf`. I do not think there are any resources remaining that publish to this topic.
- The `"internal"` hosted zone in `dns.tf` When I search in the `terraform` directory I do not see any remaining references to `internal`
- The VPC in `networking.tf`. Since our remaining resources are not within the VPC it looks like we may be able to remove this as well.

nanotubing:
> I was able to run plan against production successfully. This looks correct. There are two resources that look like we may also be able to remove.

I pushed another commit that removes these resources. I did not get any errors in the plan output, and everything seems correct to my eyes
[2376_plan_output_c.txt](https://github.com/open-apparel-registry/open-apparel-registry/files/10903470/2376_plan_output_c.txt)

jwalgran (approved):

nanotubing:
This terraform changeset has been applied, and the OAR production environment is gone.

Issue #2369, opened 2023-02-20
IndexError: list index out of range raised when downlaoding from embeded map and header does not match data fields


jwalgran:
## Overview

`IndexError: list index out of range` raised when downloading from embedded map and header does not match data fields

### Steps to Reproduce

**NOTE Make sure to close the tab after testing this to prevent looping requests ( #2368)**

* Browse https://opensupplyhub.org/facilities?contributors=25&embed=1
* Download

### Demo

In this screenshot note the discrepancy in length between `data_values` and `list_fields`. The code expects these to be the same length with fields in the same order.

<img width="601" alt="Screenshot 2023-02-20 at 12 47 46 PM" src="https://user-images.githubusercontent.com/17363/220188078-07f3b5bd-7c6f-4882-bb4e-83d7cf8a09a7.png">

### Additional context

https://rollbar.com/OpenApparelRegistry/OpenSupplyHub/items/218/

jwalgran:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Civic-Apps-5c8fabc1177a1c1f6672f901/issues/open-apparel-registry/open-apparel-registry/2369?planning-poker) with Zenhub @mstone121

jwalgran:
@mariel-oar labeled client action needed to get your approval for fixing this bug. I would classify this as relatively high priority because it results in embedded map downloads that appear to never finish and is the root cause behind #2368

mariel-oar:
Yes, please fix this @jwalgran...removing CAN

PR #2379, opened 2023-03-07
Fix facility download

TaiWilkin:
## Overview

Handles the case when an individual list item's raw data differs in length from the number of headers.

Connects #2369 

## Demo

[facilities.csv](https://github.com/open-apparel-registry/open-apparel-registry/files/10910514/facilities.csv)

## Notes

Based on the way the lists are processed, there really shouldn't be a case where any item has raw data of a different length than the headers. I wanted to identify the source of the damaged data, but when testing out the error in production, the contributor whose list was failing previously didn't throw any errors. I attempted to identify a recent change to their data that might have fixed it, but was unable to. I also loaded several of the user's recent lists into my local database, and wasn't able to replicate the error that way. 

I ended up deleting some of the raw data fields for a list item locally, and was able consistently trigger the error that way. 

## Testing Instructions

On `ogr/develop`:
* Upload a list with custom fields and fully process it. 
* Add the custom fields to your embedded map.
* Use `./scripts/manage shell_plus` to remove the custom fields from the `raw_data` of one of the items in your list.
* Visit your embedded map, and download the facilities. 
      - If you have more than a page worth of facilities, the infinite loop error will be triggered.
      - If you have less than a page of facilities, there will be an error but no infinite loop.

On this branch:
* Attempt to download the same embedded map. It should succeed. 
 
## Checklist

- [ ] CI passes after rebase

### DIFF
--- a/src/django/api/helpers.py
+++ b/src/django/api/helpers.py
@@ -69,7 +69,11 @@ def get_list_contributor_field_values(item, fields):
     for f in fields:
         if f['column_name'] in list_fields:
             index = list_fields.index(f['column_name'])
-            f['value'] = try_parse_int_from_float(data_values[index])
+            if 0 <= index < len(data_values):
+                value = data_values[index]
+            else:
+                value = None
+            f['value'] = try_parse_int_from_float(value)
 
     return fields
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Removed
 
 ### Fixed
+- Fix facility download [#2379](https://github.com/open-apparel-registry/open-apparel-registry/pull/2379)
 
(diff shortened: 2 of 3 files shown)


src/django/api/helpers.py
```diff
@@ -69,7 +69,11 @@ def get_list_contributor_field_values(item, fields):
     for f in fields:
         if f['column_name'] in list_fields:
             index = list_fields.index(f['column_name'])
-            f['value'] = try_parse_int_from_float(data_values[index])
+            if 0 <= index < len(data_values):
```
jwalgran, line 72: TIL that I no longer have to write this

```python
if 0 <= index and index < len(data_values):
```

Thank you

jwalgran (approved):
Using the test instruction I was able to reproduce the exception on `ogr/develop` and saw them resolved on this branch.

TaiWilkin:
Thank you for reviewing!

PR #2382, opened 2023-03-13
Enable deletion protection for RDS databases

jwalgran:
## Overview

We want to be forced to take the extra step of disabling this to avoid accidental deletion of our most valuable resource.

Connects #2376 

## Demo

Running a `plan` on OS Hub staging includes this resource change

```
  # module.database_enc.aws_db_instance.postgresql will be updated in-place
  ~ resource "aws_db_instance" "postgresql" {
      ~ deletion_protection                   = false -> true
        id                                    = "opensupplyhub-enc-stg"
        name                                  = "opensupplyhub"
        tags                                  = {
            "Environment" = "Staging"
            "Name"        = "DatabaseServer"
            "Project"     = "OpenSupplyHub"
        }
        # (59 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
```

## Testing Instructions

* Verify the change

## Checklist

- [x] CI passes after rebase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 - Update spelling of TR to Türkiye and accept multiple versions [#2375](https://github.com/open-apparel-registry/open-apparel-registry/pull/2375)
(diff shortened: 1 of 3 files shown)


nanotubing (approved):
I confirmed that the plan output seems correct, so I ran an `infra apply` against staging, and it applied cleanly. I logged into the console, and verified that deletion prevention is on.

jwalgran:
Thanks for the review.

